PR #4, opened 2020-07-06
Start to define syntax and semantics

glyn:
Start to define syntax and semantics. This only goes as far as `$` but sets the scene for other subpaths to be added.

Add temporary warning about this not being ready for publication.

Also add a utility for generating HTML and text files from XML and for checking the XML and ABNF.

### DIFF
--- a/draft-normington-jsonpath-latest.xml
+++ b/draft-normington-jsonpath-latest.xml
@@ -105,7 +105,40 @@
       </section>
     </section>
    <section numbered="true" toc="default">
-      <name>JSONPath Syntax and Semantics</name>         
+      <name>JSONPath Syntax and Semantics</name>
+         <section numbered="true" toc="default">
+            <name>Node Matching Semantics</name>
+            <t>A JSONPath is a string which matches zero or more nodes of any JSON document.
+            </t>
+            <t>A JSONPath is abstractly a sequence of matchers. Each matcher takes a list of nodes and produces a possibly different list of nodes.
+               The initial list of nodes, consisting of just the root node of a given JSON document, is passed to the first matcher of the sequence.
+               The output from each matcher is passed to the next matcher in the sequence, and so forth until the end of the sequence is reached.
+               The list of nodes produced by the last matcher is the result of matching the JSONPath to the given JSON document.
+            </t>
+            <t>If the JSONPath is empty, this correspondds to an empty sequence of matchers and matches the root node of the JSON document.
+               That is, the result is a list consisting of just the root node of the JSON document.
+            </t>
+            <t>As a consequence, if any of the matchers matches no nodes, then the whole JSONPath matches no nodes.
+            </t>
+            <t>The semantics of each matcher are defined for each type of node. Each node holds a JSON value
+               (as defined by <xref target="RFC8259" format="default">RFC&nbsp;8259</xref>), that is, an object, array, number, string, or one
+               of the literals <tt>true</tt>, <tt>false</tt>, or <tt>null</tt>.
+            </t>
+            <section numbered="true" toc="default">
+               <name>Basics</name>
+               <t>JSONPaths start by matching the root node of a JSON document. There is an explicit notation <tt>$</tt> for a matcher which matches just the root node:
+               </t>
+               <sourcecode type="abnf">
+path = root subpath ; 
+root = %x24         ; $ matches the root node of a document
+
+subpath = identity  ; see below for alternatives
+identity = ""       ; matches all input nodes
+               </sourcecode>
+               <t>Valid JSONPaths MUST conform to the grammar of <tt>path</tt>.
+               </t>
+         </section>
+         </section>       
    </section>
    <section anchor="IANA" numbered="true" toc="default">
       <name>IANA Considerations</name>
--- a/draft-normington-jsonpath-latest.xml
+++ b/draft-normington-jsonpath-latest.xml
@@ -109,37 +109,79 @@
    <section numbered="true" toc="default">
       <name>JSONPath Syntax and Semantics</name>
          <section numbered="true" toc="default">
-            <name>Node Matching Semantics</name>
-            <t>A JSONPath is a string which matches zero or more nodes of any JSON document.
+            <name>Overview</name>
+            <t>A JSONPath is a string which matches zero or more nodes of any JSON document. A valid JSONPath conforms to the ABNF syntax defined by this document.
             </t>
-            <t>A JSONPath is abstractly a sequence of matchers. Each matcher takes a list of nodes and produces a possibly different list of nodes.
-               The initial list of nodes, consisting of just the root node of a given JSON document, is passed to the first matcher of the sequence.
-               The output from each matcher is passed to the next matcher in the sequence, and so forth until the end of the sequence is reached.
-               The list of nodes produced by the last matcher is the result of matching the JSONPath to the given JSON document.
-            </t>
-            <t>If the JSONPath is empty, this correspondds to an empty sequence of matchers and matches the root node of the JSON document.
-               That is, the result is a list consisting of just the root node of the JSON document.
-            </t>
-            <t>As a consequence, if any of the matchers matches no nodes, then the whole JSONPath matches no nodes.
-            </t>
-            <t>The semantics of each matcher are defined for each type of node. Each node holds a JSON value
-               (as defined by <xref target="RFC8259" format="default">RFC&nbsp;8259</xref>), that is, an object, array, number, string, or one
-               of the literals <tt>true</tt>, <tt>false</tt>, or <tt>null</tt>.
-            </t>
-            <section numbered="true" toc="default">
-               <name>Basics</name>
-               <t>JSONPaths start by matching the root node of a JSON document. There is an explicit notation <tt>$</tt> for a matcher which matches just the root node:
-               </t>
-               <sourcecode type="abnf">
-path = root subpath ; 
-root = %x24         ; $ matches the root node of a document
-
-subpath = identity  ; see below for alternatives
-identity = ""       ; matches all input nodes
-               </sourcecode>
-               <t>Valid JSONPaths MUST conform to the grammar of <tt>path</tt>.
-               </t>
          </section>
+         <section numbered="true" toc="default">
+            <name>Terminology</name>
+            <t>A JSON document is logically a tree of nodes with the document as the root node of the tree.
+            </t>
+            <t>Each node holds a JSON value (as defined by <xref target="RFC8259" format="default">RFC&nbsp;8259</xref>) of one of the types object, array, number, string, or one
+               of the literals <tt>true</tt>, <tt>false</tt>, or <tt>null</tt>. The type of the JSON value held by a node is sometimes referred to as the type of the node.
+            </t>
+            <t>The descendants of a node are taken to include the node itself and the descendants of any children of the node.
+            </t>
+         </section>
+         <section numbered="true" toc="default">
+            <name>Implementation</name>
+            <t>An implementation of this specification, from now on referred to simply as "an implementation", SHOULD takes two inputs, a JSONPath and a JSON document, and produce
+               a possibly empty list of nodes of the JSON document which match the JSONPath or an error (but not both).
+            </t>
+            <t>If there is no matching node and no error has occurred, an implementation MUST return an empty list of nodes.
+            </t>
+            <t>Syntax errors in the JSONPath SHOULD be detected before matching is attempted since they do not depend on the JSON document.
+               Therefore, an implementation SHOULD take a JSONPath and produce an optional syntax error and then,
+               if and only if an error was not produced, SHOULD take a JSON document and produce a list of matches or an error (but not both).
+            </t>
+            <t>Alternatively, an implementation MAY take a JSONPath and a JSON document and produce a list of matches or an optional error (but not both).
+            </t>
+            <t>For any implementation, if a syntactically invalid JSONPath is provided, the implementation MUST return an error.
+            </t>
+            <t>If a syntactially invalid JSON document is provided, any implementation SHOULD return an error.
+            </t>
+         </section>
+         <section numbered="true" toc="default">
+            <name>Syntax</name>
+            <t>Syntactically, a JSONPath consists of a root selector (<tt>$</tt>), which selects the root node of a JSON document, followed by a possibly empty sequence of <em>matchers</em>.
+            </t>
+            <sourcecode type="abnf">
+json-path = root-selector *matcher
+root-selector = %x24               ; $ selects document root node
+            </sourcecode>
+            <t>The syntax and semantics of each matcher is defined below.
+            </t>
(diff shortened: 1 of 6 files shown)


draft-normington-jsonpath-latest.xml
```diff
@@ -104,7 +107,40 @@
       </section>
     </section>
    <section numbered="true" toc="default">
-      <name>JSONPath Syntax and Semantics</name>         
+      <name>JSONPath Syntax and Semantics</name>
+         <section numbered="true" toc="default">
+            <name>Node Matching Semantics</name>
+            <t>A JSONPath is a string which matches zero or more nodes of any JSON document.
+            </t>
+            <t>A JSONPath is abstractly a sequence of matchers. Each matcher takes a list of nodes and produces a possibly different list of nodes.
```
remorhaz, line 115: > A JSONPath is abstractly a sequence of matchers.

Things getting not that simple if we support subpaths and/or nested filters, and we have [a consensus](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_subpaths.html) here.

draft-normington-jsonpath-latest.xml
```diff
@@ -104,7 +107,40 @@
...
+               The initial list of nodes, consisting of just the root node of a given JSON document, is passed to the first matcher of the sequence.
+               The output from each matcher is passed to the next matcher in the sequence, and so forth until the end of the sequence is reached.
+               The list of nodes produced by the last matcher is the result of matching the JSONPath to the given JSON document.
+            </t>
+            <t>If the JSONPath is empty, this correspondds to an empty sequence of matchers and matches the root node of the JSON document.
```
remorhaz, line 120: > correspondds

A typo.

draft-normington-jsonpath-latest.xml
glyn: @remorhaz I'm not sure I follow. If we take that example `$[?(@.address.city=='Berlin')]`, that will break down into two matchers:
* `$` which matches the root document
*  `[?(@.address.city=='Berlin')]` which filters the array to pick out the "Berlin" element.
Some of the matchers will be defined in terms of others, but even so each JSONPath will correspond to a sequence of "top-level" matchers.

draft-normington-jsonpath-latest.xml
glyn: Fixed, thanks.

draft-normington-jsonpath-latest.xml
remorhaz: Well, if we don't speak about "inner" matchers here, then I agree with the definition.

glyn:
I've thought of some clarifications and other things that should be said initially, so changed this PR to draft while I work on it some more. I'd also like to be able to motivate things better to address the kinds of question that @remorhaz raised.

glyn:
@remorhaz I've done some rework and aligned the syntax (such as it is) more closely with Proposal A. I've also started to talk about implementations. Please could you take a look and provide feedback. If it's helpful, here's a [rendered HTML version](https://glyn.github.io/internet-draft/).

draft-normington-jsonpath-latest.xml
```diff
@@ -109,37 +109,81 @@
...
                </t>
-               <sourcecode type="abnf">
-path = root subpath ; 
-root = %x24         ; $ matches the root node of a document
-
-subpath = identity  ; see below for alternatives
-identity = ""       ; matches all input nodes
-               </sourcecode>
-               <t>Valid JSONPaths MUST conform to the grammar of <tt>path</tt>.
+               <t>An implementation which produces a scalar value or an error MUST produce a scalar value only when the list of matching nodes comprises a
```
remorhaz, line 138: > when the list of matching nodes comprises a single node
This approach can cause many troubles, I think.

Look, there are two cases:
1. Query is able to match zero or one node. It is an _attribute_ of a query, and it can be calculated using just query text, without any JSON data: all matchers in chain should match one or zero nodes (like `.foo') to tag query as "single-matching".
2. Query is able to match arbitrary amount of nodes. It's an attribute of a query, too, and it's enough to have a single "multi-matching" matcher (like `*`) in chain to tag all the query as "multi-matching".

But on some JSON data multi-matching query will match exactly one node, which makes it possible to behave like single-matching query 1; I mean, node list with a single node can be reduced to a single node.

Query attributes can be used for analysis by client software to ensure that given query is garanteed to match either exactly one either none of nodes. Making the type of returned data dependent on input is dangerous and complicated: well, okay, we've got an array here - is this a single node that is an array or is this a list of matched nodes? I see no way to distinguish.

I suggest not to mix "naturally single-matching queries" with "occasionally single-matching queries".

glyn: Intersting. Let's analyse this a bit. The nub of the problem seems to be:
> Making the type of returned data dependent on input is dangerous and complicated: well, okay, we've got an array here - is this a single node that is an array or is this a list of matched nodes? I see no way to distinguish.

The output type isn't data (JSONPath or input document) dependent _for a given implementation_, but it would be implementation dependent.

Let's look at both implementation options.

Firstly, the preferred (SHOULD) approach of always returning a list of nodes or error. We can overlook the error case as that's unambiguous. If a single node matched, then the output would be a list comprising that node. There is no ambiguity. The "outermost" list always represents the fact the implementation is returning a list of nodes.

Secondly, the non-preferred "scalar" approach of always returning a scalar or error. If a single node matched which contains a scalar value, that value is returned. Again, I don't think there is any ambiguity.

remorhaz: But how should "the non-preferred _scalar_ approach" behave if query matches several nodes on concrete data? If it falls back on "preferred approach", the result can be ambiguous. If it doesn't, we should define the behavior better: it can return first node from list, or return an error, or whatever.

remorhaz: Ah, I missed the point: you want to allow "single" mode only for scalar values. That indeed removes ambiguity. But in the same time, the non-preferred approach still seems very wierd:

- `$.foo` on `{"foo":"bar"}` returns `"bar"`.
- `$.foo` on `{"foo":["bar","baz"]}` returns `[["bar","baz"]]` - isn't it weird? Or it should produce an error?

What do you think about replacing "non-preferred approach" with optional _reduce_ operation on "preferred approach" results?

glyn: > Ah, I missed the point: you want to allow "single" mode only for scalar values. That indeed removes ambiguity. But in the same time, the non-preferred approach still seems very wierd:
> 
> * `$.foo` on `{"foo":"bar"}` returns `"bar"`.

Correct.

> * `$.foo` on `{"foo":["bar","baz"]}` returns `[["bar","baz"]]` - isn't it weird? Or it should produce an error?

It should return an error since `[["bar","baz"]]` isn't a "list of matching nodes comprises a single node whose value is [a] scalar value".

> 
> What do you think about replacing "non-preferred approach" with optional _reduce_ operation on "preferred approach" results?

I like the idea of framing the behaviour in terms of _reduce_, but I'm not sure that would provide the "escape route" that I was aiming for. You see I don't personally think the non-preferred approach is a good option, but I suggest it so that implementations which behave this way today could become spec compliant without necessarily completely breaking all their users.

In fact, I put up the current non-preferred approach as a strawman for discussion and I'd be interested in others' opinions (@cburgmer, @gregsdennis, @Cleidiano, and other potential reviewers).

Another, more radical option, would be to drop the non-preferred approach altogether and leave it to implementations to define their behaviour _in terms of the standard_ should they choose to do so.

Another intermediate option would be to add some non-normative text explaining how to return a scalar in terms of the standard behaviour. I quite like that approach as it would at least help "scalar implementations" to adopt a similar approach.

Avoiding a normative definition of scalar implementations would also get us out of the difficulty of covering that use case in a future compliance test suite.

remorhaz: > I like the idea of framing the behaviour in terms of reduce, but I'm not sure that would provide the "escape route" that I was aiming for.

Well, it's a big question if existing questions conform to this "escape route", anyway. Just try execute `$.store` [here](https://jsonpath.herokuapp.com/) on Goessner's example - and you'll get a single JSON object. Execute `$.store.book` - and get a single array (exactly what I've meant by "ambiguous"). As for me, I'd vote for the most radical option of dropping the non-preferred approach.

Reduction is a needful operation on practice, though: there are tons of cases when we may want to convert result to exactly one node or error. But I'm not sure if we should bother about such an operation in this document: given a list of nodes there's no problem to implement such a behavior even if some implementation doesn't provide it out of box.

glyn: I agree. So let's drop it for now and add some _non-normative_ text later if there's any pressure to do so. I'll force push in a few minutes...

draft-normington-jsonpath-latest.xml
```diff
@@ -104,7 +107,75 @@
...
+            </t>
+            <t>Syntax errors in the JSONPath SHOULD be detected before matching is attempted since they do not depend on the JSON document.
+               Therefore, an implementation SHOULD take a JSONPath and produce an optional syntax error and then,
+               if and only if an error was not produced, SHOULD take a JSON document and produce a list of matches or an error (but not both).
+            </t>
+            <t>Alternatively, an implementation MAY take a JSONPath and a JSON document and produce a list of matches or an optional error (but not both).
+            </t>
+            <t>For any implementation, if a syntactically invalid JSONPath is provided, the implementation MUST return an error.
+            </t>
+            <t>If a syntactially invalid JSON document is provided, an implementation MAY return an error or MAY return an empty list of matches.
```
remorhaz, line 131: > MAY return an error or MAY return an empty list of matches

What is the rationale behind letting the implementation to decide?

draft-normington-jsonpath-latest.xml
```diff
@@ -104,7 +107,75 @@
...
+            <sourcecode type="abnf">
+jsonpath = root *matcher
+root = %x24              ; $ matches the root node of a document
+            </sourcecode>
+            <t>The syntax and semantics of each matcher is defined below.
+            </t>
+         </section>
+         <section numbered="true" toc="default">
+            <name>Semantics</name>
+            <t><tt>root</tt> not only matches the root node of the input document, but it produces as output a list consisting of one node: the input document.
```
remorhaz, line 147: > matches the root node

If `$` is a matcher, then it can appear at any position in chain, `$[1]$`, for example. Maybe it's better to define `$` and `@` as _selectors_ and a JsonPath query as selector, followed by matchers sequence (possibly empty)?

draft-normington-jsonpath-latest.xml
glyn: I had in mind implementations, like my own, with an API consisting of:
* a constructor function which takes a JSONPath string and returns a rich JSONPath type or a syntax error. This gets syntax errors out of the way "early" (i.e. before any JSON documents are encountered) and allows "compilation" of the JSONPath to be performed once before the compiled version is used, possibly many times, to search JSON documents.
* a method on the JSONPath type which takes a _parsed_ form of a JSON document and which returns the search results or an error.

But you've identified a mistake there. Such implementations will _never_ encounter a syntactically invalid JSON document, so they needn't worry about how to handle syntax errors.

So I'm now leaning towards a stricter requirement:
> If a syntactically invalid JSON document is provided, an implementation MUST return an error.

WDYT?

draft-normington-jsonpath-latest.xml
glyn: Yes, I was trying to avoid describing `$` as a matcher for the reason you point out. I guess saying that `$` _matches_ the root node gives the impression that `$` is a matcher (regardless of what the ABNF might say). Can you think of a better wording? If not, I'll try again...

glyn: > Maybe it's better to define `$` and `@` as _selectors_ and a JsonPath query as selector, followed by matchers sequence (possibly empty)?

I was trying to stick pretty closely to the Proposal A grammar. Also, we need to prevent `@` at the start of a JSONPatth. But your point is well made and will come in useful once we get to filters. Do we really need to do more right now?

draft-normington-jsonpath-latest.xml
remorhaz: Yes, I think that's better.

draft-normington-jsonpath-latest.xml
remorhaz: > we need to prevent @ at the start of a JSONPath

In fact, we do not. Semantic of `@` is "current node", and current node at the start of JsonPath is obviously a root node, so there's no practical reason to avoid `@` at start.

But it's, I'd say, a brave movement and should be discussed more thoroughly. For now, I believe, it's enough to choose between "sequence of matchers" and "selector followed by optional sequence of matchers".

draft-normington-jsonpath-latest.xml
glyn: Ok. I'm now wondering how strict to be about checking the JSON document syntax because it's not really the prime responsibility of a JSONPath implementation. In particular, I don't want to over-burden JSONPath implementations in languages which happen to have imperfect JSON parsing libraries. So how about we weaken the wording to:
> If a syntactically invalid JSON document is provided, any implementation SHOULD return an error.

?

draft-normington-jsonpath-latest.xml
glyn: I'd like us to adopt an approach for the first version of the internet draft of avoiding new constructs whenever possible as that will increase the migration cost for existing implementations and users. So I think in this PR we should stick with `$` followed by a possibly empty sequence of matchers. I'll reword the above text to avoid calling `$` a matcher and use the "selector" terminology instead, which we can then piggy-back upon nicely when we come to filters. Let me draft that and push the change for your perusal.

draft-normington-jsonpath-latest.xml
remorhaz: Well, in fact many implementations will work without JSON text itself, using the result of JSON parsing as a structure; or even without JSON at all (yaml, for example). XML has DOM, that allows to isolate document structure from text source (and syntax); JSON lacks such tool.

Maybe we should just define JSON data structure as a tree in terms of typed and (partially) ordered nodes, thus totally isolating JsonPath from JSON parsing issues; but for now SHOULD is fairly enough.

draft-normington-jsonpath-latest.xml
remorhaz: > avoiding new constructs whenever possible

That's not a problem - we can define selectors with the following grammar:

```
selector       = root_selector / other_selector
root_selector  = "$"
other_selector = "@"
```

And then define queries like this, avoiding:

```
root_path   = root_selector [ matcher_list ]
filter_path = selector [ matcher_list ]
```

It's not the concrete grammar proposal, just showing an idea how to forbid `@` at the start of outer path.

glyn: Thanks @remorhaz. I've used just what we need of this for the moment. I adopted the convention from the ABNF RFC of spelling compound grammar elements using "-" rather than "_".

PR #10, opened 2020-07-17
union child and index matchers

glyn:
Also define encoding of paths.

Also use JSON Path instead of JSONPath.

See [rendered HTML](https://glyn.github.io/internet-draft/) for ease of reading.

### DIFF
--- a/draft-normington-jsonpath-latest.xml
+++ b/draft-normington-jsonpath-latest.xml
@@ -12,7 +12,7 @@
       submissionType="IETF"
       xml:lang="en"
       tocInclude="true"
-      tocDepth="3"
+      tocDepth="4"
       symRefs="true"
       sortRefs="true"
       version="3">
@@ -74,7 +74,7 @@
         keywords will be used for the search engine. -->
 
    <abstract>
-      <t>JSONPath defines a string syntax for identifying values
+      <t>JSON Path defines a string syntax for identifying values
    within a JavaScript Object Notation (JSON) document.</t>
     </abstract>
     <note>
@@ -88,7 +88,7 @@
     <section numbered="true" toc="default">
       <name>Introduction</name>
       <!-- move the following to an Acknowledgements appendix and write a proper introduction -->
-      <t>JSONPath was introduced by Stefan Goessner as a simple form of XPath for JSON.
+      <t>JSON Path, or rather JSONPath, was introduced by Stefan Goessner as a simple form of XPath for JSON.
          See his original article <xref target="Goessner" format="default">Goessner</xref>.
       </t>
       <t>JSON is defined by <xref target="RFC8259" format="default">RFC&nbsp;8259</xref>.
@@ -103,14 +103,21 @@
 
       <section numbered="true" toc="default">
         <name>ABNF Syntax</name>
-        <t>The syntax in this document conforms to ABNF as defined by <xref target="RFC5234" format="default">RFC&nbsp;5234</xref>.</t>
+        <t>The syntax in this document conforms to ABNF as defined by <xref target="RFC5234" format="default">RFC&nbsp;5234</xref>.
+        </t>
+        <t>ABNF terminal values in this document define Unicode code points rather than their UTF-8 encoding.
+           For example, the Unicode PLACE OF INTEREST SIGN (U+2318) would be defined in ABNF as <tt>%x2318</tt>.
+        </t>
       </section>
     </section>
    <section numbered="true" toc="default">
-      <name>JSONPath Syntax and Semantics</name>
+      <name>JSON Path Syntax and Semantics</name>
          <section numbered="true" toc="default">
             <name>Overview</name>
-            <t>A JSONPath is a string which matches zero or more nodes of any JSON document. A valid JSONPath conforms to the ABNF syntax defined by this document.
+            <t>A JSON Path is a string which matches zero or more nodes of any JSON document. A valid JSON Path conforms to the ABNF syntax defined by this document.
+            </t>
+            <t>A JSON Path MUST be encoded using UTF-8. To parse a JSON Path according to the grammar in this document, its UTF-8 form SHOULD first be decoded into Unicode code points as described
+               in <xref target="RFC3629" format="default">RFC&nbsp;3629</xref>.
             </t>
          </section>
          <section numbered="true" toc="default">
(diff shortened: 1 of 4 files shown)


draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+union = %x5B ws union-elements ws %x5D ; [...]
+ws = *%x20                             ; zero or more spaces
+union-elements = union-element /
+                 union-element ws %x2C ws union-elements
+                                       ; ,-separated list
+                  </sourcecode>
+               </section>       
+               <section numbered="true" toc="default">
+                  <name>Semantics</name>
+                  <t>A union matches any node which is matched by at least one of the union elements and selects the concatenation of the
```
remorhaz, line 245: > by at least one of the union

It is a question if union should "merge" union results or "concatenate" the results. We [have a consensus](https://cburgmer.github.io/json-path-comparison/results/union_with_numbers_in_decreasing_order.html) for reverse ordered indexes, that shows us that we should define the order of merge/concatenation; but the situation of fetching the same index twice (like `$[1,1]`) is still not investigated.

But this question can be discussed later.

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+               <section numbered="true" toc="default" anchor="unionchild">
+                  <name>Union Child</name>
+                  <section numbered="false" toc="exclude">
+                     <name>Syntax</name>
+                     <t>A union child is a union element consisting of a quoted string.
+                     </t>
+                     <sourcecode type="abnf">
+union-element = union-child ; see below for more alternatives
+union-child = %x27 *single-quoted %x27 / ; 'string'
+              %x22 *double-quoted %x22   ; "string"
```
remorhaz, line 258: Maybe this one should refer to JSON string definition as of [RFC 8259.7](https://tools.ietf.org/html/rfc8259#section-7)? Otherwise we can get duplicated non-terminals if we allow JSON structures in filters.

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+
+single-quoted = %x5C %x27 / ; \' (escaped single quote)
+                %x5C %x5C / ; \\ (escaped reverse solidus)
+                %x00-26 / %x28-5B %x5D-D7FF / %xE000-10FFFF
```
remorhaz, line 262: Probably slash omitted after `5B`?

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+                %x00-21 / %x23-5B / %x5D-D7FF / %xE000-10FFFF
+                            ; any Unicode code point except " or \
+                     </sourcecode>
+                     <t>Note: the syntax does not allow any character other than a quote or reverse solidus to be escaped.
+                     </t>
+                  </section>       
+                  <section numbered="false" toc="exclude">
+                     <name>Semantics</name>
+                     <t>If the union child is a quoted string, the string must first have the surrounding quotes removed and then be unescaped
+                        as describe below.
```
remorhaz, line 276: Probably "as described"?

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+                        The union child matches any node which is an object having a key equal to the unescaped value of the union child string
+                        and selects the value corresponding to that key.
+                        It does not match a node which is not an object.
+                     </t>
+                     <t>After the surrounding quotes have been removed, a string is unescaped by replacing each escaped quote by the corresponding
```
remorhaz, line 281: Maybe it's better to fully describe the unescaping process at the beginning of the section without splitting it with matching description? So that at first we describe how do we _get_ the key and then - how we _use_ the key.

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+               </section>             
+               <section numbered="true" toc="default">
+                  <name>Array Index</name>
+                  <section numbered="false" toc="exclude">
+                     <name>Syntax</name>
+                     <t>An array index is a union element consisting of an integer (in base 10).
+                     </t>
+                     <sourcecode type="abnf">
+union-element =/ integer
+integer = [%x2D] (%x30 / (%x31-39 *%x30-39))
```
remorhaz, line 294: Once again, maybe it's better to refer to JSON standard that [defines](https://tools.ietf.org/html/rfc8259#section-6) `int` non-terminal?

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+                            ; optional - followed by 0 or
+                            ; sequence of digits with no leading zero
+                     </sourcecode>
+                     <t>Note: the syntax does not allow integers with leading zeros such as <tt>01</tt> and <tt>-01</tt>.
+                     </t>
+                  </section>       
+                  <section numbered="false" toc="exclude">
+                     <name>Semantics</name>
+                     <t>An array index matches any node which is an array and selects any element of the array indexed by the integer index.
```
remorhaz, line 303: Maybe we should avoid defining index at all and rather consider it as a special case of a slice (when onbly start index is given)?

draft-normington-jsonpath-latest.xml
```diff
@@ -218,6 +225,94 @@ dot-child-name = TBD
...
+                        It does not match a node which is not an array.
```
remorhaz, line 304: > It does not match

Maybe `it MUST not match`?

draft-normington-jsonpath-latest.xml
glyn: @remorhaz I suggest you raise an issue on this so we don't forget it.

draft-normington-jsonpath-latest.xml
glyn: Nice catch.

draft-normington-jsonpath-latest.xml
glyn: Actually, I'm inclined to switch to RFC 8259 syntax for strings for consistency alone. But I think we need to allow for single-quoted as well as double-quoted strings, for example like this:
```
union-child = %x22 *double-quoted %x22 / ; "string"
              %x27 *single-quoted %x27   ; 'string'

double-quoted = dq-unescaped /
          escape (
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX


      dq-unescaped = %x20-21 / %x23-5B / %x5D-10FFFF

single-quoted = sq-unescaped /
          escape (
              %x27 /          ; '    apostrophe      U+0027
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX

      sq-unescaped = %x20-26 / %x28-5B / %x5D-10FFFF

escape = %x5C                 ; \
```
@cburgmer this is the first time we would be deviating from the syntax in Proposal A. WDYT?

draft-normington-jsonpath-latest.xml
glyn: Yes.

draft-normington-jsonpath-latest.xml
glyn: Good idea.

draft-normington-jsonpath-latest.xml
glyn: Agreed, although I think the syntaxes are equivalent in this case.

draft-normington-jsonpath-latest.xml
glyn: I wanted to follow Proposal A and keep the grammar productions for index and slice separate because that approach neatly avoids some the issues by distinguishing between union elements which can select multiple elements and those which can select at most one element. But we may be able to organise the document to keep these definitions close to each other.

draft-normington-jsonpath-latest.xml
glyn: Probably. I haven't been consistent in applying MUST etc. and it would certainly be an improvement.

draft-normington-jsonpath-latest.xml
remorhaz: > I think we need to allow for single-quoted as well as double-quoted strings

Of course, but that can be done like this:
```
union-child = %x27 *single-quoted %x27 /  ; 'string'
              json-string ; see RFC blah-blah-blah
```
I mean, JSON string already includes double-quotes in its definition.

glyn: I think it would be better to keep the ABNF self-contained:
- It's easier to read without jumping over to another RFC. (We could make the production names the same if that helps and have a comment "copied from RFC nnnn")
- I'd like our text to be immutable, but the ABNF in another RFC could mutate because of errata (which we wouldn't _necessarily_ want to adopt, at least not without our own errata)
- it should be possible to extract the ABNF and generate a lexer/parser from it.

glyn: @cabo Have you any general advice about whether it is best to inline or refer to ABNF from existing RFCs?

cabo: Funny, I just answered the same question on a different list :-)

Generally, right now, it is better to copy, because there is no actionable way to refer.
Until that get's unwieldy, so if you need the date-time production from RFC 3339, that carries so much baggage you'd better refer.

We are solving the same problem for CDDL right now in the CBOR WG, so maybe we'll have something that will spill over to ABNF as wel.  For now, copy over refer, I'd say.

glyn: I'm going to stick with the copy of the ABNF and also preserve the symmetry between single and double quoted strings as I think that's a lot more readable.

draft-normington-jsonpath-latest.xml
glyn: I'm going to fix this in a later PR by making a pass through the the whole document.

draft-normington-jsonpath-latest.xml
glyn: Actually, I've decided not to change this because the JSON standard doesn't provide a suitable signed integer production. We can revisit this when float literals make their appearance.

PR #11, opened 2020-07-28
union slices

glyn:
See the [rendered HTML](https://glyn.github.io/internet-draft/) version of this PR.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1246,55 +1246,55 @@ dt+dd:empty::before{
         <a href="#" onclick="scroll(0,0)" class="toplink">▲</a><h2 id="name-table-of-contents-2">
 <a href="#name-table-of-contents-2" class="section-name selfRef">Table of Contents</a>
         </h2>
-<nav class="toc"><ul class="compact toc ulEmpty">
-<li class="compact toc ulEmpty" id="section-toc.1-1.1">
+<nav class="toc"><ul class="toc ulEmpty compact">
+<li class="toc ulEmpty compact" id="section-toc.1-1.1">
             <p id="section-toc.1-1.1.1" class="keepWithNext"><a href="#section-1" class="xref">1</a>.  <a href="#name-introduction-2" class="xref">Introduction</a><a href="#section-toc.1-1.1.1" class="pilcrow">¶</a></p>
-<ul class="compact toc ulEmpty">
-<li class="compact toc ulEmpty" id="section-toc.1-1.1.2.1">
+<ul class="toc ulEmpty compact">
+<li class="toc ulEmpty compact" id="section-toc.1-1.1.2.1">
                 <p id="section-toc.1-1.1.2.1.1" class="keepWithNext"><a href="#section-1.1" class="xref">1.1</a>.  <a href="#name-requirements-language-2" class="xref">Requirements Language</a><a href="#section-toc.1-1.1.2.1.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.1.2.2">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.1.2.2">
                 <p id="section-toc.1-1.1.2.2.1" class="keepWithNext"><a href="#section-1.2" class="xref">1.2</a>.  <a href="#name-abnf-syntax-2" class="xref">ABNF Syntax</a><a href="#section-toc.1-1.1.2.2.1" class="pilcrow">¶</a></p>
 </li>
             </ul>
 </li>
-          <li class="compact toc ulEmpty" id="section-toc.1-1.2">
+          <li class="toc ulEmpty compact" id="section-toc.1-1.2">
             <p id="section-toc.1-1.2.1"><a href="#section-2" class="xref">2</a>.  <a href="#name-json-path-syntax-and-semanti" class="xref">JSON Path Syntax and Semantics</a><a href="#section-toc.1-1.2.1" class="pilcrow">¶</a></p>
-<ul class="compact toc ulEmpty">
-<li class="compact toc ulEmpty" id="section-toc.1-1.2.2.1">
+<ul class="toc ulEmpty compact">
+<li class="toc ulEmpty compact" id="section-toc.1-1.2.2.1">
                 <p id="section-toc.1-1.2.2.1.1"><a href="#section-2.1" class="xref">2.1</a>.  <a href="#name-overview-2" class="xref">Overview</a><a href="#section-toc.1-1.2.2.1.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.2.2.2">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.2.2.2">
                 <p id="section-toc.1-1.2.2.2.1"><a href="#section-2.2" class="xref">2.2</a>.  <a href="#name-terminology-2" class="xref">Terminology</a><a href="#section-toc.1-1.2.2.2.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.2.2.3">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.2.2.3">
                 <p id="section-toc.1-1.2.2.3.1"><a href="#section-2.3" class="xref">2.3</a>.  <a href="#name-implementation-2" class="xref">Implementation</a><a href="#section-toc.1-1.2.2.3.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.2.2.4">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.2.2.4">
                 <p id="section-toc.1-1.2.2.4.1"><a href="#section-2.4" class="xref">2.4</a>.  <a href="#name-syntax-6" class="xref">Syntax</a><a href="#section-toc.1-1.2.2.4.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.2.2.5">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.2.2.5">
                 <p id="section-toc.1-1.2.2.5.1"><a href="#section-2.5" class="xref">2.5</a>.  <a href="#name-semantics-6" class="xref">Semantics</a><a href="#section-toc.1-1.2.2.5.1" class="pilcrow">¶</a></p>
 </li>
-              <li class="compact toc ulEmpty" id="section-toc.1-1.2.2.6">
+              <li class="toc ulEmpty compact" id="section-toc.1-1.2.2.6">
                 <p id="section-toc.1-1.2.2.6.1"><a href="#section-2.6" class="xref">2.6</a>.  <a href="#name-matchers-2" class="xref">Matchers</a><a href="#section-toc.1-1.2.2.6.1" class="pilcrow">¶</a></p>
-<ul class="compact toc ulEmpty">
-<li class="compact toc ulEmpty" id="section-toc.1-1.2.2.6.2.1">
+<ul class="toc ulEmpty compact">
+<li class="toc ulEmpty compact" id="section-toc.1-1.2.2.6.2.1">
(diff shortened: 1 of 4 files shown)


glyn:
@cabo fyi I still need to make an editorial pass to use the terms MUST, SHOULD, etc. consistently. But you may care to skim read this version to get a feel for where we are at.

draft-normington-jsonpath-latest.txt
```diff
@@ -205,9 +205,14 @@ Internet-Draft   JavaScript Object Notation (JSON) Path        July 2020
...
 
    A matcher may match a node and, if it matches, may then select zero
-   or more descendants of the node for further processing.  But there's
-   more to it than this: each matcher acts on a list of nodes and
-   produces a list of nodes, as follows.
+   or more descendants of the node for further processing.  A
+   syntactically valid matcher MUST NOT produce errors and so some
+   operations which might be considered erroneous, such as indexing
+   beyond the end of an array, simply result in fewer descendants being
+   selected.
```
jchesterpivotal, line 212: I suggest breaking this into two sentences. "A syntactically valid matcher MUST NOT produce errors. This means that some operations which might be considered erroneous, such as indexing beyond the end of an array, result in fewer descendants being selected".

draft-normington-jsonpath-latest.txt
```diff
@@ -424,20 +424,20 @@ Internet-Draft   JavaScript Object Notation (JSON) Path        July 2020
    selects the value corresponding to the key.  It does not match a node
    which is not an object.
 
-2.6.2.4.  Array Index
+2.6.2.4.  Array Access
 
 Syntax
 
+   Array access is a union element which matches any node which is an
+   array and which selects zero or more elements of the array.  Array
```
jchesterpivotal, line 432: I found this difficult to follow. Perhaps something like "Array access is a union element which matches zero or more elements of an array node."

draft-normington-jsonpath-latest.txt
```diff
@@ -464,7 +487,52 @@ Semantics
 
    A negative index "j" selects an element of an array of length "len"
    if and only if "0 <= j + len < len", in which case it selects the
-   same element as selected by the non-negative index "j + len".
+   same element as the non-negative index "j + len".
+
+   An array slice matches any node which is an array and selects any
+   elements of the array indexed by the members of the slice, as defined
```
jchesterpivotal, line 493: Same comment as above, this is a little difficult for me to follow.

draft-normington-jsonpath-latest.txt
```diff
@@ -464,7 +487,52 @@ Semantics
...
+   below, in the same order as the slice.  It does not match a node
+   which is not an array.
```
jchesterpivotal, line 495: Should "does not match a node which is not an array" appear at the top of the section with the introduction about Array Access?

draft-normington-jsonpath-latest.txt
glyn: Yes, I like that thanks.

docs/index.html
```diff
@@ -1556,23 +1558,44 @@ <h6 id="name-semantics-9">
...
+            <h5 id="name-array-access-2">
+<a href="#section-2.6.2.4" class="section-number selfRef">2.6.2.4. </a><a href="#name-array-access-2" class="section-name selfRef">Array Access</a>
             </h5>
 <section id="section-2.6.2.4.1">
               <h6 id="name-syntax-10">
 <a href="#name-syntax-10" class="section-name selfRef">Syntax</a>
               </h6>
-<p id="section-2.6.2.4.1-1">An array index is a union element consisting of an integer (in base 10).<a href="#section-2.6.2.4.1-1" class="pilcrow">¶</a></p>
+<p id="section-2.6.2.4.1-1">Array access is a union element which matches any node which is an array and which selects zero or more elements of the array.
+                        Array access takes the form of an index, which selects at most one element, or a slice, which selects zero or more elements.<a href="#section-2.6.2.4.1-1" class="pilcrow">¶</a></p>
```
Cleidiano, line 1569: The expression like `$[0, 1]` will not be included in the spec?

draft-normington-jsonpath-latest.txt
glyn: I'm using the terminology of a matcher _matching_ a node and then _selecting_ subnodes. There's a clear distinction for example with "dot child" where the key matches the key in an object but the matcher selects the value subnode. In the array case, we can probably get away with shortening this. Let's see how that pans out.

draft-normington-jsonpath-latest.txt
glyn: Yes. I was trying to avoid duplication in the syntax and semantics section, but I think it reads better with a bit of repetition.

When my wife was a maths teacher, she used to tell me "You only learn by repetition. You only learn by repetition."

draft-normington-jsonpath-latest.txt
glyn: Tried a rewording. See what you think.

docs/index.html
glyn: Yes it will. According to the syntax definition `[0, 1]` is a union of two array accesses.

Cleidiano: Oh, I forgot that index access is a type of union element, sorry.

glyn: No problem. If others make the same mistake, we may need to think about how to address that in the document.

(I am trying to avoid adding numerous examples, which might be one fix, as that can weaken a standards document because less attention is paid to the normative text, by both readers and authors.)

docs/index.html
```diff
@@ -1354,24 +1354,26 @@ <h3 id="name-semantics-6">
...
+               A syntactically valid matcher MUST NOT produce errors. This means that some operations which might be considered erroneous, such as indexing beyond the end of an array,
+               simply result in fewer descendants being selected.<a href="#section-2.5-2" class="pilcrow">¶</a></p>
+<p id="section-2.5-3">But a matcher doesn't just act on a single node: each matcher acts on a list of nodes and produces a list of nodes, as follows.<a href="#section-2.5-3" class="pilcrow">¶</a></p>
+<p id="section-2.5-4">After the root selector, the remainder of the JSON Path is processed by passing lists of nodes from one matcher to the next ending up with a list of nodes which is the result of
+               matching the JSON Path to the input JSON document.<a href="#section-2.5-4" class="pilcrow">¶</a></p>
+<p id="section-2.5-5">Each matcher acts on its input list of nodes as follows. For each node in the list, the matcher may or may not match the node. If the matcher
                matches the node, it selects zero or more descendants of the node.
```
remorhaz, line 1363: > zero or more descendants of the node

Are we sure we're not going to allow nodes from input list to get included in the output list (like `descendant-or-self::` axis in XPath)?

glyn: Ah, I think you missed the following in the Terminology section 2.2:
> The descendants of a node are taken to include the node itself and the descendants of any children of the node.

docs/index.html
```diff
@@ -1354,24 +1354,26 @@ <h3 id="name-semantics-6">
...
-<p id="section-2.5-10">Finally, <code>.b</code> matches any input node of type object and selects any value of the input node corresponding to the key <code>"b"</code>.
-               The result is the concatenation of three lists each of length one containing <code>0</code>, <code>1</code>, <code>2</code>, respectively.<a href="#section-2.5-10" class="pilcrow">¶</a></p>
-<p id="section-2.5-11">As a consequence of this approach, if any of the matchers matches no nodes, then the whole JSON Path matches no nodes.<a href="#section-2.5-11" class="pilcrow">¶</a></p>
-<p id="section-2.5-12">In what follows, the semantics of each matcher are defined for each type of node.<a href="#section-2.5-12" class="pilcrow">¶</a></p>
+               The output list of nodes of a matcher is the concatenation of the lists of selected descendents for each input node.<a href="#section-2.5-5" class="pilcrow">¶</a></p>
+<p id="section-2.5-6">A specific, non-normative example will make this clearer. Suppose the input document is: <code>{"a":[{"b":0},{"b":1},{"b":2}]}</code>. As we will see later, the JSON Path
```
remorhaz, line 1365: Maybe we should define the last element of array as `{"c":2}` - just to demonstrate, how non-matching node is filtered out?

docs/index.html
```diff
@@ -1354,24 +1354,26 @@ <h3 id="name-semantics-6">
...
+<p id="section-2.5-9">Next, <code>.a</code> matches any input node of type object and selects any value of the input node corresponding to the key <code>"a"</code>.
+               The result is again a list of one node: <code>[{"b":0},{"b":1},{"b":2}]</code>.<a href="#section-2.5-9" class="pilcrow">¶</a></p>
+<p id="section-2.5-10">Next, <code>[*]</code> matches any input node which is an array and selects all the elements of the input node.
+               The result is a list of three nodes: <code>{"b":0}</code>, <code>{"b":1}</code>, and <code>{"b":2}</code>.<a href="#section-2.5-10" class="pilcrow">¶</a></p>
+<p id="section-2.5-11">Finally, <code>.b</code> matches any input node of type object and selects any value of the input node corresponding to the key <code>"b"</code>.
+               The result is the concatenation of three lists each of length one containing <code>0</code>, <code>1</code>, <code>2</code>, respectively.<a href="#section-2.5-11" class="pilcrow">¶</a></p>
```
remorhaz, line 1374: > result is the concatenation

I think we should explicitly define here (or somewhere else) what happens when concatenated lists contain same node (we have unions, so it's absolutely possible).

docs/index.html
```diff
@@ -1556,35 +1558,87 @@ <h6 id="name-semantics-9">
...
+array-slice = [ integer ] ws %x3A ws [ integer ]
+                   [ ws %x3A ws non-zero-integer ]
+                            ; start:end or start:end:step
+
+non-zero-integer = [%x2D] %x31-39 *%x30-39
+                            ; optional - followed by
+                            ; sequence of digits with no leading zero
+</pre><a href="#section-2.6.2.4.1-7" class="pilcrow">¶</a>
+</div>
+<p id="section-2.6.2.4.1-8">Note: if the third integer (the <code>step</code> value) is present, it MUST be non-zero.<a href="#section-2.6.2.4.1-8" class="pilcrow">¶</a></p>
```
remorhaz, line 1599: > it MUST be non-zero

I'm not sure if we should make zero step illegal; maybe it's better to define that array-slice with zero step just matches nothing? I don't feel that this situation is worth complicationg syntax / throwing a syntax error; it's just like selecting an index out of range - so why we would want to especially handle this situation?

Another argument is query templates: user may use something like `sprintf("$.a[%d:%d:%d]", x, y, z)` and it may be quite surprising to get syntax errors on zero `z` and not getting it on any other value.

docs/index.html
remorhaz: Yep. But in this case it seems that the term "descendant" doesn't fit very well. What would you think about replacing it with "branch"?

glyn: I don't follow why "descendant" doesn't fit well. Please could you elaborate?

("branch" seems like a strange term, unless you have a particular definition in mind.)

docs/index.html
glyn: Added.

docs/index.html
glyn: I'll add a placeholder for now. Do you happen to know if there's a consensus on this?

docs/index.html
glyn: Yes, I take your point. It's too risky to redefine the normal meaning of descendant as people could miss, or forget, the redefinition. I'll try rewording this.

glyn: Done. See what you think.

draft-normington-jsonpath-latest.txt
```diff
@@ -246,16 +246,16 @@ Internet-Draft   JavaScript Object Notation (JSON) Path   September 2020
...
-   "{"b":0}", "{"b":1}", and "{"b":2}".
+   "{"b":0}", "{"b":1}", and "{"c":2}".
 
-   Finally, ".b" matches any input node of type object and selects any
-   value of the input node corresponding to the key ""b"".  The result
-   is the concatenation of three lists each of length one containing
-   "0", "1", "2", respectively.
+   Finally, ".b" matches any input node of type object with a key "b"
+   and selects the value of the input node corresponding to that key.
+   The result is the concatenation of two lists each of length one
```
remorhaz, line 257: In fact, there are three lists concatenated: two of length one and one of length zero.

remorhaz (approved):

draft-normington-jsonpath-latest.txt
glyn: Good catch. Fixed.

PR #13, opened 2020-08-09
Define dot child name

remorhaz:


draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +202,32 @@ root-selector = %x24               ; $ selects document root node
...
             %x2E %x2A             ; .*
-dot-child-name = TBD
+dot-child-name = union-child / ; "key" or 'key' (see below)
+                 unquoted-key
+unquoted-key = 1*(
+                  %x2D / ; -
+                  DIGIT / ; 0-9
+                  ALPHA / ; A-Z / a-z
+                  %x5F / ; _
+                  %x80-10FFFF ; any non-ASCII Unicode character
```
glyn, line 218: I'm not convinced this should include surrogates such as `D800`.

draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +202,32 @@ root-selector = %x24               ; $ selects document root node
...
                </section>       
                <section numbered="false" toc="exclude">
                   <name>Semantics</name>
-                  <t>A dot child name which is not a single asterisk (<tt>*</tt>) matches any node which is an object having a key equal to the dot child name
-                  and selects the value corresponding to that key.
-                  It does not match a node which is not an object.
+                  <t>A dot child name which is not a single asterisk (<tt>*</tt>) is considered to have a key. It matches any node which is an object having a name equal to that key
+                      and selects the value corresponding to that name. It does not match a node which is not an object.
                   </t>
+                   <t>If dot child name is unquoted then its key is equal to the name. For a quoted name the process of extracting the key is defined below.
```
glyn, line 229: "defined below" is a bit risky. Best to describe and link to the specific section.

glyn (changes requested):

draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +202,32 @@ root-selector = %x24               ; $ selects document root node
...
                   </t>
-                  <t>A dot child name corresponds to a key in a JSON object, without the surrounding double quotes (<tt>"</tt>).
+                  <t>A dot child name corresponds to a name in a JSON object. It can be unquoted or surrounded with quotes, either double (<tt>"</tt>) or single (<tt>'</tt>).
                   </t>
                   <sourcecode type="abnf">
 matcher = dot-child             ; see below for alternatives
 dot-child = %x2E dot-child-name / ; .&lt;dot-child-name&gt;
             %x2E %x2A             ; .*
-dot-child-name = TBD
+dot-child-name = union-child / ; "key" or 'key' (see below)
```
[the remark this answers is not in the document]
remorhaz: Well, what I see now is: the majority of implementations support the syntax itself (so we should support it, too, I believe). But there's a difference in a way of decoding the key: some implementations consider quotes as non-special part of the ID ("raw" approach) and some "unquote" the string. The following links show that there's no clear "majority" in the approach, but I think that "unquote" approach is much less surprising than "raw" one.

https://cburgmer.github.io/json-path-comparison/results/dot_notation_with_double_quotes.html
https://cburgmer.github.io/json-path-comparison/results/dot_notation_with_single_quotes.html

draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +202,32 @@ root-selector = %x24               ; $ selects document root node
...
+                 unquoted-key
+unquoted-key = 1*(
+                  %x2D / ; -
+                  DIGIT / ; 0-9
```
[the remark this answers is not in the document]
remorhaz: `DIGIT` and `ALFA` are defined in [appendix B.1 of ABNF RFC](https://tools.ietf.org/html/rfc5234#appendix-B.1), and they are widely used in other RFCs without re-defining. For example, [JSON RFC](https://tools.ietf.org/html/rfc8259#section-6) uses `DIGIT` rule as a part of JSON number definition.

draft-normington-jsonpath-latest.xml
remorhaz: Maybe it's better to move the definition itself up, so the reader have no need to jump to it while reading from the beginning?

draft-normington-jsonpath-latest.xml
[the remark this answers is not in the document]
glyn: I agree that we could adopt that approach, but as we've discussed elsewhere, there isn't a formal way to pull in ABNF definitions from an RFC, so this introduces a formal weakness which I'd personally prefer to avoid.

draft-normington-jsonpath-latest.xml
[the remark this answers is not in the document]
glyn: It's tricky because either way could be surprising. If we adopt the "unquote" approach, then why bother having bracket child notation at all? It would really be superfluous. I'd find that surprising.

Also, both Gössner's implementations adopt the "raw" approach and I feel that we should probably use Gössner to break any "ties" now that we are going to merge into an internet draft which will include him as an author.

draft-normington-jsonpath-latest.xml
remorhaz: I just follow the philosophy of JSON strings here. [Section 7 of JSON RFC](https://tools.ietf.org/html/rfc8259#section-7) defines non-ASCII range as "continuum" of codepoints within Unicode range, and [section 8.2](https://tools.ietf.org/html/rfc8259#section-8.2) explicitly describes the surrogate case.

draft-normington-jsonpath-latest.xml
glyn: Seems reasonable. Why not try that and see if it work outs?

draft-normington-jsonpath-latest.xml
glyn: I really like following JSON string syntax. Since it's an area that will be debated again, I think it would be worth adding some text to this document (like the comment above) to call that out.

draft-normington-jsonpath-latest.xml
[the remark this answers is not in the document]
remorhaz: Okay, I see no problem in redefining some core ABNF rules that we use, I'll do that.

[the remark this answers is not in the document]
glyn: Thanks!

draft-normington-jsonpath-latest.xml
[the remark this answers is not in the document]
remorhaz: > If we adopt the "unquote" approach, then why bother having bracket child notation at all?

Interesting point, very good. But on the other hand, we have no "bracket child" anymore: brackets define union now (and the former "bracket child" is just a special case of singular union) - that's the main reason for having them, I believe.

I feel that if both interpretations are surprising, then we should probably forbid the quotes in dot-child-name at all, bringing some existing queries as a sacrifice to clarity. What do you think?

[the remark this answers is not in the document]
glyn: That seems reasonable to me and it's consistent with the definition of `unquoted-key` which already prohibits single and double quotes from appearing anywhere.

glyn:
@remorhaz I'm soon away until 14 Sept. after which it would be great to be able to merge this. Would you have some time to apply those changes?

remorhaz:
Yep, in a couple of days.

gregsdennis:
Related: #15

draft-normington-jsonpath-latest.xml
remorhaz: We don't need to do that anymore, because we've rejected supporting quoted dot-child-names.

remorhaz:
@glyn, I think we can finally merge this pR.

glyn:
> @glyn, I think we can finally merge this PR.

Thanks. I'll take a look in the next couple of days.

glyn:
@remorhaz When I run scripts/gen.sh or scripts/docker-gen.sh I get some additional changes to the generated HTML. I'm running xml2rfc v3.1.1 on macOS. Did you forget to regenerate or are our versions of xml2rfc behaving differently?

draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +216,34 @@ root-selector = %x24               ; $ selects document root node
...
                   </t>
                   <sourcecode type="abnf">
 matcher = dot-child             ; see below for alternatives
 dot-child = %x2E dot-child-name / ; .&lt;dot-child-name&gt;
             %x2E %x2A             ; .*
-dot-child-name = TBD
+dot-child-name = 1*(
+                   %x2D / ; -
+                   %x30-39 / ; 0-9 (core ABNF: DIGIT)
+                   %x41-5A / %x61-7A / ; A-Z / a-z (core ABNF: ALPHA)
```
glyn, line 228: I think it would be neater to define separate rules for `DIGIT` and `ALPHA`. Then we could probably do without the "core ABNF" comments, which are slightly distracting for the general reader.

remorhaz:
> Did you forget to regenerate?

Hmm, I don't think so. I've added an `author` element recently (after all other changes were made) and I see it in target .txt, so the regeneration seems to have been performed.

>  or are our versions of xml2rfc behaving differently?

Maybe, but in dockerfile I just did the same actions that were defined in readme, so that's wierd. What are the changes exactly?

glyn:
> > or are our versions of xml2rfc behaving differently?
> 
> Maybe, but in dockerfile I just did the same actions that were defined in readme, so that's wierd. What are the changes exactly?

See https://github.com/jsonpath-standard/internet-draft/commit/2816a2ce6512c34fb7bca6e50c1c64775f57f6ee. This change implies you are on an old version of xml2rfc:
```
-<meta content="xml2rfc 2.47.0" name="generator">
+<meta content="xml2rfc 3.1.1" name="generator">
```

draft-normington-jsonpath-latest.xml
remorhaz: Done!

remorhaz:
Hmm, let me try to upgrade xml2rpc version in dockerfile...

draft-normington-jsonpath-latest.xml
```diff
@@ -202,24 +216,34 @@ root-selector = %x24               ; $ selects document root node
...
+                   %x5F / ; _
+                   %x80-10FFFF ; any non-ASCII Unicode character
+                 )
                   </sourcecode>
-                  <!-- TODO: define dot-child-name using an identifier like syntax. See https://github.com/cburgmer/json-path-comparison/issues/42 -->
                   <t>More general child names, such as the empty string, are supported by "Union Child" (<xref target="unionchild" format="default"/>).
                   </t>
+                  <t>Note that <tt>dot-child-name</tt> rule follows the philosophy of JSON strings when working with non-ASCII range of Unicode: it's
+                     allowed to contain bit sequences that cannot encode Unicode character (an single unpaired UTF-16 surrogate, for example). It can lead
+                     to unpredictable behavior of software that receives such query.
```
glyn, line 239: This text is similar to that in [RFC 7159 section 8.2](https://tools.ietf.org/html/rfc7159#section-8.2), but our goal is slightly different. RFC 7159 defined the JSON data format whereas we are defining a standard for JSONPath implementations. So we should probably say something about implementations rather than the more general "software". How about the following (which also has some editorial improvements)?
```suggestion
                  <t>Note that the <tt>dot-child-name</tt> rule follows the philosophy of JSON strings and is
                     allowed to contain bit sequences that cannot encode Unicode characters (a single unpaired UTF-16 surrogate, for example).
                     The behaviour of an implementation is undefined for child names which do not encode Unicode characters.
```

draft-normington-jsonpath-latest.xml
glyn: nit: Also, it would be nice to line up the comments vertically. :-)

remorhaz:
I've just rebuilt my local docker image without any changes and it pulled the fresh version of xml2rpc, so the unexpected changes should go away.

glyn:
> I've just rebuilt my local docker image without any changes and it pulled the fresh version of xml2rpc, so the unexpected changes should go away.

Yes, that's better thanks. Unfortunately, when I regenerate, I still see a diff: https://github.com/jsonpath-standard/internet-draft/commit/eabc0f0cee03bc6afaae23377a27b6fcc27fae2e. It looks like xml2rfc is non-deterministic! `¯\_(ツ)_/¯`. I guess there's not much we can do without patching xml2rfc...

remorhaz:
That's not very convenient, yes. But I don't see any quick solution, either.

glyn (approved):
Looks great now, thanks. I'm adding a few editorial changes and will merge soon.

glyn:
Ticket raised against xml2rfc: https://trac.tools.ietf.org/tools/xml2rfc/trac/ticket/553.

PR #22, opened 2020-09-24
Use "selector" instead of "matcher"

glyn:
Fixes https://github.com/jsonpath-standard/internet-draft/issues/20

See the [rendered HTML](https://glyn.github.io/internet-draft/) version of this PR.

### DIFF
--- a/draft-normington-jsonpath-latest.txt
+++ b/draft-normington-jsonpath-latest.txt
@@ -5,8 +5,8 @@
 Internet Engineering Task Force                       G. Normington, Ed.
 Internet-Draft                                              VMware, Inc.
 Intended status: Standards Track                           E. Surov, Ed.
-Expires: 27 March 2021                           TheSoul Publishing Ltd.
-                                                       23 September 2020
+Expires: 28 March 2021                           TheSoul Publishing Ltd.
+                                                       24 September 2020
 
 
                  JavaScript Object Notation (JSON) Path
@@ -37,7 +37,7 @@ Status of This Memo
    time.  It is inappropriate to use Internet-Drafts as reference
    material or to cite them other than as "work in progress."
 
-   This Internet-Draft will expire on 27 March 2021.
+   This Internet-Draft will expire on 28 March 2021.
 
 Copyright Notice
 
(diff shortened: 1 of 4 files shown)


glyn:
I particularly welcome @gregsdennis's comments since he raised https://github.com/jsonpath-standard/internet-draft/issues/20.

draft-normington-jsonpath-latest.txt
```diff
@@ -191,98 +191,98 @@ Internet-Draft   JavaScript Object Notation (JSON) Path   September 2020
...
 
-   Each matcher acts on its input list of nodes as follows.  For each
-   node in the list, the matcher may or may not match the node.  If the
-   matcher matches the node, it selects zero or more descendants of the
-   node, or the node itself.  The output list of nodes of a matcher is
-   the concatenation of the lists of selected nodes for each input node.
+   Each selector acts on its input list of nodes as follows.  For each
+   node in the list, the selector selects zero or more nodes, each of
+   which is a descendant of the node or the node itself.  The output
+   list of nodes of a selector is the concatenation of the lists of
```
gregsdennis, line 232: "distinct concatenation"

gregsdennis (changes requested):
I think the `matcher -> selector` changes are good.  I'd just add the word "distinct" in the place indicated.

Reading through, I saw a few other language nuances that could be cleared up.  I'll raise them in separate issues for discussion.

draft-normington-jsonpath-latest.txt
glyn: Does distinct concatenation mean with duplicates removed? That term is unfamiliar to me.

gregsdennis: Correct, no duplicates. I think mine currently regards the same value from different locations as distinct from one another, meaning it'll keep both.

It's worth discussing further, probably. Maybe just merge this for now, and I'll open an issue for it.

gregsdennis (approved):

draft-normington-jsonpath-latest.txt
glyn: Yes, I'd like to keep that as a separate issue. It's somewhat related to the placeholder we have in the spec about whether to remove duplicates from unions.

PR #28, opened 2020-10-06
Allow step of array slice to be omitted

glyn:
This allows slices of the form [0:10:].

### DIFF
--- a/draft-normington-jsonpath-latest.txt
+++ b/draft-normington-jsonpath-latest.txt
@@ -5,8 +5,8 @@
 Internet Engineering Task Force                       G. Normington, Ed.
 Internet-Draft                                              VMware, Inc.
 Intended status: Standards Track                           E. Surov, Ed.
-Expires: 28 March 2021                           TheSoul Publishing Ltd.
-                                                       24 September 2020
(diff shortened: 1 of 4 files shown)


draft-normington-jsonpath-latest.txt
```diff
@@ -519,7 +519,7 @@ Internet-Draft   JavaScript Object Notation (JSON) Path   September 2020
    (in base 10 and which may be omitted) separated by colons.
 
    array-slice = [ integer ] ws %x3A ws [ integer ]
-                      [ ws %x3A ws integer ]
+                      [ ws %x3A [ ws integer ] ]
```
gregsdennis, line 522: Should the `ws` be included in the brackets?  Maybe `ws [ %x3A ws [ integer ] ]` is better?

gregsdennis (changes requested):

draft-normington-jsonpath-latest.txt
glyn: It's syntactically equivalent because of the `ws` in the union syntax, but I agree it is more consistent.

glyn: changed

gregsdennis (approved):

PR #30, opened 2020-10-12
Run scripts/docker-gen.sh

mkmik:
running `docker-gen.sh` now runs xml2rfc version `3.2.1`.
This PR just checks in the minor modifications in the output that this tool does since `3.1.1`, so that next PRs will have more meaningful diffs in the output

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -11,10 +11,28 @@
        JSON Path defines a string syntax for identifying values
    within a JavaScript Object Notation (JSON) document. 
     " name="description">
-<meta content="xml2rfc 3.1.1" name="generator">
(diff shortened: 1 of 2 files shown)


glyn:
> running `docker-gen.sh` now runs xml2rfc version `3.2.1`.
> This PR just checks in the minor modifications in the output that this tool does since `3.1.1`, so that next PRs will have more meaningful diffs in the output

Unfortunately, xml2rfc is [non-deterministic](https://trac.tools.ietf.org/tools/xml2rfc/trac/ticket/553), so this PR won't make future diffs and more meaningful. Happy to merge it if the `3.1.1`->`3.2.1` has any value, but I doubt that.

Issue #29, opened 2020-10-12
Define behaviour for very large array indices and slice components


glyn:
Currently, there is no limit to the integer values which may be specified as array indices or as components (start, end, or step) of an array slice. The spec needs to define what behaviour is acceptable when very large positive or negative values are used.

For example, suppose a value such as 2^65 or -2^65 is used. May the implementation reject that with a (syntax or "find") error, should it clamp the value to the range of a signed 64 bit integer (*) before proceeding, or should this be handled some other way?

*: this may be too onerous for some implementations, but a signed 32 bit limit could conceivably be too draconian.

As @mkmik points out, JSON does not have sparse arrays, so such large values are unlikely to index valid elements. They simply (!) need to be handled in a consistent and helpful way.

This was prompted by a [discussion](https://github.com/jsonpath-standard/jsonpath-reference-implementation/pull/22#issuecomment-705771250) with @gregsdennis.

mkmik:
a notable language that lacks native 64-bit integers is JavaScript:

```
Welcome to Node.js v14.10.1.
Type ".help" for more information.
> x=1024*1024*1024
1073741824
> x*x
1152921504606847000
```
which is `24` off from the correct value which is `1152921504606846976`. This is because in JS numbers are implemented as 64-bit floats. There are ways around it, using BitNumber etc, but implementors need to deal with it and in some cases it might impact the API of the library.

gregsdennis:
For reference, here's how [JSON Schema] (http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.6.2) puts it.

> JSON Schema is programming language agnostic, and supports the full range of values described in the data model. Be aware, however, that some languages and JSON parsers may not be able to represent in memory the full range of values describable by JSON.

Fairly straightforward, and it relieves implementations from being _required_ to support some features due to limitations put forth by their framework.

glyn:
> For reference, here's how [JSON Schema](http://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.6.2) puts it.
> 
> > JSON Schema is programming language agnostic, and supports the full range of values described in the data model. Be aware, however, that some languages and JSON parsers may not be able to represent in memory the full range of values describable by JSON.
> 
> Fairly straightforward, and it relieves implementations from being _required_ to support some features due to limitations put forth by their framework.

[...]

PR #31, opened 2020-10-16
Change array slicing specification

mkmik:
Related to #29 and jsonpath-standard/jsonpath-reference-implementation#22

@glyn

### DIFF
--- a/draft-normington-jsonpath-latest.txt
+++ b/draft-normington-jsonpath-latest.txt
@@ -5,8 +5,10 @@
 Internet Engineering Task Force                       G. Normington, Ed.
 Internet-Draft                                              VMware, Inc.
 Intended status: Standards Track                           E. Surov, Ed.
-Expires: 10 April 2021                           TheSoul Publishing Ltd.
-                                                          7 October 2020
+Expires: 29 April 2021                           TheSoul Publishing Ltd.
+                                                            M. Mikulicic
+                                                            VMware, Inc.
+                                                         26 October 2020
 
 
                  JavaScript Object Notation (JSON) Path
@@ -37,31 +39,25 @@ Status of This Memo
    time.  It is inappropriate to use Internet-Drafts as reference
    material or to cite them other than as "work in progress."
 
-   This Internet-Draft will expire on 10 April 2021.
+   This Internet-Draft will expire on 29 April 2021.
 
 Copyright Notice
 
    Copyright (c) 2020 IETF Trust and the persons identified as the
    document authors.  All rights reserved.
 
-
-
-
-
-
-
-
-
-
-Normington & Surov        Expires 10 April 2021                 [Page 1]
-
-Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
-
-
    This document is subject to BCP 78 and the IETF Trust's Legal
    Provisions Relating to IETF Documents (https://trustee.ietf.org/
    license-info) in effect on the date of publication of this document.
    Please review these documents carefully, as they describe your rights
+
+
+
+Normington, et al.        Expires 29 April 2021                 [Page 1]
+
+Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
+
+
    and restrictions with respect to this document.  Code Components
    extracted from this document must include Simplified BSD License text
    as described in Section 4.e of the Trust Legal Provisions and are
@@ -71,7 +67,7 @@ Table of Contents
 
    1.  Introduction  . . . . . . . . . . . . . . . . . . . . . . . .   2
      1.1.  Requirements Language . . . . . . . . . . . . . . . . . .   2
-     1.2.  ABNF Syntax . . . . . . . . . . . . . . . . . . . . . . .   3
+     1.2.  ABNF Syntax . . . . . . . . . . . . . . . . . . . . . . .   2
    2.  JSON Path Syntax and Semantics  . . . . . . . . . . . . . . .   3
      2.1.  Overview  . . . . . . . . . . . . . . . . . . . . . . . .   3
      2.2.  Terminology . . . . . . . . . . . . . . . . . . . . . . .   3
@@ -85,13 +81,13 @@ Table of Contents
          2.6.2.2.  Semantics . . . . . . . . . . . . . . . . . . . .   7
          2.6.2.3.  Union Child . . . . . . . . . . . . . . . . . . .   7
          2.6.2.4.  Array Access  . . . . . . . . . . . . . . . . . .   9
-   3.  IANA Considerations . . . . . . . . . . . . . . . . . . . . .  11
-   4.  Security Considerations . . . . . . . . . . . . . . . . . . .  11
-   5.  Alternatives  . . . . . . . . . . . . . . . . . . . . . . . .  11
-   6.  References  . . . . . . . . . . . . . . . . . . . . . . . . .  11
-     6.1.  Normative References  . . . . . . . . . . . . . . . . . .  12
-     6.2.  Informative References  . . . . . . . . . . . . . . . . .  12
-   Authors' Addresses  . . . . . . . . . . . . . . . . . . . . . . .  12
+   3.  IANA Considerations . . . . . . . . . . . . . . . . . . . . .  12
+   4.  Security Considerations . . . . . . . . . . . . . . . . . . .  12
+   5.  Alternatives  . . . . . . . . . . . . . . . . . . . . . . . .  12
+   6.  References  . . . . . . . . . . . . . . . . . . . . . . . . .  13
+     6.1.  Normative References  . . . . . . . . . . . . . . . . . .  13
+     6.2.  Informative References  . . . . . . . . . . . . . . . . .  13
+   Authors' Addresses  . . . . . . . . . . . . . . . . . . . . . . .  14
 
 1.  Introduction
 
@@ -106,18 +102,17 @@ Table of Contents
    "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
    document are to be interpreted as described in RFC 2119 [RFC2119].
 
-
-
-
-Normington & Surov        Expires 10 April 2021                 [Page 2]
-
-Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
-
-
 1.2.  ABNF Syntax
 
    The syntax in this document conforms to ABNF as defined by RFC 5234
    [RFC5234].
+
+
+
+Normington, et al.        Expires 29 April 2021                 [Page 2]
+
+Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
+
 
    ABNF terminal values in this document define Unicode code points
    rather than their UTF-8 encoding.  For example, the Unicode PLACE OF
@@ -157,19 +152,6 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
    If no node is selected and no error has occurred, an implementation
    MUST return an empty list of nodes.
 
-
-
-
-
-
-
-
-
-Normington & Surov        Expires 10 April 2021                 [Page 3]
-
-Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
-
-
    Syntax errors in the JSON Path SHOULD be detected before selection is
    attempted since these errors do not depend on the JSON document.
    Therefore, an implementation SHOULD take a JSON Path and produce an
@@ -180,6 +162,13 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
    Alternatively, an implementation MAY take a JSON Path and a JSON
    document and produce a list of nodes or an optional error (but not
    both).
+
+
+
+Normington, et al.        Expires 29 April 2021                 [Page 3]
+
+Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
+
 
    For any implementation, if a syntactically invalid JSON Path is
    provided, the implementation MUST return an error.
@@ -218,14 +207,6 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
    with a list of nodes which is the result of applying the JSON Path to
    the input JSON document.
 
-
-
-
-Normington & Surov        Expires 10 April 2021                 [Page 4]
-
-Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
-
-
    Each selector acts on its input list of nodes as follows.  For each
    node in the list, the selector selects zero or more nodes, each of
    which is a descendant of the node or the node itself.  The output
@@ -236,6 +217,14 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
    the input document is: "{"a":[{"b":0},{"b":1},{"c":2}]}".  As we will
    see later, the JSON Path "$.a[*].b" selects the following list of
    nodes: "0", "1".  Let's walk through this in detail.
+
+
+
+
+Normington, et al.        Expires 29 April 2021                 [Page 4]
+
+Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
+
 
    The JSON Path consists of "$" followed by three selectors: ".a",
    "[*]", and ".b".
@@ -277,7 +266,18 @@ Syntax
 
 
 
-Normington & Surov        Expires 10 April 2021                 [Page 5]
+
+
+
+
+
+
+
+
+
+
+
+Normington, et al.        Expires 29 April 2021                 [Page 5]
 
 Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 
@@ -333,7 +333,7 @@ Semantics
 
 
 
-Normington & Surov        Expires 10 April 2021                 [Page 6]
+Normington, et al.        Expires 29 April 2021                 [Page 6]
 
 Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 
@@ -389,7 +389,7 @@ Syntax
 
 
 
-Normington & Surov        Expires 10 April 2021                 [Page 7]
+Normington, et al.        Expires 29 April 2021                 [Page 7]
 
 Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 
@@ -445,7 +445,7 @@ Semantics
 
 
 
-Normington & Surov        Expires 10 April 2021                 [Page 8]
+Normington, et al.        Expires 29 April 2021                 [Page 8]
 
 Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 
@@ -485,9 +485,9 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 Syntax
 
    Array access is a union element which selects zero or more elements
-   of an array node.  It selects no nodes from a node which is not an
-   array.  Array access takes the form of an index, which selects at
-   most one element, or a slice, which selects zero or more elements.
+   of an array node.  Array access takes the form of an index, which
+   selects at most one element, or a slice, which selects zero or more
+   elements.
 
    union-element =/ array-index / array-slice
 
@@ -501,7 +501,7 @@ Syntax
 
 
 
-Normington & Surov        Expires 10 April 2021                 [Page 9]
+Normington, et al.        Expires 29 April 2021                 [Page 9]
 
 Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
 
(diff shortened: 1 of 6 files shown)


draft-normington-jsonpath-latest.txt
```diff
@@ -524,73 +524,100 @@ Internet-Draft   JavaScript Object Notation (JSON) Path     October 2020
...
+Limits
+
+   The RECOMMENDED maximum length of the array is 2^64.  The numerical
+   domain of array indices is thus "0 <= b < 2^64".
 
-   For zero "step", the array slice is empty and no elements are
-   selected.
+   The RECOMMENDED domain of input slice bounds is "-2^64 <= b < 2^64".
+   If not possible, implementations SHOULD at least aim for "-2^63 <= b
+   < 2^63".
```
gregsdennis, line 599: I'm not a fan of specifying a limit, though I do agree with the usage of RECOMMENDED to allow leeway for implementations.

Instead of specifying a hard limit, I think we should specify a behavior when an implementation encounters values outside of its language's range.

mkmik: perhaps we should define a reasonable lower bound as a MUST? e.g. I don't think it makes sense to have conforming implementations that only support 256 element arrays

glyn: Perhaps a signed 32 bit range could be mandatory (MUST) and a signed 64 bit range recommended (SHOULD).

mkmik: SGTM, w.r.t array length.

What about slice start and end positions? The RI offers an API that allows for full 64-bit absolute range *and* a full 64-bit range for "end relative" positions. If "end relative" positions are "naively" implemented with negative numbers (as suggested by the standard's own syntactic layer), that would limit implementations to only 63 bits to specify an absolute position (the same problem arises for those implementations that choose a 32-bit index space, and use 31-bit signed positions).

glyn: On balance, I'd prefer to keep the spec simple rather than complicate it for "non-naive" implementations. Signed 64 bit values of slice start and end positions should be fine for all real-world JSON arrays. Implementations that choose to support only signed 32 bit values could be at a disadvantage for very large JSON arrays, but in that case, it would probably be better to use an implementation that supports signed 64 bit values.

gregsdennis:
No.  Stop talking about actual values. We shouldn't be doing that. We should define ***behavior***, not limits. Limits are an implementation detail. What happens when those limits are exceeded is for us to define.

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                        The indices of the slice are the same as the values of i iterated over by this <tt>for</tt> loop in C:
-                     </t>
+                     <t>A slice expression produces a new array containing a subset of the elements of the input array.</t>
+                     <t>A slice is defined by the two bounds, <tt>start</tt> and <tt>end</tt>,
+                        and a <tt>step</tt> which governs the direction of iteration and the iteration stride.
+                        Each of those parameters is optional.</t>
+                     <t>Negative bounds are interpreted as offsets from the end of the array, namely <tt>b + len</tt>.
+                        This remapping is performed before any other evaluation step.</t>
+                     <t>A slice bound <tt>b</tt> is considered valid if <tt>0 &lt;= b &lt; len</tt>.
+                        An invalid bound is treated as an omitted bound. Omitted bounds take a default value
```
glyn, line 394: I'm not keen on the "invalid" terminology, as this connotes an error or mistake. Such bounds are perfectly valid. We can either describe how they are clamped, depending on the step value. Or we could allow them to default to +/- infinity and make it clear how very large positive and negative explicit (including defaulted) bounds are interpreted.

draft-normington-jsonpath-latest.xml
```diff
@@ -20,13 +20,13 @@
   <!-- category values: std, bcp, info, exp, and historic
     ipr values: trust200902, noModificationTrust200902, noDerivativesTrust200902,
        or pre5378Trust200902
-    you can add the attributes updates="NNNN" and obsoletes="NNNN" 
+    you can add the attributes updates="NNNN" and obsoletes="NNNN"
```
glyn, line 23: These whitespace changes are distracting. Was this deliberate or just the result of the editor?

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        Each of those parameters is optional.</t>
+                     <t>Negative bounds are interpreted as offsets from the end of the array, namely <tt>b + len</tt>.
+                        This remapping is performed before any other evaluation step.</t>
+                     <t>A slice bound <tt>b</tt> is considered valid if <tt>0 &lt;= b &lt; len</tt>.
+                        An invalid bound is treated as an omitted bound. Omitted bounds take a default value
+                        that depends on the value of the <tt>step</tt> parameter:</t>
+                     <t>If <tt>step &gt; 0</tt>, then <tt>start</tt> defaults to <tt>0</tt>
+                        while <tt>end</tt> defaults to <tt>len</tt>.</t>
+                     <t>If <tt>step &lt; 0</tt>, then <tt>start</tt> defaults to <tt>len-1</tt>,
+                        while an omitted <tt>end</tt> causes the iteration to terminate at 0 (inclusive).</t>
```
glyn, line 399: So the earlier statement of omitted bounds taking a default value isn't _quite_ accurate as in this case, we are avoiding saying what the default value is.

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                     <t>If <tt>step</tt> is <tt>0</tt>, the resulting array is empty.</t>
+                     <t>If <tt>step</tt> is omitted, it defaults to <tt>1</tt>.</t>
+                     <t>The resulting array is then constructed by iterating from <tt>start</tt> (inclusive)
+                        to <tt>end</tt> (exclusive), collecting array elements at index <tt>i</tt>,
```
glyn, line 403: `i` is undefined.

We don't say _how_ individual elements are indexed. In particular, index 0 selects the first element of the array.

"collecting" could be changed to "selecting" for consistency with the rest of the spec.

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        and incrementing the iterator <tt>i</tt> by <tt>step</tt> after each iteration. </t>
+                     <t>If <tt>len</tt> is <tt>0</tt> the resulting array is empty.</t>
+                     <t>If <tt>step &gt; 0</tt> and <tt>start &gt; end</tt> the resulting array is empty.</t>
+                     <t>If <tt>step &lt; 0</tt> and <tt>start &lt; end</tt> the resulting array is empty.</t>
+                  </section>
```
glyn, line 408: I hope these three paragraphs are actually consequences of the preceding text. If so, it would be better to be clear that they are consequences rather than additional rules. E.g. "As a consequence of the above:" follow by a list of three bullets.

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                     </t>
-                     <t>For zero <tt>step</tt>, the array slice is empty and no elements are selected.
-                     </t>
-                  </section>       
-               </section>             
-            </section>       
-         </section>       
+                     <t>If the <tt>end</tt> bound is omitted, the iteration must terminate at 0 inclusive.
+                     This can be achieved by setting <tt>end = -1</tt>, if the underyling numeric type allows it.
+                     Alternatively the termination condition can be changed to <tt>i &gt;= 0</tt>.</t>
```
glyn, line 426: I think we might be going into too much implementation detail here, which sets a nasty precedent for the rest of the spec. If we keep the pseudocode, I wonder if we can somehow indicate that it isn't normative.

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                     <t>The RECOMMENDED maximum length of the array is 2^64.
+                        The numerical domain of array indices is thus <tt>0 &lt;= b &lt; 2^64</tt>.</t>
+                     <t>The RECOMMENDED domain of input slice bounds is <tt>-2^64 &lt;= b &lt; 2^64</tt>.
+                        If not possible, implementations SHOULD at least aim for <tt>-2^63 &lt;= b &lt; 2^63</tt>.</t>
+                  </section>
+                  <section numbered="false" toc="exclude">
+                     <name>Idioms</name>
+                     <t>First N elements: <tt>$[:N]</tt>.</t>
+                     <t>Last N elements: <tt>$[-N:]</tt>.</t>
+                     <t>All elements in reverse order: <tt>$[::-1]</tt>.</t>
```
glyn, line 439: I think I'd prefer to mention these inline to motivate the slice notation.

glyn (changes requested):
I've made quite a few comments, mostly stylistic and trying to make the spec more formal.

glyn:
> No. Stop talking about actual values. We shouldn't be doing that. We should define _**behavior**_, not limits. Limits are an implementation detail. What happens when those limits are exceeded is for us to define.

Yes, I have some sympathy for this. From a spec point of view, we should either get correct slice behaviour or an error (still to be mentioned in the text). Implementations which choose poor limits will soon get an appropriate reputation.

This is somewhat analogous to specifying `malloc` non-deterministically: it either succeeds or it fails, but we don't try to say when it will fail as that would get into implementation detail.

glyn:
Also, I can't decide whether it would be a good thing to mention that slicing is based on Python slices. WDYT @mkmik  @gregsdennis?

mkmik:
currently the aspirational spec deviates from python slices in one detail: when step is 0 we yield an empty slice, while python raises an erro

mkmik:
Javascript's `.slice` method has the same semantics w.r.t the negative indices and behaviour w.r.t out of bound indices, but it lacks a `step`. Which is good because the JSON in JSONPath has a historical connection with JS. Should we mention that the core slicing syntax (excluding the step part) is based on JavaScript semantics?

https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(array) may not be exhaustive but shows that python is the only language that has inclusive start, exclusive end and a step.

The "python connection" thus comes into play only once we introduce `step`. If we want to support `step` (not sure if we have to), then I think it would be quite confusing if the behaviour of JSONPath were different than the only other (popular) language that has the `first:end:step` syntax.

Whether the spec should explicitly acknowledge this connection is another question. One one hand it would be weird to not acknowledge it, but on the other hand, what if the spec ends up being subtly different from python's? is there an unambiguous spec of the slicing expression for python or is it de-facto implementation defined?

glyn:
> currently the aspirational spec deviates from python slices in one detail: when step is 0 we yield an empty slice, while python raises an erro[r]

Yeah, that's a bit of a pain when it comes to mentioning Python slices. In an earlier version, a zero step was disallowed by the syntax, which is another solution. But someone didn't like that...

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                     </t>
-                     <t>The members of a slice depend on whether the step value is positive, negative, or zero as follows. If the step value is
-                        omitted, it defaults to <tt>1</tt>.
-                     </t>
-                     <t>For positive integer <tt>step</tt> and integers <tt>start</tt> and <tt>end</tt>, the array slice
-                        <tt>start:end:step</tt> consists of the indices starting at <tt>start</tt> and increasing repeatedly
-                        by <tt>step</tt> while the resultant index is less than <tt>end</tt>.
-                        The indices of the slice are the same as the values of i iterated over by this <tt>for</tt> loop in C:
-                     </t>
+                     <t>A slice expression produces a new array containing a subset of the elements of the input array.</t>
```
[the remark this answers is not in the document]
mkmik: if step is negative the output of the slice expression cannot be represented by a pure selection of the input array

draft-normington-jsonpath-latest.xml
```diff
@@ -381,57 +381,66 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                     <t>An array index selects any element of an array which is indexed by the integer index.
-                        It selects no nodes from a node which is not an array.
-                     </t>
-                     <t>A non-negative index <tt>i</tt> selects an element of an array of length <tt>len</tt> if and only if <tt>0 &lt;= i &lt; len</tt>.
-                        The index <tt>0</tt> selects the first element of a non-empty array and, providing <tt>i &lt; len - 1</tt>, index <tt>i + 1</tt>
-                        selects the element immedately after (in array order) the element selected by index <tt>i</tt>.
-                     </t>
-                     <t>A negative index <tt>j</tt> selects an element of an array of length <tt>len</tt> if and only if <tt>0 &lt;= j + len &lt; len</tt>,
-                        in which case it selects the same element as the non-negative index <tt>j + len</tt>.
-                     </t>
```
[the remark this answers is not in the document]
mkmik: uh I thought we had another section about (non-slice) indexing; will put back

draft-normington-jsonpath-latest.xml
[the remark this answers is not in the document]
glyn: Why not? Ordering? In my mind, selection doesn't imply a particular ordering.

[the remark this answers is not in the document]
mkmik: yeah ordering was my concern. To me "selects" sounds like the original ordering is preserved and we just apply a "filter" that removes extra elements. What we do here is to apply a transformation, we "map" the input array into an output array.

that said, that's only my interpretation, if "selects" clearly conveys the desired meaning of "remapping" I'm ok with it.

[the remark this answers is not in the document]
glyn: "select" will only convey specific meanings if we say so. I'd prefer not to try to overload the term and be explicit about the ordering properties of slices. Maybe in this introductory statement we can say:
```suggestion
                     <t>A slice expression selects a subset of the elements of the input array, in the same order as the array or the reverse order, depending on the expression.</t>
```

glyn:
> is there an unambiguous spec of the slicing expression for python or is it de-facto implementation defined?

[Apparently](https://docs.python.org/2.0/ref/slicings.html) not.

gregsdennis:
The other aspect to consider is that if Python updates their definition of a slice (to include a step, perhaps) then or spec is out of date.  It may be worth mentioning Python as a source of inspiration, but then proceed to define the slice anyway.

glyn:
> The other aspect to consider is that if Python updates their definition of a slice (to include a step, perhaps) then our spec is out of date. It may be worth mentioning Python as a source of inspiration, but then proceed to define the slice anyway.

Absolutely. I wasn't trying to suggest that we delegate the spec to Python, but that we acknowledge the inspiration as you say. They may want to lift our spec in due course. ;-)

draft-normington-jsonpath-latest.txt
mkmik: PTAL

draft-normington-jsonpath-latest.xml
mkmik: PTAL

draft-normington-jsonpath-latest.xml
mkmik: PTAL

mkmik:
@glyn , @gregsdennis PTAL;

I looked around how people explain the python slice behavior with negative indices and I also gave this as an exercise to a a more junior colleague. I think there might be value in adding a little "level of indirection" by introducing the concept of "lower" and "upper" bound, to contextualize the need for a different terminating condition.

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
                   </section>
                   <section numbered="false" toc="exclude">
                      <name>Semantics</name>
-                     <t>An array index selects any element of an array which is indexed by the integer index.
-                        It selects no nodes from a node which is not an array.
```
glyn, line 388: The behaviour of array access for non-array nodes is now undefined.

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                        <tt>start:end:step</tt> consists of the indices starting at <tt>start</tt> and increasing repeatedly
-                        by <tt>step</tt> while the resultant index is less than <tt>end</tt>.
-                        The indices of the slice are the same as the values of i iterated over by this <tt>for</tt> loop in C:
-                     </t>
-                     <sourcecode type="c">
-for (i = start; i &lt; end; i = i + step) {
-  ...
-}
+                     <t>Array slicing is inspired by the behaviour of the <tt>Array.prototype.slice</tt> method
+                        of the JavaScript language as defined by ECMA-262, with the addition of the <tt>step</tt>
```
glyn, line 388: Probably worth adding a non-normative reference for ECMA-262.

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+  END IF
                      </sourcecode>
-                     <t>When <tt>step</tt> is negative, the default value of <tt>start</tt> is <tt>len - 1</tt> (where <tt>len</tt> is the length of the array)
-                        and the default value of <tt>end</tt> is <tt>-1</tt> (literally, not in the earlier sense of a shorthand for <tt>len - 1</tt>).
-                     </t>
-                     <t>For zero <tt>step</tt>, the array slice is empty and no elements are selected.
-                     </t>
+
+                     <t>The semantics of the array indexing expression <tt>[i]</tt> is defined as syntactic sugar for
+                        <tt>a[i:Normalize(i)+1:1]</tt>.</t>
```
glyn, line 439: It seems strange to talk about _syntactic_ sugar in the semantics section. Perhaps like the following would be clearer:
```suggestion
                     <t>The result of the array indexing expression <tt>[i]</tt> is defined to be the result of the array slicing expression
                        <tt>a[i:Normalize(i)+1:1]</tt>.</t>
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+
+  IF start is omitted THEN
+    start = len - 1
+  END IF
+  IF end is omitted THEN
+    end = -len - 1
+  END IF
+
+END IF
                      </sourcecode>
```
glyn, line 424: This is fine, but a table would have been equally easy to follow.

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+    i = i + step
+  END WHILE
+
+END IF
+                     </sourcecode>
+
+                     <t>The case when <tt>step = 0</tt> intentionally doesn't select any array element
+                        and thus produces an empty result array.</t>
+
+                     <t>All parameters, including the array length, SHOULD be representable as a signed 64-bit integer.
```
glyn, line 489: I'm not sure what value this statement adds and it could even be viewed as implementation bias. The following statements say what to do when a value is encountered which the implementation can't cope with.

mkmik: Makes sense, thanks

scripts/gen.sh
```diff
@@ -7,6 +7,15 @@ readonly script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
\ No newline at end of file
+xml2rfc ./draft-normington-jsonpath-latest.xml --text --html && \
+aex draft-normington-jsonpath-latest.txt \
+| grep -v "lower =" \
+| grep -v "upper =" \
+| grep -v "i =" \
+| grep -v "step =" \
+| grep -v "step &gt;=" \
+| grep -v "start =" \
+| grep -v "end =" \
```
glyn, line 18: `aex` really is a thing of beauty, isn't it. ;-)

scripts/gen.sh
```diff
@@ -7,6 +7,15 @@ readonly script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+| bap -S path -q && cp draft-normington-jsonpath-latest.html docs/index.html
+# TODO fix aex to no extract ABNF-looking rules from pseudocode blocks
```
glyn, line 20: `aex` really is a thing of beauty, isn't it. ;-)
```suggestion
# TODO fix aex to not extract ABNF-looking rules from pseudocode blocks
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        An implementation MUST raise an error when parsing an expression that doesn't fit the
+                        integer representation used to represent the slice expression parameters.
```
glyn, line 491: Section 2.3 allows implementations to defer raising errors until applying the JSONPath to a document. So how about something like this instead?
```suggestion
                        An implementation MUST raise an error if any of the slice expression parameters does not fit in the implementation's representation of an integer.
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +384,115 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        If a successfully parsed slice expression is evaluated against an array whose size doesn't
+                        fit in the integer representation of the slice expression parameters, the implementation
```
glyn, line 493:
```suggestion
                        fit in the implementation's representation of an integer, the implementation
```

glyn:
@mkmik please remember to add yourself to the list of authors.

draft-normington-jsonpath-latest.xml
```diff
@@ -435,7 +435,7 @@ FUNCTION Normalize(i):
   END IF
                      </sourcecode>
 
-                     <t>The semantics of the array indexing expression <tt>[i]</tt> is defined as syntactic sugar for
+                     <t>The result of the array indexing expression <tt>[i]</tt> is defined to be the result of the array slicing expression
                         <tt>a[i:Normalize(i)+1:1]</tt>.</t>
```
glyn, line 439: Whoops!:
```suggestion
                        <tt>[i:Normalize(i)+1:1]</tt>.</t>
```

draft-normington-jsonpath-latest.xml
glyn: To be clear, I was suggesting deleting the sentence.

draft-normington-jsonpath-latest.xml
mkmik: thanks for the suggestion, rewrote as a table

draft-normington-jsonpath-latest.xml
mkmik: good catch; I put it back where `a(i)` is defined

draft-normington-jsonpath-latest.xml
mkmik: added

draft-normington-jsonpath-latest.xml
mkmik: suggestion applied

draft-normington-jsonpath-latest.xml
```diff
@@ -63,6 +63,20 @@
         <email>esurov.tsp@gmail.com</email>
      </address>
     </author>
+    <author fullname="Marko Mikulicic" initials="M" role="editor" surname="Mikulicic">
```
glyn, line 66: Yes, according to the [definition](https://tools.ietf.org/html/rfc7991#section-2.7.6) of the author tag, `editor` is the only option for a role. f you'd prefer not to act an an editor:
```suggestion
    <author fullname="Marko Mikulicic" initials="M" surname="Mikulicic">
```

draft-normington-jsonpath-latest.xml
glyn: I can see it in the syntax section, but it's probably better to move that to somewhere near the top of the semantics section.

draft-normington-jsonpath-latest.xml
mkmik: thanks

draft-normington-jsonpath-latest.xml
mkmik: moved on top

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-                     </t>
-                     <sourcecode type="c">
-for (i = start; i &lt; end; i = i + step) {
-  ...
-}
+                     <t>Array slicing is inspired by the behaviour of the <tt>Array.prototype.slice</tt> method
+                        of the JavaScript language as defined by the ECMA-262 standard
+                        <xref target="ECMA-262" format="default">ECMA-262</xref>,
+                        with the addition of the <tt>step</tt> parameter, which is inspired by the Python slice expression.
+                        This section will fully specify its behaviour, without dependencies on externally defined behaviour.
```
glyn, line 404:
```suggestion
                        This section specifies the behaviour fully, without depending on JavaScript or Python behaviour.
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        Array indexing is defined in terms of the slicing operation.</t>
```
glyn, line 405:
```suggestion
                        Array indexing is defined in terms of array slicing.</t>
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        as the array or the reverse order, depending on the sign of the <tt>step</tt> parameter.
+                        It selects no nodes from a node which is not an array.</t>
+
+                     <t>A slice is defined by the two slice parameters, <tt>start</tt> and <tt>end</tt>,
+                        and an iteration delta, <tt>step</tt>. Each of these parameters is optional.
+                        <tt>len</tt> is the length of the input array.</t>
+
+                     <t>The default value for <tt>step</tt> is <tt>1</tt>.
+                        The default values for <tt>start</tt> and <tt>end</tt> depend on the sign of <tt>step</tt>,
+                        as defined in the following table:</t>
```
glyn, line 417:
```suggestion
                        as follows:</t>
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+
+                     <table>
+                       <name>Defaults</name>
```
glyn, line 420: Perhaps the following?
```suggestion
                       <name>Default array slice start and end values</name>
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                         <tr> <th>Condition</th>    <th>start</th>   <th>end</th> </tr>
+                       </thead>
+                       <tbody>
+                         <tr> <td>step &gt;= 0</td> <td>0</td>       <td>len</td> </tr>
+                         <tr> <td>step &lt; 0</td>  <td>len - 1</td> <td>-len - 1</td> </tr>
+                       </tbody>
+                     </table>
+
+                     <t>Slice expression parameters <tt>start</tt> and <tt>end</tt> are not directly usable
+                        as slice bounds and must be first normalized. Normalization is defined as:</t>
```
glyn, line 431:
```suggestion
                        as slice bounds and must first be normalized. Normalization is defined as:</t>
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
-for (i = start; i &gt; end; i = i + step) {
-  ...
-}
+
+                     <t>The result of the array indexing expression <tt>[i]</tt> is defined to be the result of the array
+                        slicing expression <tt>[i:Normalize(i)+1:1]</tt>.</t>
+
+                     <t>Slice expression parameters <tt>start</tt> and <tt>end</tt> are used to derive slice bounds
+                        <tt>lower</tt> and <tt>upper</tt>. The direction of the iteration defined
+                        by the sign of <tt>step</tt> determines which of the parameters is the lower bound and which
```
glyn, line 447:
```suggestion
                        <tt>lower</tt> and <tt>upper</tt>. The direction of the iteration, defined
                        by the sign of <tt>step</tt>, determines which of the parameters is the lower bound and which
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+    upper = MIN(MAX(n_end, 0), len)
+  ELSE
+    upper = MIN(MAX(n_start, -1), len-1)
+    lower = MIN(MAX(n_end, -1), len-1)
+  END IF
+
+  RETURN (lower, upper)
+</sourcecode>
+
+                     <t>The slice expression selects indices between the lower and upper bounds.
```
glyn, line 466:
```suggestion
                     <t>The slice expression selects elements with indices between the lower and upper bounds.
```

draft-normington-jsonpath-latest.xml
```diff
@@ -384,50 +397,105 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+END IF
                      </sourcecode>
-                     <t>When <tt>step</tt> is negative, the default value of <tt>start</tt> is <tt>len - 1</tt> (where <tt>len</tt> is the length of the array)
-                        and the default value of <tt>end</tt> is <tt>-1</tt> (literally, not in the earlier sense of a shorthand for <tt>len - 1</tt>).
-                     </t>
-                     <t>For zero <tt>step</tt>, the array slice is empty and no elements are selected.
-                     </t>
+
+                     <t>The case when <tt>step = 0</tt> intentionally doesn't select any array element
+                        and thus produces an empty result array.</t>
```
glyn, line 491:
```suggestion
                     <t>When <tt>step = 0</tt>, no elements are selected and the result array is empty.
                       This is intentional.</t>
```

glyn (approved):
Excellent - a great improvement. Thanks!

PR #36, opened 2020-10-27
Gentle introduction to array access

glyn:
See also a [rendered version](https://glyn.github.io/internet-draft/#name-array-access-2).

### DIFF
--- a/draft-normington-jsonpath-latest.xml
+++ b/draft-normington-jsonpath-latest.xml
@@ -397,57 +397,84 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
                   </section>
                   <section numbered="false" toc="exclude">
                      <name>Semantics</name>
-                     <t>Array slicing is inspired by the behaviour of the <tt>Array.prototype.slice</tt> method
-                        of the JavaScript language as defined by the ECMA-262 standard
-                        <xref target="ECMA-262" format="default">ECMA-262</xref>,
-                        with the addition of the <tt>step</tt> parameter, which is inspired by the Python slice expression.
-                        This section specifies the behaviour fully, without depending on JavaScript or Python behaviour.
-                        Array indexing is defined in terms of array slicing.</t>
+                     <section numbered="false" toc="exclude">
+                        <name>Informal Introduction</name>
+                        <t>This section is non-normative.</t>
+                        <t>Array indexing is a way of selecting a particular element of an array using a 0-based index.
+                           For example, the expression <tt>[0]</tt> selects the first element of a non-empty array.
+                        </t>
+                        <t>Negative indices index from the end of an array. For example, the expression <tt>[-2]</tt>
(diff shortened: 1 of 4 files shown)


mkmik (approved):
LGTM!

draft-normington-jsonpath-latest.xml
```diff
@@ -397,57 +397,84 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                        </t>
+                        <t>When <tt>step</tt> is <tt>0</tt>, no elements are selected. This is the one case which differs from the
+                           behaviour of Python, which raises an error in this case.
+                        </t>
+                        <t>The following section specifies the behaviour fully, without depending on JavaScript or Python behaviour.
+                        </t>
+                     </section>
+                     <section numbered="false" toc="exclude">
+                        <name>Normative Definition</name>
+                        <t>Array access includes array slicing and array indexing, which is defined in terms of array slicing.</t>
```
mkmik, line 432: "Array access" sounds a bit odd to me; I know it's the title of the section so it doesn't come from this PR.

What about "Array selection" ?

glyn: Good point. Actually, we use the term "selector" elsewhere, so I'll try that and see how it works out.

glyn: @mkmik Done. See what you think. (I avoided renaming "union child" to "union child selector" but we may decide to do that for consistency.)

glyn:
Raised https://github.com/jsonpath-standard/internet-draft/issues/37 which I think should be addressed as a global editorial change, separate from this PR.

draft-normington-jsonpath-latest.xml
```diff
@@ -397,57 +397,84 @@ array-slice = [ integer ] ws %x3A ws [ integer ]
...
+                           elements of an array in reverse order.
+                        </t>
+                        <t>When <tt>step</tt> is <tt>0</tt>, no elements are selected. This is the one case which differs from the
+                           behaviour of Python, which raises an error in this case.
+                        </t>
+                        <t>The following section specifies the behaviour fully, without depending on JavaScript or Python behaviour.
+                        </t>
+                     </section>
+                     <section numbered="false" toc="exclude">
+                        <name>Normative Definition</name>
```
glyn, line 431: I'm wondering about specifically flagging this section as normative because that might then cast some doubt over whether other sections are normative. Maybe a bland title such as "Details" would be safer?

mkmik: SGTM

draft-normington-jsonpath-latest.xml
mkmik: LGTM

draft-normington-jsonpath-latest.xml
glyn: Ok, I went for "Detailed Semantics".

PR #38, opened 2020-10-27
Use xpath to feed only abnf soucecode to aex and bap

mkmik:
thus avoding other types of sourcecode (such as pseudocode) to be validated by the ABNF checker.

aex is still needed because it reformats the content for bap somehow (didn't investigate).

### DIFF
--- a/scripts/gen.sh
+++ b/scripts/gen.sh
@@ -8,14 +8,7 @@ readonly script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$script_dir/.."
 
 xml2rfc ./draft-normington-jsonpath-latest.xml --text --html && \
-aex draft-normington-jsonpath-latest.txt \
-| grep -v "lower =" \
-| grep -v "upper =" \
-| grep -v "i =" \
-| grep -v "step =" \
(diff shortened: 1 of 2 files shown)


scripts/gen.sh
```diff
@@ -8,14 +8,8 @@ readonly script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-| grep -v "upper =" \
-| grep -v "i =" \
-| grep -v "step =" \
-| grep -v "step &gt;=" \
-| grep -v "start =" \
-| grep -v "end =" \
+xmlstarlet sel -T -t -v '//sourcecode[@type="abnf"]' ./draft-normington-jsonpath-latest.xml 2>/dev/null \
+| aex \
 | bap -S path -q && cp draft-normington-jsonpath-latest.html docs/index.html
 # TODO fix aex to not extract ABNF-looking rules from pseudocode blocks
```
glyn, line 14: I think the TODO is addressed by this PR.

glyn (approved):
Maybe delete the TODO since you have provided a good workaround.

scripts/gen.sh
mkmik: good catch done

PR #41, opened 2020-11-25
Editorial improvements

glyn:
Thanks for feedback from Daniel Parker and Carsten Bormann.

### DIFF
--- a/draft-normington-jsonpath-latest.md
+++ b/draft-normington-jsonpath-latest.md
@@ -89,7 +89,7 @@ normative:
 --- abstract
 
 
-JSON Path defines a string syntax for identifying values
+JSONPath defines a string syntax for identifying values
 within a JavaScript Object Notation (JSON) document.
 
 --- note_
@@ -101,7 +101,7 @@ as an Internet Draft.**
 
 # Introduction
 
-JSON Path, or rather JSONPath, was introduced by Stefan Goessner as a simple
+JSONPath was introduced by Stefan Goessner as a simple
 form of XPath for JSON.
 See his original article {{Goessner}}.
 
@@ -122,14 +122,14 @@ in ABNF as `%x2318`.
 
 
 
-# JSON Path Syntax and Semantics
+# JSONPath Syntax and Semantics
 
 ## Overview
 
-A JSON Path is a string which selects zero or more nodes of any JSON document.
-A valid JSON Path conforms to the ABNF syntax defined by this document.
+A JSONPath is a string which selects zero or more nodes of a piece of JSON.
+A valid JSONPath conforms to the ABNF syntax defined by this document.
 
-A JSON Path MUST be encoded using UTF-8. To parse a JSON Path according to
+A JSONPath MUST be encoded using UTF-8. To parse a JSONPath according to
 the grammar in this document, its UTF-8 form SHOULD first be decoded into
 Unicode code points as described
 in {{RFC3629}}.
@@ -137,8 +137,7 @@ in {{RFC3629}}.
 
 ## Terminology
 
-A JSON document is logically a tree of nodes with the document as the root
-node of the tree.
+A JSON value is logically a tree of nodes.
 
 Each node holds a JSON value (as defined by {{RFC8259}}) of one of the
 types object, array, number, string, or one of the literals `true`,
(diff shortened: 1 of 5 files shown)


draft-normington-jsonpath-latest.md
```diff
@@ -425,22 +428,25 @@ An array index is a union element consisting of an integer (in base 10).
 ~~~~ abnf
 array-index = integer
 
-integer = [%x2D] (%x30 / (%x31-39 *%x30-39))
+integer = ["-"] ("0" / (%x31-39 *%x30-39))
```
cabo, line 431: You may want to make use of DIGIT and DIGIT1 here

draft-normington-jsonpath-latest.md
```diff
@@ -425,22 +428,25 @@ An array index is a union element consisting of an integer (in base 10).
...
                             ; optional - followed by 0 or
                             ; sequence of digits with no leading zero
+DIGIT1 = %x31-39            ; non-zero digit
 ~~~~
 
 Note: the syntax does not allow integers with leading zeros such as `01` and `-01`.
 
-An array slice is a union element consisting of two or three integers (in
-base 10 and which may be omitted) separated by colons.
+An array slice is a union element consisting of optional integers (in base 10) separated by colons.
```
cabo, line 439: I think we need to be more aware about saying "is" when we mean "is represented by".

cabo (commented):
Quick look...

mkmik (approved):

draft-normington-jsonpath-latest.md
glyn: Actually, I mean "is" rather than "is represented by". But the phrase "is a union element consisting of" is causing a lot of grief. It probably is better to make the array slice description independent of where it is used, thus:
```suggestion
An array slice consists of optional integers (in base 10) separated by colons.
```

draft-normington-jsonpath-latest.md
glyn: Whoops. Of course.

gregsdennis, line 431: What was the guidance around using quoted characters vs their hex equivalent?

draft-normington-jsonpath-latest.md
gregsdennis, line 439: Why remove the count here?  Should we specify that it is three integers, each of which may be omitted?

draft-normington-jsonpath-latest.md
```diff
@@ -425,22 +428,25 @@ An array index is a union element consisting of an integer (in base 10).
...
 
 ~~~~ abnf
-array-slice = [ integer ] ws %x3A ws [ integer ]
-                   [ ws %x3A ws [ integer ] ]
-                            ; start:end or start:end:step
+array-slice = [ start ] ws ":" ws [ end ]
+                   [ ws ":" ws [ step ] ]
+start = integer
+end = integer
+step = integer
```
gregsdennis, line 446: Does the markdown translation align the `=`?

gregsdennis (changes requested):
Couple questions and a suggestion

draft-normington-jsonpath-latest.md
glyn: That change was based on input from @cabo.

draft-normington-jsonpath-latest.md
glyn: Good point. Added "three".

draft-normington-jsonpath-latest.md
glyn: No. Does that matter?

draft-normington-jsonpath-latest.md
gregsdennis: Could you link to the input?  Why have both?

draft-normington-jsonpath-latest.md
gregsdennis: Typically I see them aligned for readability, but technically, no.

glyn: I suggest we leave that for now and do a fine-tuning editorial pass later for consistency.

draft-normington-jsonpath-latest.md
glyn: @cabo's [guidance](https://mailarchive.ietf.org/arch/msg/jsonpath/jip4OCSgTPH1LZYmL1iMCgxFiMk/) was to use hex for ranges and otherwise use quoted characters except where case sensitivity is needed.

glyn:
> Couple questions and a suggestion

I'd like to merge so that @cabo can proceed to raise PRs which add introductory material from the Gössner draft. None of this is set in concrete, so there will be ample opportunities to revisit the text.

glyn:
@gregsdennis Merged. Let me know if you have outstanding issues and we can address them separately.

PR #42, opened 2020-11-26
Add introductory text from draft-goessner

cabo:
Start work on a merge of the two drafts.


draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+    "bicycle": {
+      "color": "red",
+      "price": 19.95
+    }
+  }
+}
+~~~~
+{: fig-example-item title="Example JSON data item"}
+
+The examples in {{tbl-example}} presume an underlying script language
```
glyn, line 351: Let's avoid underlying scripting language dependencies here too.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+| `/store/*`             | `$.store.*`                               | all things in store, which are some books and a red bicycle. |
+| `/store//price`        | `$.store..price`                          | the price of everything in the store.                        |
+| `//book[3]`            | `$..book[2]`                              | the third book                                               |
+| `//book[last()]`       | `$..book[(@.length-1)]`<br>`$..book[-1:]` | the last book in order.                                      |
+| `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
+| `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
+| `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheapier than 10                            |
+| `//*`                  | `$..*`                                    | all Elements in XML document. All members of JSON data item. |
```
glyn, line 367: I'm not sure that there is much value in keeping the first column given that JSONPath is already established.

Please can we make the result descriptions consistently end or not end with a period. I prefer to avoid the period.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+that allows obtaining the number of items in an array, testing for the
+presence of a map member, and performing numeric comparisons of map
+member values with a constant.
+
+| XPath                  | JSONPath                                  | Result                                                       |
+|------------------------|-------------------------------------------|--------------------------------------------------------------|
+| `/store/book/author`   | `$.store.book[*].author`                  | the authors of all books in the store                        |
+| `//author`             | `$..author`                               | all authors                                                  |
+| `/store/*`             | `$.store.*`                               | all things in store, which are some books and a red bicycle. |
+| `/store//price`        | `$.store..price`                          | the price of everything in the store.                        |
```
glyn, line 361:
```suggestion
| `/store//price`        | `$.store..price`                          | the prices of everything in the store.                        |
```

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+| `//book[3]`            | `$..book[2]`                              | the third book                                               |
+| `//book[last()]`       | `$..book[(@.length-1)]`<br>`$..book[-1:]` | the last book in order.                                      |
```
glyn, line 363: Why not more simply:
```suggestion
| `//book[last()]`       | `$..book[(@.length-1)]`<br>`$..book[-1]` | the last book in order.                                      |
```

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+| `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
+| `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
+| `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheapier than 10                            |
+| `//*`                  | `$..*`                                    | all Elements in XML document. All members of JSON data item. |
+{: #tbl-example title="Example JSONPath expressions applied to the example JSON data item"}
+
+<!-- back to normington draft; not yet merged up where needed (e.g., terminology). -->
```
glyn, line 370: Good. When do we propose to merge terminology? Before or after offering the merged document to the IETF JSONPath Working Group?

glyn (changes requested):
I like the general direction. I'd like to remove any reference to the mythical "underlying scripting engine". I've also suggested some editorial improvements.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+  When a JSON Object as defined in {{-json}} is meant, we specifically
+  say JSON Object.
+
+Query:
+: Short name for JSONPath expression.
+
+Argument:
+: Short name for the JSON data item a JSONPath expression is applied to.
+
+Output Path:
```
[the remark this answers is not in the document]
mkmik: I'm in two minds. One of the main advantages of targeting the jsonpath specification at describing how an engine produces a sequence of _output paths_ as opposed to a sequence of _values_, is that it actually simplifies discussing about what a "value" actually is and removes potential causes of confusion that may arise in case an engine uses a compressed representation of a json tree, e.g. one where repeated subtrees may be actually reused.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+2007-02-21 {{JSONPath-orig}} and provides a more normative definition for it.
+In its current state, it is a strawman document showing what needs to
+be covered.
 
 JSON is defined by {{RFC8259}}.
 
-## Requirements Language
+JSONPath is not intended as a replacement, but as a more powerful
+companion, to JSON Pointer {{RFC6901}}. \[insert reference to section
+where the relationship is detailed.]
```
gregsdennis, line 148: Referencing Pointer is good, but I wonder if we can do more. The purposes of the two syntaxes are different.  Pointer is for isolating a single location within a document. Path is a query syntax which pulls multiple locations.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+
+or the *bracket–notation*
+
+~~~~
+$['store']['book'][0]['title']
+~~~~
+
+for paths input to a JSONPath processor.  Where a JSONPath processor
+uses JSONPath expressions for internal purposes or as output paths,
+these will always be converted to the more general *bracket–notation*.
```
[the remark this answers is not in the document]
gregsdennis: Your suggestions still mention "internal purposes."  I think this needs to left out entirely.  My implementation doesn't use any notation internally; I have a data model.

This spec shouldn't concern itself with the internal working of implementations.  It should be more of a functional requirement set.  As long as it gives the expected output, it should be compliant, regardless of how it got there.  (This isn't high school algebra where you have to show your work.)

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+
+JSONPath allows the wildcard symbol `*` for member names and array
+indices. It borrows the descendant operator `..` from {{E4X}} and
+the array slice syntax proposal `[start:end:step]` {{SLICE}} from ECMASCRIPT 4.
+
+JSONPath can employ an *underlying scripting language*, expressions of
```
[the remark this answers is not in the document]
gregsdennis: According the underlying scripting language is good, but it means that we'll have to eventually define what's valid in expressions. (Can of worms, I know.)

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+
+<!-- note that the weirdness about the vertical bar above is intentional -->
+
+XPath has a lot more to offer (location paths in unabbreviated syntax,
+operators and functions) than listed here.  Moreover there is a
+significant difference how the subscript operator works in Xpath and
+JSONPath:
+
+* Square brackets in XPath expressions always operate on the *node set* resulting from the previous path fragment. Indices always start at 1.
+* With JSONPath, square brackets operate on the *object* or *array* addressed by the previous path fragment. Indices always start at 0.
```
gregsdennis, line 308: JSON objects are specifically unordered, and many languages simply _can't_ reference them in the serialized sequence (Ruby comes to mind.)  We went over and over this recently in JSON Schema.

gregsdennis (changes requested):

draft-normington-jsonpath-latest.md
cabo: The wording her certainly can be improved.  The existing text tries to emphasize that array indices start at 1 in XPath and at 0 in JSONPath, it certainly is not trying to invent numeric indices for maps (JSON objects).

Thanks for a good set of actionable proposals for improvement!  I'll submit an update on Wednesday.

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
cabo: Now, is that a suggestion or a promise?  I'll redundantly leave both sentences in for now.

draft-normington-jsonpath-latest.md
cabo: I think we don't need to have terminology adjusted before proposing the merged document for adoption.  But we can start any discussion we need now.

draft-normington-jsonpath-latest.md
cabo: I'm not trying to address this in this version yet, so I just added a note.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,266 @@ informative:
...
+Query:
+: Short name for JSONPath expression.
+
+Argument:
+: Short name for the JSON data item a JSONPath expression is applied to.
+
+Output Path:
+: A simple form of JSONPath expression that identifies a Position by
+  providing a query that results in exactly that position.  Similar
+  to, but syntactically different from, a JSON Pointer {{-pointer}}.
```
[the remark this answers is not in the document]
cabo: We could.  There are subtle differences we'd need to elaborate first to do this.

draft-normington-jsonpath-latest.md
cabo: Long Wednesday.  Please check my proposed updates.

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
danielaparker: I don't think you can use JSONPointer for that.  JSONPointer only allows for locating primitive values, it doesn't work for empty arrays or empty objects.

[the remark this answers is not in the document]
danielaparker: But can a JSONPointer identify an empty object or empty array? My understanding of https://tools.ietf.org/html/rfc6901#section-7 is that that is not defined.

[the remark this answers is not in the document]
cabo: Sure.  For `{ "a": {} }`, `/a` points to the empty object.  What in Section 7 makes you think it doesn't?

draft-normington-jsonpath-latest.md
gregsdennis: I'd like to get terminology nailed down sooner rather than later, but it doesn't have to be part of this PR.

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
gregsdennis: Specifically, JSON Schema uses JSON Pointers in its `$ref` keyword to reference a location within the schema in its JSON form.  Per JSON Schema, the reference MUST resolve to a valid schema object.  However, `{}` is a perfectly valid object.

The only time Section 7 applies is when the pointer indicates a route _into_ this object, e.g. `/a/missing` in @cabo's example would be required to error because the value at `/a` doesn't contain a `missing` property.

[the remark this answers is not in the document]
cabo: RFC 6901 goes to lengths to make sure any JSON map key (which essentially is any Unicode string) can be used as a path component.  It decodes that (needing an escape for / and then for the escape), and then looks whether it is looking at an array or a map.  Any maps is directly indexed, while for an array the path component is interpreted as a decimal number (there also is the "-" special, which I think we don't need).  The latter is easy.  Doing the former requires having a bracket/string syntax for any Unicode string.

draft-normington-jsonpath-latest.md
gregsdennis: I think specifying "array indices" is sufficient here. 👍

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,282 @@ informative:
...
+
+--- note_Contributing
+
+This document picks up the popular JSONPath specification dated
+2007-02-21 and provides a normative definition for it.
+In its current state, it is a strawman document showing what needs to
+be covered.
+
+Comments and issues can be directed at the github repository *insert
+repo here* as well as (for the time when the more permanent home is
```
gregsdennis, line 132: Was this an unfulfilled todo for this PR, or is it supposed to remain open for now?

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
gregsdennis: Maybe just something to consider for the future, but not relevant for this PR, then.

draft-normington-jsonpath-latest.md
gregsdennis: I think the first column is helpful for those familiar with XPath, and helps to coalesce the concepts from the preceding table.

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
danielaparker: You're right, thanks for the correction.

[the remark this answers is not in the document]
gregsdennis: I've opened #44 for this.

draft-normington-jsonpath-latest.md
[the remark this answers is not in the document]
gregsdennis: Opened #44 to discuss whether we need output for the initial draft.  I vote we leave it for now.

draft-normington-jsonpath-latest.md
```diff
@@ -84,43 +108,282 @@ informative:
...
+| `/store/*`             | `$.store.*`                               | all things in store, which are some books and a red bicycle. |
+| `/store//price`        | `$.store..price`                          | the prices of everything in the store.                       |
+| `//book[3]`            | `$..book[2]`                              | the third book                                               |
+| `//book[last()]`       | `$..book[(@.length-1)]`<br>`$..book[-1]`  | the last book in order.                                      |
+| `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
+| `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
+| `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheapier than 10                            |
+| `//*`                  | `$..*`                                    | all Elements in XML document. All members of JSON data item. |
```
glyn, line 383: Periods removed:
```suggestion
|------------------------|-------------------------------------------|--------------------------------------------------------------|
| `/store/book/author`   | `$.store.book[*].author`                  | the authors of all books in the store                        |
| `//author`             | `$..author`                               | all authors                                                  |
| `/store/*`             | `$.store.*`                               | all things in store, which are some books and a red bicycle  |
| `/store//price`        | `$.store..price`                          | the prices of everything in the store                        |
| `//book[3]`            | `$..book[2]`                              | the third book                                               |
| `//book[last()]`       | `$..book[(@.length-1)]`<br>`$..book[-1]`  | the last book in order                                       |
| `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
| `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
| `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheaper than 10                             |
| `//*`                  | `$..*`                                    | all Elements in XML document; all members of JSON data item  |
```

glyn (approved):
LGTM. I made one more group of suggestions to remove some periods in a table.

Please could you answer @gregsdennis's blocking comment and then I'll merge.

draft-normington-jsonpath-latest.md
cabo: I already did this, with some more editorial fixes, in https://github.com/jsonpath-standard/internet-draft/pull/42/commits/11adbd2

draft-normington-jsonpath-latest.md
cabo: I would put that in when the document is at its final repo.

cabo:
> LGTM. I made one more group of suggestions to remove some periods in a table.
> 
> Please could you answer @gregsdennis's blocking comment and then I'll merge.

Which one is that?

draft-normington-jsonpath-latest.md
glyn: Thanks @cabo. That's what I expected.

glyn:
> > LGTM. I made one more group of suggestions to remove some periods in a table.
> > Please could you answer @gregsdennis's blocking comment and then I'll merge.
> 
> Which one is that?

You just answered it, thanks.

draft-normington-jsonpath-latest.md
glyn: Er, I don't see those changes in https://github.com/jsonpath-standard/internet-draft/commit/11adbd2.

cabo: Oops.  Fixed in 884bb0f

glyn: Thanks. Please could you regenerate too?

cabo: Done.

PR #46, opened 2020-12-07
More editorial fixes

cabo:
Daniel Parker's comments
Fix ABNF (OK, that is not editorial) and dehexify


draft-normington-jsonpath-latest.md
```diff
@@ -244,9 +252,8 @@ The JSONPath tool in question should:
 
 JSONPath expressions always apply to a JSON data item in the same way
 as XPath expressions are used in combination with an XML document.
-Since a JSON data item is usually anonymous and doesn't necessarily
-have a "root member object", JSONPath used the abstract name `$` to
-refer to the top level object of the data item.
+Since a JSON data item is anonymous JSONPath uses the abstract name `$` to
+refer to the top level data item of the argument.
```
glyn, line 256: Is "the argument" well defined at this point in the document? It's slightly ambiguous since there are two inputs: the JSONPath itself and the JSON value.

Also, I suggest adding a comma, thus: "Since a JSON data item is anonymous, JSONPath uses...".

(Sorry I can't use a github suggestion here as I'm reviewing the commits separately.)

draft-normington-jsonpath-latest.md
```diff
@@ -0,0 +1,877 @@
+---
+stand_alone: true
+ipr: trust200902
+docname: draft-normington-jsonpath-00
```
glyn, line 4: Should be -latest, for consistency.

glyn (changes requested):
Please address my comments and delete -00.md, update the script/README as necessary for the renaming back to -latest, and regenerate.

draft-normington-jsonpath-latest.md
cabo: There is a definition for "argument".  Of course, we could introduce this once more in the text, but maybe not here.
(And if you have a better name for the JSON data item under consideration by the JSONPath query, I'm all ears.)

glyn: Oops, missed that definition for some reason. Thanks.

glyn:
I rewrote history after checking with @cabo. HEAD now points at the commit _before_ I renamed to -00. (The renaming is tagged as draft-normington-jsonpath-00.) So this PR should look like a set of changes on the *-latest.* files.

scripts/gen.sh
```diff
@@ -7,8 +7,8 @@ readonly script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
 cd "$script_dir/.."
 
-kdrfc -h3 draft-normington-jsonpath-latest.md && \
-xmlstarlet sel -T -t -v '//artwork[@type="abnf"]' ./draft-normington-jsonpath-latest.xml 2>/dev/null \
+kdrfc -h3 draft-normington-jsonpath-00.md && \
+xmlstarlet sel -T -t -v '//artwork[@type="abnf"]' ./draft-normington-jsonpath-00.xml 2>/dev/null \
 | aex \
-| bap -S path -q && cp draft-normington-jsonpath-latest.html docs/index.html
+| bap -S path -q && cp draft-normington-jsonpath-00.html docs/index.html
 )
```
glyn, line 14: This needs reworking to match the new naming scheme `draft-normington-jsonpath.*`.

glyn:
@cabo The README also needs changing to reflect the new naming scheme `draft-normington-jsonpath.*`.

glyn (commented):
My changes have been or are being addressed. Since I'm away until 16 December, I'm going to abstain and leave it to others to approve/merge. Otherwise, I'll pick this up when I'm back.

draft-normington-jsonpath.md
```diff
@@ -166,14 +166,22 @@ The terminology of {{-json}} applies.
...
+  of atomic data, namely null, true, false, numbers, and text strings.
+  Also called a JSON value.
 
 Object:
 : Used in its generic sense, e.g., for programming language objects.
   When a JSON Object as defined in {{-json}} is meant, we specifically
   say JSON Object.
 
+Member:
+: A key/value pair in a map (JSON Object).  (Not itself a JSON value.)
```
gregsdennis, line 179: Perhaps we should define "key" and even declare "property" as a synonym.

draft-normington-jsonpath.md
```diff
@@ -263,7 +270,8 @@ $['store']['book'][0]['title']
 for paths input to a JSONPath processor.
 \[1]
 Where a JSONPath processor uses JSONPath expressions as output paths,
-these will always be converted to the more general *bracket–notation*.
+these will always be converted to normalized JSONPath expressions
+which employ the more general *bracket–notation*.
 \[2]
 Bracket notation is more general than dot notation and can serve as a
 canonical form when a JSONPath processor uses JSONPath expressions as
```
gregsdennis, line 277: Do we want to avoid the use of "expressions" here so that it's not confused with query expressions (defined on line 290)?

draft-normington-jsonpath.md
```diff
@@ -286,7 +294,7 @@ an alternative to explicit names or indices as in:
 $.store.book[(@.length-1)].title
 ~~~~
 
-The symbol `@` is used for the current object.
+The symbol `@` is used for the current item.
```
gregsdennis, line 297: "element" instead of "item" to align with the definition.

draft-normington-jsonpath.md
```diff
@@ -297,15 +305,15 @@ Here is a complete overview and a side by side comparison of the JSONPath syntax
 
 | XPath | JSONPath           | Description                                                                                                                           |
 |-------+--------------------+---------------------------------------------------------------------------------------------------------------------------------------|
-| `/`   | `$`                | the root object/element                                                                                                               |
-| `.`   | `@`                | the current object/element                                                                                                            |
+| `/`   | `$`                | the root element/item                                                                                                                 |
```
gregsdennis, line 308: Why do we need "item"?  "Element" is strictly defined.  Maybe just use that.

draft-normington-jsonpath.md
```diff
@@ -403,7 +411,7 @@ in {{RFC3629}}.
 A JSON value is logically a tree of nodes.
 
 Each node holds a JSON value (as defined by {{RFC8259}}) of one of the
-types object, array, number, string, or one of the literals `true`,
+types JSON object, array, number, string, or one of the literals `true`,
```
gregsdennis, line 414: Why are we specifying JSON for objects everywhere?  What other meaning does "object" have that warrants this specificity?

gregsdennis (changes requested):

draft-normington-jsonpath.md
cabo: Property is a synonym for member, as is entry and field.
How many of these synonyms do we need?
(RFC 8259 calls the keys "names", which is not better, except maybe in combinations such as "field name" or "member name".)

gregsdennis: "key" and "property" are by far the most common terms I see in use.

cabo: And I've been caught calling keys "labels".  Some people like to call them "tags".
My first hit when searching was https://www.w3schools.com/Js/js_json_objects.asp, which you may like or not like (they do confuse JSON objects with their representation), but for them introducing JSON objects as key/value pairs seemed obvious, not needing further explanation.

scripts/gen.sh
cabo: Done in e28d965

draft-normington-jsonpath.md
mkmik: I agree. Also other primitives like number and string are also _JSON_ numbers and _JSON_ strings, (with all the subtleties that come from that, both from spec and underlying json engine implementation choices allowed by the JSON spec)

draft-normington-jsonpath.md
mkmik: Key/value is indeed very common. "field" is less common but nevertheless attested in the wild.

I tend to prefer `field name/value` mainly for two reasons: a) it builds on top of RFC 8259 terminology b) the "field" qualifier makes sense also when referring to the value (the _field value_)

scripts/gen.sh
glyn: Thanks. For the record, I marginally prefer the explicit `-latest` suffix as it looks like a placeholder waiting to be replaced by a version number. But I can certainly live with this alternative.

draft-normington-jsonpath.md
```diff
@@ -240,13 +252,12 @@ The JSONPath tool in question should:
 * be lightweight in code size and memory consumption.
 * be runtime efficient.
 
-## Overview of JSONPath Expressions
+## Overview of JSONPath Expressions {#overview}
```
glyn, line 255: The anchor `#overview` clashes with the default anchor for the _other_ `Overview` section.

draft-normington-jsonpath.md
```diff
@@ -188,7 +196,11 @@ Output Path:
 Position:
 : A JSON data item identical to or nested within the JSON data item to
   which the query is applied to, expressed either by the value of that
-  data item or by providing a JSONPath Output Path.
+  data item or by providing a Normalized Path Expression as a JSONPath Output Path.
+
+Normalized Path Expression:
+: A query in a normalized form that identifies exactly one Position in
```
glyn, line 202: Normalized form is not yet well defined. Is it intended to be [canonical](https://en.wikipedia.org/wiki/Canonical_form) too or simply normalized (but not necessarily unique among normalized alternatives)?

draft-normington-jsonpath.md
danielaparker: In the spirit of "you can't please everyone", my own opinion from an implementer's viewpoint is that use of synonyms is best avoided. Implementer's want a parsimonious vocabulary with precise definitions, preferably one that's consistent with the core specifications (RFC 8259), the introduction of synonyms which are "suggestive" only frightens them. Words mean different things in different contexts, for example, "field" in the Java specifications refers to a type and an identifier, in a CODASYL data record it means something else. "property" isn't mentioned in RFC 8259, why introduce it here?      

Regarding "perhaps we should define 'key'", if it appears in the ABNF, it is defined.

cabo: It probably would help to keep "name" around for names in jsonpath.
But given that "key" is not defined in 8259, I'll stick with "name" for now, expanding to "member name" when needed.

draft-normington-jsonpath.md
cabo: This is one of many places that uses "jsonpath expression".  We could all rename those into "query".  Holding off with that right now.

draft-normington-jsonpath.md
cabo: The slash is XML/JSON.  Element on the JSON side means something different than element on the XML side.

draft-normington-jsonpath.md
cabo: I changed them all to object.  I continue to believe this is bad usage, but if the spec is mainly directed at people who work exclusively with JSON, that may be a simplification.

draft-normington-jsonpath.md
cabo: No, that's why it points to a section that explains in which way this is normalized.

draft-normington-jsonpath.md
gregsdennis: Maybe that distinction could be made clearer (later).

draft-normington-jsonpath.md
gregsdennis: My feeling is that this is a JSON-related specification, so it's likely to be assumed.  We _could_ add a statement in the declarations at the beginning to the effect of, "All usages of 'object', etc, mean the JSON usage unless otherwise specified."

gregsdennis (approved):
I'm happy with this as an intermediate state to be merged so that we don't block further development.

glyn (approved):
Yes, let's merge this to allow forward progress to resume.

PR #48, opened 2020-12-23
Point the draft at the official draft git repo

glyn:


draft-normington-jsonpath-latest.md
```diff
@@ -331,7 +329,7 @@ typical XML example representing a bookstore (that also has bicycles):
 
 ~~~~json
 { "store": {
-    "book": [
```
glyn, line 334: This change, to trim trailing whitespace, was made courtesy of `.editorconfig`.

cabo (approved):

cabo:
I think we should move to the Martin Thomson template for repositories soon.
Until then: Do we really need a copy of the HTML in index.html?

glyn:
> I think we should move to the Martin Thomson template for repositories soon.
> Until then: Do we really need a copy of the HTML in index.html?

Yes, so we have a convenient rendered version (at https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-jsonpath/). github pages is pretty restrictive about where the source is located and if it was in the root directory, it would probably need to be named something like `README.md` which wouldn't fit very well with our file names and generation process. If we can find a solution which avoids a copy and which still offers a rendered version, that would be great.

glyn:
@cabo Thanks for approving. I'm going to hold off merging for 24 hours as this is the protocol we've tried to observe in the past so others have a chance to review/comment.

cabo:
So where does the current github.io structure leave the topic branch HTMLs?

BTW, maybe we can avoid making a references entry for the github repo and put the link directly into the note.

glyn:
> So where does the current github.io structure leave the topic branch HTMLs?

High and dry, I'm afraid. Topic branches are not served by github pages.
 
> BTW, maybe we can avoid making a references entry for the github repo and put the link directly into the note.

Done.

cabo:
On 2020-12-23, at 15:00, Glyn Normington <notifications@github.com> wrote:
> 
> High and dry, I'm afraid. Topic branches are not served by github pages.

OK, they are with Martin Thomson’s template, e.g. see 

https://cbor-wg.github.io/CBORbis/edits-from-paul/draft-ietf-cbor-7049bis.html

for a random example.  This also shows a much better usage of index.html:

https://cbor-wg.github.io/CBORbis/

The only reason I haven’t sent a PR with Martin’s template that I haven’t figured out the new black for CI (github actions).  But we don’t have CI now either, so maybe we should start with the template and put in CI when we have figured it out (until then it’s `make; git push --all`).

Grüße, Carsten

glyn:
@cabo Martin's template looks great if we are comfortable giving contributors write access to their own branches in this repo (rather than have them use their own forks). I suggest we get PR 46 merged first so we don't lose track of the outstanding comment threads, then apply Martin's template, and then add CI.

PR #72, opened 2021-03-13
Align (and simplify) terminology.

### LINKED ISSUE #66: Clarification on Terminology
(This was original raised in #54 and now being split out)

> My own view is that the terminology should stay consistent with RFC
8259, and that the word "object" should not be used for items that are
not JSON objects in the sense of RFC 8259.
*(Daniel P)*

> To Carsten's point about what we call things, the number of distinguished
terms per RFC8259 is pretty small: JSON text, value, object, array, number,
string.  Having spent quite a bit of time specifying JSON DSLs, I find that
using just those terms doesn't seem to get in the way or cause problems, so
I'd argue that we should stick to them (and build up to higher-level
constructs as required for JSONPath).
>
> … oh, and I forgot the very useful "member".
*(Tim Bray)*

> … and “element” (the things in arrays). *(Carsten Bormann)*

> The problem with JSON value is that it also can be quite confusing due to the usual use of that term.  Pointing to a tree and saying “the values inside that tree” is not going to be felt as equivalent to “the set of all subtrees of that tree, including the tree itself”.  But if JSON value is the only term we have, it has to be.  Hence my preference to talk about data items when I mean the items themselves and not their “value”.
*(Carsten Bormann)*

> I think the key difficulty is whether each (key, value) pair in an object is "a thing" that can be identified and manipulated and potentially returned. (If we're talking analogies, then it's analogous to an attribute node in the XDM model).
*(Michael Kay)*

> ECMA-404 uses "name/value pair", which is what I understand the term
"member" to mean (Douglas Crockford uses "member").
*(Daniel P)*

> I think the term “union” is poor. If we think of it as concatenation of results, then the result is as expected.
*(Glyn Normington)*

I understand, that within RFC8259 we have JSON values of different types. They are structured somehow, which is not so much of interest here.

But while querying that structure with JSONPath it is vitally important to identify that hierarchical structure as a tree. So in fact we build up a higher-level construct here. We also need to call "the things" in the tree somehow. I was able to identify

* "node" or "item" of a tree
* "member" of an object
* "name/value" or "key/value" pair alias "member"
* "element" of an array

but could not see an agreement here.

I agree to Glyn calling the term "union" poor (s. below).

cabo:
I created a pull request so we have something concrete to discuss to close #66.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -201,12 +201,6 @@ Query:
 Argument:
 : Short name for the JSON data item a JSONPath expression is applied to.
 
-Nodelist:
-: Output of applying a query to an argument: a list of nodes within
-  that argument.
-  While this list can be represented in JSON, e.g. as an array, the
-  nodelist is an abstract concept unrelated to JSON data items.
-
 Output Path:
 : A simple form of JSONPath expression that identifies a node by
   providing a query that results in exactly that node.  Similar
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -168,8 +168,7 @@ Data Item:
 
   The term _node_ has the same meaning, but is used to emphasize the
   location of an item within the Argument, i.e., an item identical to
-  or nested within the JSON data item to which the query is applied
-  to.
+  or nested within the JSON data item to which the query is applied.
 
 Object:
 : A JSON object as defined in {{-json}}.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -201,6 +201,12 @@ Query:
 Argument:
 : Short name for the JSON data item a JSONPath expression is applied to.
 
+Nodelist:
+: Output of applying a query to an argument: a list of nodes within
+  that argument.
+  While this list can be represented in JSON, e.g. as an array, the
+  nodelist is an abstract concept unrelated to JSON data items.
+
 Output Path:
 : A simple form of JSONPath expression that identifies a node by
   providing a query that results in exactly that node.  Similar
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -190,7 +190,7 @@ Element:
   context for XML elements.)
 
 Index:
-: An unsigned number that identifies a specific element in an array.
+: A non-negative integer that identifies a specific element in an array.
   The term, in particular in the form indexing, is also sometimes used
   for either an index or a member name, when both arrays and objects
   can be fed to the indexing operation.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -164,6 +164,12 @@ Data Item:
   composed of containers, namely JSON objects and arrays, and
   of atomic data, namely null, true, false, numbers, and text strings.
   Also called a JSON value.
+  In this document, often abbreviated to _item_.
+
+  The term _node_ has the same meaning, but is used to emphasize the
+  location of an item within the Argument, i.e., an item identical to
+  or nested within the JSON data item to which the query is applied
+  to.
 
 Object:
 : A JSON object as defined in {{-json}}.
@@ -176,9 +182,19 @@ Name:
 : The name in a name/value pair constituting a member.  (Also known as
   "key", "tag", or "label".)
 
+Array:
+: A JSON array as defined in {{-json}}.
+  Never used in its generic sense, e.g., for programming language objects.
+
 Element:
 : An item in an array.  (Also used with a distinct meaning in XML
   context for XML elements.)
+
+Index:
+: An unsigned number that identifies a specific element in an array.
+  The term, in particular in the form indexing, is also sometimes used
+  for either an index or a member name, when both arrays and objects
+  can be fed to the indexing operation.
 
 Query:
 : Short name for JSONPath expression.
@@ -187,18 +203,9 @@ Argument:
 : Short name for the JSON data item a JSONPath expression is applied to.
 
 Output Path:
-: A simple form of JSONPath expression that identifies a Position by
-  providing a query that results in exactly that position.  Similar
+: A simple form of JSONPath expression that identifies a node by
+  providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
-
-Position:
-: A JSON data item identical to or nested within the JSON data item to
-  which the query is applied to, expressed either by the value of that
-  data item or by providing a Normalized Path Expression as a JSONPath Output Path.
-
-Normalized Path Expression:
-: A query in a normalized form that identifies exactly one Position in
-  an Argument; see {{overview}}.
 
 
 ## Inspired by XPath
@@ -272,7 +279,7 @@ $['store']['book'][0]['title']
 for paths input to a JSONPath processor.
 \[1]
 Where a JSONPath processor uses JSONPath expressions as output paths,
-these will always be converted to normalized JSONPath expressions
+these will always be converted to Output Paths
 which employ the more general *bracket–notation*.
 \[2]
 Bracket notation is more general than dot notation and can serve as a
@@ -785,7 +792,8 @@ as follows:
 {: title="Default array slice start and end values"}
 
 Slice expression parameters `start` and `end` are not directly usable
-as slice bounds and must first be normalized. Normalization is defined as:
+as slice bounds and must first be normalized.
+Normalization for this purpose is defined as:
 
 ~~~~
 FUNCTION Normalize(i):


draft-ietf-jsonpath-base.md
```diff
@@ -164,6 +164,12 @@ Data Item:
   composed of containers, namely JSON objects and arrays, and
   of atomic data, namely null, true, false, numbers, and text strings.
   Also called a JSON value.
+  In this document, often abbreviated to _item_.
+
+  The term _node_ has the same meaning, but is used to emphasize the
+  location of an item within the Argument, i.e., an item identical to
+  or nested within the JSON data item to which the query is applied
+  to.
```
glyn, line 172:
```suggestion
  or nested within the JSON data item to which the query is applied.
```

draft-ietf-jsonpath-base.md
```diff
@@ -176,30 +182,31 @@ Name:
...
+Array:
+: A JSON array as defined in {{-json}}.
+  Never used in its generic sense, e.g., for programming language objects.
+
 Element:
 : An item in an array.  (Also used with a distinct meaning in XML
   context for XML elements.)
 
+Index:
+: An unsigned number that identifies a specific element in an array.
```
glyn, line 193: Negative indices are also used in the spec.

cabo: OK, we'll need to differentiate between the index as in number of element and the index as in expression value supplied to indexing.  That bothered me already about the existing text.

goessner: Much better now ... Index definition is the real location and might not be confused with slice expression syntax.

goessner: `FUNCTION Normalize(i)` should have `len` consistently as second argument, i.e. `FUNCTION Normalize(i,len)` ...

cabo: Good catch; I have put the fix for Normalize into a separate PR #73

timbray (commented):
I think in an ideal world, it would be good to use only and exactly the 8259 terminology. "Value", "JSON Text", "Array", "Object".  The latest draft isn't fresh enough in my mind to have an opinion as to whether this is possible.

goessner (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -176,30 +181,37 @@ Name:
...
+  for either an index or a member name, when both arrays and objects
+  can be fed to the indexing operation.
+
 Query:
 : Short name for JSONPath expression.
 
 Argument:
 : Short name for the JSON data item a JSONPath expression is applied to.
 
+Nodelist:
```
gregsdennis, line 204: Please see my [comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/60#issuecomment-801313287).

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
```diff
@@ -164,6 +164,11 @@ Data Item:
   composed of containers, namely JSON objects and arrays, and
   of atomic data, namely null, true, false, numbers, and text strings.
   Also called a JSON value.
+  In this document, often abbreviated to _item_.
```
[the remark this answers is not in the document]
cabo: It is indeed not a sentence.

draft-ietf-jsonpath-base.md
```diff
@@ -176,30 +181,31 @@ Name:
...
+Array:
+: A JSON array as defined in {{-json}}.
+  Never used in its generic sense, e.g., for programming language objects.
+
 Element:
 : An item in an array.  (Also used with a distinct meaning in XML
   context for XML elements.)
 
+Index:
+: An unsigned number that identifies a specific element in an array.
```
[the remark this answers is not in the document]
cabo: I disagree.  I prefer the simpler term.  Would like to hear from others whether they also want to fall back to "non-negative".

draft-ietf-jsonpath-base.md
```diff
@@ -176,30 +181,31 @@ Name:
...
+: A simple form of JSONPath expression that identifies a node by
+  providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
-Position:
-: A JSON data item identical to or nested within the JSON data item to
-  which the query is applied to, expressed either by the value of that
-  data item or by providing a Normalized Path Expression as a JSONPath Output Path.
-
-Normalized Path Expression:
```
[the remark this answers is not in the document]
cabo: I don't think we need to define the term.  We already have one form of normalization in the document, the normalization that creates output paths is just another normalization.

draft-ietf-jsonpath-base.md
cabo: I have reread both comments now.

gregsdennis: Yeah, so my point of having you reread those comments was to illustrate my discomfort with the "nodelist" concept.  You have neither refuted those comments nor changed the PR, so this remains unresolved.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: Where is that definition? Would it not make sense to define it in... the definitions section?

draft-ietf-jsonpath-base.md
cabo: Some discussion about nodelist has happened.  Right now, the discussion in this WG is a bit fragmented; I'll try to dig this out when I have time to search for it.  Should have pointed to it in the first place...

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: The normalization we already have is in Section 3.6.2.4 (it is not about output paths, btw.)
I don't think that needs to be in the definitions, as it is local to that section.  That would probably also be the case when text about the normalization is written that leads to Output Paths.

draft-ietf-jsonpath-base.md
danielaparker: Just to note, I share with @gregsdennis some discomfort with the "nodelist" concept, and am skeptical that it's needed to describe JSONPath. I don't think the idea of a "nodelist" is required to describe the distinguishing feature of JSONPath, that it returns a collection of values (or references) to items that already exist in the original JSON document. We have from the [very beginning](https://goessner.net/articles/JsonPath/) the notion that "the return value of jsonPath is an array, which is also a valid JSON structure. So you might want to apply jsonPath to the resulting structure". Practically all implementations support something like:
```
JSON = jsonPath(JSON, path, value option)    
```
or 
```
JSON = jsonPath(JSON, path, path option)
```
With duplicates allowed, which is the case for legacy implementations, the computational model is simply to collect values (or their associated paths). Where is the role for "nodelists"?

With no duplicates, which is the case for some modern implementations, the computational model is to compute both values and their associated paths, and to keep path-value pairs with unique paths. Again, where is the role for "nodelists"?

In XPath, with it's vastly more complicated data model, I understand the need for "nodelist". But I don't understand why we need to take this XSLT idea and transpose it to JSONPath. It seems to me to be an unnecessary concept for describing JSONPath.

I would also suggest that the draft  isn't clear about what it means by "node". In a discussion in the terminology issue, @glyn suggested that if I wasn't familiar with trees and nodes, I could consult Wikipedia. But what's important is not what I understand or don't understand, what's important is what the draft says. In XPath specifications, the meaning of "node" is crystal clear, in the draft, it is not. I think that lack of clarity is somewhat related to the problem of making an argument for the "nodelist" concept that readers such as myself might find convincing.

I think there should be an issue raised for "nodelist, and the justification thereof".  Perhaps @cabo could start one?

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: I prefer non-negative for the reason @gregsdennis gave.

glyn (approved):
I left a comment or two, but am generally comfortable with these changes.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: I changed this as requested.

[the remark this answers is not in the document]
cabo: Actually, I made this change because I don't really care too much, but I don't agree with it at all.

People are using unsigned type because they care about the application data type (which is unsigned), only rarely because the bits are then laid out in a specific way.

There is a long history of computer languages struggling with naming unsigned types.  FORTRAN only had INTEGER, which was signed.  In the 1970s, languages started to address the issue, but didn't find terminology that they could use from mathematics.  "Natural" is both (1..) and (0..) in common usage, so it is ambiguous.  A number of languages used "Cardinal" for unsigned integers.  C led the pack with going for "unsigned", and that is the term that has stuck.  But, incredibly, as late as the mid 1990s languages were designed that didn't have unsigned types (e.g., Java).

To me, requiring the clumsy term "non-negative" for unsigned is a bit like requiring to say "octet" for byte, except that the latter was actually a reasonable thing until about 1981 (when non-8-bit bytes became extinct).

Anyway, I'm not arguing for using "unsigned" in this PR, because I want to make progress, but, for the record, that's what it *should* be.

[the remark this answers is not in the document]
glyn: @cabo I appreciate your flexibility.

A couple of things bothered me about using "unsigned":
* In JSONPath we are dealing with integers represented as strings rather than values of a numeric type.
* If we (and I hope we don't!) allowed indices such as "-0" and "+5", it would seem strange to describe these as "unsigned".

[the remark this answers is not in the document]
goessner: @cabo I enjoyed your entertaining nostalgic review of integer types.

@glyn I am not aware of the fact, that in JSONPath we are dealing with integers represented as strings exclusively.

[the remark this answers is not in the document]
cabo: We have two kinds of integers:

1. those coming from the JSON data item.  These are no longer in text form when we get to them; they are true numbers (no necessarily type-annotated to be integers).
2. those written in the JSONPath query.  I would expect that our processing model parses that query and feeds an AST to the process that already has true integers etc. as well.  But we haven't fully defined the processing model, so that is next.

goessner (approved):

glyn (approved):

draft-ietf-jsonpath-base.md
cabo: (I moved nodelist over to a new PR that I'm preparing.)

PR #73, opened 2021-03-13
Use second parameter, len, of Normalize throughout

cabo:
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/pull/72#discussion_r593778729


glyn (approved):

glyn:
Let's get this merged as it really is uncontroversial.

Issue #59, opened 2021-03-09
Relative paths support


bettio:
**Proposal**: JSONPath should have support to relative paths (to a current context), so it can be used as a building block for more advanced usages. 

XPath makes a clear distinction between relative and absolute paths, e.g.:

given the following XML input:
```xml
<root>
    <e>foo</e>
    <a>
        <b>
            <c><d>1</d><e>bar1</e></c>
            <c><d>2</d><e>bar2</e></c>
            <c><d>3</d><e>bar3</e></c>
        </b>
    </a>
</root>
```

and the following template, which uses the relative XPath `d` and the absolute XPath `/root/e`:

```xml
<?xml version="1.0"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
    <xsl:template match="/">
        <x>
            <xsl:for-each select="root/a/b/c">
                <item>
                    <y><xsl:value-of select="d"/></y>
                    <z><xsl:value-of select="/root/e"/></z>
                </item>
            </xsl:for-each>
        </x>
    </xsl:template>
</xsl:stylesheet>
```

the following XML can be produced:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<x>
    <item>
        <y>1</y>
        <z>foo</z>
    </item>
    <item>
        <y>2</y>
        <z>foo</z>
    </item>
    <item>
        <y>3</y>
        <z>foo</z>
    </item>
</x>
```

I think that in a similar fashion JSONPath should allow relative paths as a building feature for more advanced usages.

*Remark: I acknowledge that templating/transforms are out of scope here, however I believe that we must consider enabling more advanced usages.* 

**Suggested implementation**: JSONPath spec should mention that `$` is always the document root, while any JSONPath which doesn't start with '$' is relative to the current context.
This wouldn't change the behavior for several existing implementations since they can assume that current context == document root.
**Additional note**: There is also a number of implementations for which a JSONPath without starting `$` always evaluates to `[]`, given the previous definition we can state that their current context is always set to `null`, therefore this change would allow both as valid JSONPath spec implementations.

bettio:
> I think it already does, with paths starting with @, where @ represents the current node.

Actually it doesn't: the draft spec on this repo doesn't allow it at all, @goessner's spec doesn't mention it and [the comparison](https://cburgmer.github.io/json-path-comparison/) doesn't test it.

[...]

PR #75, opened 2021-03-19
WIP: allow relative paths

bettio:
*Draft not yet ready for merge or complete*
See also GH #59

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -391,6 +391,7 @@ constant.
 | `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
 | `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheaper than 10                             |
 | `//*`                  | `$..*`                                    | all elements in XML document; all members of JSON data item  |
+| `author`               | `@.author`                                | `"J. R. R. Tolkien"` when fourth book is the current item    |
 {: #tbl-example title="Example JSONPath expressions applied to the example JSON data item"}
 
 <!-- back to normington draft; not yet merged up where needed (e.g., terminology). -->
@@ -449,13 +450,16 @@ return an error.
 
 ## Syntax
 
-Syntactically, a JSONPath consists of a root selector (`$`), which
-selects the root node of a JSON value, followed by a possibly empty
-sequence of *selectors*.
+Syntactically, a JSONPath consists of either a root selector (`$`), which
+selects the root node of a JSON value, or a current item selector (`@`), which
+selects a child node of a JSON value, followed by a possibly empty sequence of
+*selectors*.
 
 ~~~~ abnf
-json-path = root-selector *selector
+json-path = start-selector *selector
+start-selector = root-selector / current-item-selector
 root-selector = "$"               ; $ selects document root node
+current-item-selector = "@"       ; @ selects current node
 ~~~~
 
 The syntax and semantics of each selector is defined below.
@@ -516,6 +520,18 @@ This is the concatenation of three lists, two of length one containing `0`, `1`,
 
 As a consequence of this approach, if any of the selectors selects no nodes,
 then the whole JSONPath selects no nodes.
+
+The current item selector `@` when used outside a filter behaves as the document
+root selector unless a child node is given to the evaluator.
+When a child node is used as current item, evaluator must select it as starting
+node instead of the document root when an expression starting with `@` is
+evaluated.
+
+When a child node is used as current item, the current item selector `@` selects
+it as starting node instead of the document root.
+
+The root selector `$` should be prefered to the current item selector `@` when
+a path relative to the current item is not required.
 
 In what follows, the semantics of each selector are defined for each type
 of node.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -450,10 +450,9 @@ return an error.
 
 ## Syntax
 
-Syntactically, a JSONPath consists of either a root selector (`$`), which
-selects the root node of a JSON value, or a current item selector (`@`), which
-selects a child node of a JSON value, followed by a possibly empty sequence of
-*selectors*.
+Syntactically, a JSON Path consists of a start selector, which may either be the
+root selector $ or the current item selector @, followed by a possibly empty
+sequence of *selectors*.
 
 ~~~~ abnf
 json-path = start-selector *selector
@@ -531,7 +530,8 @@ When a child node is used as current item, the current item selector `@` selects
 it as starting node instead of the document root.
 
 The root selector `$` should be prefered to the current item selector `@` when
-a path relative to the current item is not required.
+a path relative to the current item is not required or when no current item can
+be defined
 
 In what follows, the semantics of each selector are defined for each type
 of node.


draft-ietf-jsonpath-base.md
```diff
@@ -449,13 +449,15 @@ return an error.
 
 ## Syntax
 
-Syntactically, a JSONPath consists of a root selector (`$`), which
-selects the root node of a JSON value, followed by a possibly empty
-sequence of *selectors*.
+Syntactically, a JSONPath consists of either a root selector (`$`), which
+selects the root node of a JSON value, or a current item selector (`@`),
```
cabo, line 453: This doesn't say what a current item selector actually selects.
(The other text also needs to be fixed, for which I'll probably generate a separate PR.)

draft-ietf-jsonpath-base.md
```diff
@@ -449,13 +449,15 @@ return an error.
...
+followed by a possibly empty sequence of *selectors*.
 
 ~~~~ abnf
-json-path = root-selector *selector
+json-path = start-selector *selector
+start-selector = root-selector / current-item-selector
```
glyn, line 458: I'm not sure whether it makes sense to start a JSONPath with a current item selector as this is not "in the middle of" an array selection, for example, so what would the current item even mean?

Also, is there a consensus among existing implementations to support such a thing?

cabo: Right, the ABNF currently does not discuss nested queries, so there is no place where this could dock with.

draft-ietf-jsonpath-base.md
bettio: Thanks for your early feedback, remeber that it is still a draft / WIP.
It is not complete and it might contain typos and other mistakes.

draft-ietf-jsonpath-base.md
bettio: > I'm not sure whether it makes sense to start a JSONPath with a current item selector as this is not "in the middle of" an array selection, for example, so what would the current item even mean?

TL;DR: it is the item set when calling the JSONPath evaluator, it is quite useful if we are going to embed/use JSONPath with advanced tooling.  

Anyway please take a look to the following comments that explains that with further details:
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-797212814
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-797116919

Also supporting `@` for all JSONPaths has the good side effect of simplifying grammar when defining filter expressions (no special cases). 

> Also, is there a consensus among existing implementations to support such a thing?

Jayway supports that, which is an influential implementation (I think we should consider some kind of weighted consensus) and I wish to support this also on my implementation [ExJSONPath](https://github.com/ispirata/exjsonpath/).

TBH I didn't check this for all existing implementations, so further implementations supporting this syntax might exist.  

@glyn: Please, may you take a look to #59 that has all the information you are looking for.

glyn: Thanks. I've been following the discussion in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59 and so far I haven't seen a compelling need for this feature. It seems to increase the surface area of the spec in a way which, as @timbray mentioned, isn't compatible with our charter.

bettio: Actually it removes the need for any special case when using filters, since we may allow any jsonpath expression, simplifying grammar and existing implementations.
Also it has a XPath equivalent (which exists for good reasons), so we aren't introducing anything new here (I just added to this PR an example for it).

draft-ietf-jsonpath-base.md
```diff
@@ -517,6 +521,18 @@ This is the concatenation of three lists, two of length one containing `0`, `1`,
...
 then the whole JSONPath selects no nodes.
 
+The current item selector `@` when used outside a filter behaves as the document
+root selector unless a child node is given to the evaluator.
+When a child node is used as current item, evaluator must select it as starting
+node instead of the document root when an expression starting with `@` is
+evaluated.
+
+When a child node is used as current item, the current item selector `@` selects
+it as starting node instead of the document root.
```
glyn, line 531: What happens when a child node is used as the current item and the current item selector is being used inside a filter expression? I presume the current (e.g. array) item "wins", but currently the PR doesn't say.

bettio: The same that happens when doing, $a[?(@.b[?(@.c == "foo")])].
This need to be explained with or without this PR in the filter section.

draft-ietf-jsonpath-base.md
bettio: @glyn if you still don't agree about the whole PR I wish to discuss this on #59, otherwise if you agree with the proposed change but you wish to address any specific change contained here we can keep discussing here.
I rather like using this PR to discuss the changes not the concept.
So let me know if you still have any argument against this, and if you are still worried about increasing the surface area of the spec.

draft-ietf-jsonpath-base.md
bettio: @cabo let me know if the changes to this paragraph are now ok for you. Let me know if you have any further advice, otherwise I feel like we can resolve this conversation.

draft-ietf-jsonpath-base.md
gregsdennis: I think it might be helpful to introduce (and define) the idea of scope here.  Otherwise we should restrict against nested filter queries.

draft-ietf-jsonpath-base.md
glyn: @bettio Done: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-803263672

danielaparker: > I'm not sure whether it makes sense to start a JSONPath with a current item selector as this is not "in the middle of" an array selection, for example, so what would the current item even mean?
> 
The "current node" would mean the same thing as it means [in JMESPath](https://jmespath.org/specification.html#current-node), where it is clearly defined. At the start of JSONPath evaluation, it is the JSON document itself, what else could it be? As the expression is evaluated, the value that the current node represents changes to reflect the node currently being evaluated.  Given an "?(<expr>)" within [], the processing model requires an iteration over the elements of the then current item (say an array); the expression following the "?" is evaluated with each of those elements, and an "@" appearing in that expression is substituted for with those elements.

gregsdennis: > ... but @ changes to reflect the node being evaluated.

"... but `@` changes to reflect the _current scope_.

glyn: > The "current node" would mean the same thing as it means [in JMESPath](https://jmespath.org/specification.html#current-node), where it is clearly defined. At the start of JSONPath evaluation, it is the JSON document itself, what else could it be?

That seems to contradict the wording of this PR:
> The current item selector `@` when used outside a filter behaves as the document
root selector unless a child node is given to the evaluator.
When a child node is used as current item, evaluator must select it as starting
node instead of the document root when an expression starting with `@` is
evaluated.

bettio: I'll make an example for the benefit of this conversation, I rather discuss any disagreement comment about this feature on issue #59, so we just keep the conversation about pros/cons in just one place, and here we can just discuss the wording for what we decide at #59.

Let's assume the following eval function:
```
json_path_eval(document_root, path)
```

this implementation will always eval `@.foo` in the same way as `$.foo` as kindly mentioned by @danielaparker in previous comment.

any implementation might also augment the API by adding an additional (optional) parameter (an overload, or whatever the language supports):
```
json_path_eval(document_root, current_item, path)
```

therefore `@.foo` evals to `current_item.foo`, while `$.foo` evals to `document_root.foo` (an example is provided [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-797116919) ).

It is up to the API user to decide to use current_item instead of the document root (@gregsdennis explained it well [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-797212814) ).

Clearly we shouldn't describe the json_path_eval API in our draft (the API is clearly implementation specific).
We just describe the syntax (which is already supported by a number of implementations) and we formalize that a child node can be optionally provided, and in that situation "When a child node is used as current item, evaluator must select it as starting node instead of the document root when an expression starting with @ is evaluated.".

When a current item/an explicit starting scope is not provided, "The current item selector @ when used outside a filter behaves as the document root selector"

An implementation that doesn't care can safely just assume that `@` = `$` unless when current scope is changed by filters.

This wording should enable tooling as described in #59.

Let me know if you have any advice about how to improve wording so we make sure that the behavior described in #59 is accurately described, in case there are no further objections about the wording let's resolve these comments.

draft-ietf-jsonpath-base.md
bettio: @gregsdennis: do you think we need to introduce the idea of scope right now, or just wait the filters PR?

gregsdennis: I would assume that nested filter queries are out for right now.  There's some discussions [somewhere in the comments on another issue](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/17#issuecomment-698757622) regarding nested filters.  For now this just adds the syntax.  I think it'd better to bring in scope when we start defining filters.  Then we can start looking at the syntax behavior a bit more.

draft-ietf-jsonpath-base.md
danielaparker: > > The "current node" would mean the same thing as it means [in JMESPath](https://jmespath.org/specification.html#current-node), where it is clearly defined. At the start of JSONPath evaluation, it is the JSON document itself, what else could it be?
> 
> That seems to contradict the wording of this PR:
>

I agree :-) And I prefer the [JMESPath specification wording](https://jmespath.org/specification.html#current-node). In the JMESPath wording, there is no "if it's used outside the filter it means this" or "if it's used inside the filter it means that". There is just one consistent meaning for the "current-node", represented by `@`, it means the node currently being evaluated.

The original [JSONPath article](https://goessner.net/articles/JsonPath/) doesn't say much about the current node, it uses `@` to represent it (referred to as "current object"),  and gives two examples:

```
$.store.book[(@.length-1)].title                   (1)

$.store.book[?(@.price < 10)].title               (2)
```
Both of those examples are fully consistent with the JMESPath wording. In (1), the "current node" @ represents the array `$.store.book`, in (2), the `?` connotes iteration over the array, and `@` represents the array element.

It's easy to generalize that at `$`, the current node `@` represents '$', at `$.store`, it represents `$.store`, and so on.

danielaparker: > > ... but @ changes to reflect the node being evaluated.
> 
> "... but `@` changes to reflect the _current scope_.

See my reply to Glyn. I prefer the [JMESPath specification wording](https://jmespath.org/specification.html#current-node), which is very clear, and completely justifies paths that start with `@`.

draft-ietf-jsonpath-base.md
bettio: Added a reminder [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/17#issuecomment-803584276), I'm resolving this conversation.

draft-ietf-jsonpath-base.md
bettio: @danielaparker: does replacing

> The current item selector `@` when used outside a filter behaves as the document root selector unless a child node is given to the evaluator.

with

> The current item selector `@` can be used to represent the current node being evaluated. The initial current node defaults to the document root unless a child node is given to the evaluator as initial current node.

looks better to you?

Maybe another phrasing might be:


> The current item selector `@` can be used to represent the current node being evaluated. The initial current node defaults to the document root unless a different node is chosen.

bettio: I think I might also replace:

>When a child node is used as current item, evaluator must select it as starting
node instead of the document root when an expression starting with `@` is
evaluated.

with

> Therefore when a child node is set as initial current item, evaluator must select it as starting node instead of the document root when an expression starting with `@` is evaluated.

danielaparker: @bettio wrote:

> @danielaparker: does replacing
> 
> > The current item selector `@` when used outside a filter behaves as the document root selector unless a child node is given to the evaluator.
> 
> with
> 
> > The current item selector `@` can be used to represent the current node being evaluated. The initial current node defaults to the document root unless a child node is given to the evaluator as initial current node.
> 
> looks better to you?
> 
> Maybe another phrasing might be:
> 
> > The current item selector `@` can be used to represent the current node being evaluated. The initial current node defaults to the document root unless a different node is chosen.

I think it's enough to say "The current item selector `@` can be used to represent the current node being evaluated", full stop. That's about equivalent to the meaning of `@` in JMESPath, and to the meaning of "." in XPATH 3.1, see [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/59#issuecomment-803618374). 

For a JSONPath evaluator, it only knows about one document, the one provided as an argument, along with the JSONPath expression. That current node is initialized to that document, and evaluation begins.

Sure, you can have higher level functions built on top of the JSONPath evaluator, that pass a different document to the evaluator depending on whether the path starts with `$` or `@`. Implementers can support such functions if they wish. But the evaluator doesn't know about that, it only knows about one document.

bettio: @danielaparker:
> For a JSONPath evaluator, it only knows about one document, the one provided as an argument, along with the JSONPath expression. The current node is initialized to that document, and evaluation begins.

I would rather stay explicit, so we make sure that it cannot be misunderstood. Therefore writing "The initial current node defaults to the document root" doesn't hurt and it helps an unambiguous understanding.

@danielaparker:
> unless a child node is given to the evaluator as initial current node.

This helps remarking that `@` at the beginning of an expression is not necessarily a `$` synonymous according to the following proposal:   

@gregsdennis:
> Proposal: allow paths to start with @. It operates no differently than $, but can serve as an important indicator to tooling and other systems that consume JSON Path.

my goal is making sure that starting evaluation with current item pointing to a child node and optionally referencing it with `@` at the beginning of an expression is compliant with the specification.
I feel like that just stopping at  "The current item selector @ can be used to represent the current node being evaluated" might leave my goal in an unclear status.

By the way when filtering will be written we'll also remark that current item is updated.

@danielaparker:
> Sure, you can have higher level functions built on top of the JSONPath evaluator, that pass a different document to the evaluator depending on whether the path starts with $ or @. Implementers can support such functions if they wish. But the evaluator doesn't know about that, it only knows about one document.

For the record, my evaluator has 3 parameters: `json_path_eval(document_root, current_item, path)`. It is implemented as a recursive function and `document_root` is always passed unchanged, while `current_item` is updated. The caller may call it doing something like:
`json_path_eval(root, root.foo, "@.bar")` which is equivalent to `json_path_eval(root, root, "$.foo.bar")`.
No higher level function magic here ;)
Anyway let's keep out of the discussion implementation detail, again I just need to make sure that `@.bar` (with the meaning that I just showed in the example) is not left out of the spec.

Lastly the PR just aims to allow `@` at the beginning of the expression, I think more PRs are needed for a complete `@` formalization.  

@danielaparker thanks for your feedback, looking forward to further comments :)

bettio:
As a summary I'm aiming to address with this PR: 
- [ ] allowing `@` at the beginning of an expression
- [ ] `@` at the beginning of an expression defaults to document root
- [ ] `@` at the beginning of an expression can be != `$` on certain use-cases

I'm not addressing:
- [ ] `@` scoping
- [ ]  complete description / formalization of `@`

danielaparker:
@bettio wrote:

> As a summary I'm aiming to address with this PR:
> 
> * [x]  allowing `@` at the beginning of an expression
> * [x]  `@` at the beginning of an expression defaults to document root
> * [x]  `@` at the beginning of an expression can be != `$` on certain use-cases
> 
Fair enough, it can be left open what the current node is initialized to, I have no objection.

Daniel

draft-ietf-jsonpath-base.md
bettio: @gregsdennis: wish to read any feedback about the wording from you too :) Your help has been valuable so far

draft-ietf-jsonpath-base.md
```diff
@@ -449,13 +450,16 @@ return an error.
 
 ## Syntax
 
-Syntactically, a JSONPath consists of a root selector (`$`), which
-selects the root node of a JSON value, followed by a possibly empty
-sequence of *selectors*.
+Syntactically, a JSONPath consists of either a root selector (`$`), which
+selects the root node of a JSON value, or a current item selector (`@`), which
+selects a child node of a JSON value, followed by a possibly empty sequence of
+*selectors*.
```
gregsdennis, line 456: I think this could align with the syntax descriptor below better.

> Syntactically, a JSON Path consists of a start selector, which may either be the root selector `$` or the current item selector `@`, followed by...

draft-ietf-jsonpath-base.md
gregsdennis, line 531: Usage of "node" here?  I think this is one we've decided to avoid.

draft-ietf-jsonpath-base.md
```diff
@@ -517,6 +521,18 @@ This is the concatenation of three lists, two of length one containing `0`, `1`,
 As a consequence of this approach, if any of the selectors selects no nodes,
 then the whole JSONPath selects no nodes.
 
+The current item selector `@` when used outside a filter behaves as the document
+root selector unless a child node is given to the evaluator.
+When a child node is used as current item, evaluator must select it as starting
+node instead of the document root when an expression starting with `@` is
+evaluated.
+
+When a child node is used as current item, the current item selector `@` selects
```
gregsdennis, line 530: I'm not sure who's working with English as a first language, so please don't take offense, but this could use some grammatical massaging, specifically the use of the word "the" in a few places.

draft-ietf-jsonpath-base.md
```diff
@@ -517,6 +521,18 @@ This is the concatenation of three lists, two of length one containing `0`, `1`,
...
+it as starting node instead of the document root.
+
+The root selector `$` should be prefered to the current item selector `@` when
+a path relative to the current item is not required.
```
gregsdennis, line 534: ... or when no current item can be defined.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
cabo: I don't know who "we" is; I think we are still discussing terminology.

cabo: (I'm not so sure about "child node", the term "child" is usually used for direct descendants.)

gregsdennis: Yes, we're still discussing it, which means that it's something to avoid for now.

bettio: ok, I can replace `node` with `item` and `child node` with `subdocument`. I think that further improvements might be applied on future PRs.

cabo: Please wait for PR#72 to settle.

cabo: "document" certainly is not the terminology we want to use.

bettio: I think that I can replace "`The current item selector`" ... "`with @ is evaluated.`"

with

```
The current item selector `@` can be used to represent the current item being
evaluated. The initial current item defaults to the document root unless a
subdocument is given to the evaluator as initial current item.
```

I think I can safely remove "`When a child node is used as current item, evaluator must select it as starting
node instead of the document root when an expression starting with @ is
evaluated.`" and  "`When a child node is used as current item, the current item selector @ selects
it as starting node instead of the document root.`" since the proposed contains the same information but with a better phrasing.

@gregsdennis Let me know if it sounds good to you so I can push changes I described here.

draft-ietf-jsonpath-base.md
bettio: >  but this could use some grammatical massaging

It is not my first language, so any advice or feedback is welcome

Anyway I think that I'll remove this and just keep:
```
The current item selector `@` can be used to represent the current item being
evaluated. The initial current item defaults to the document root unless a
subdocument is given to the evaluator as initial current item.
```

as described in the other conversation, if you agree too that "`When a child node is used as current item, the`"... is just redundant we can remove it and close this conversation since we don't need rephrasing it.

draft-ietf-jsonpath-base.md
gregsdennis: I think @cabo makes a good point about resolving it in another issue when we finish discussing "node."

No worries.

draft-ietf-jsonpath-base.md
bettio: @cabo: just pushed further changes to this, I think that this conversation is outdated and it can be archived. Let me know if you have any further feedback.

cabo: Sorry, github does not make it easy to find the current state of a PR where a conversation has led to changes.  So please ignore my comments if they relate to outdated contents.

I think the specific wordsmithing of this PR can wait until some other parts of the draft get more stable.

draft-ietf-jsonpath-base.md
bettio: @gregsdennis: Ok, I'm keeping node for now, we can improve this with a future PR.

@cabo I removed document and replaced it again with "root node". I also replaced "child node" with "descendant node" which is more accurate.

So I will commit & push:
```
The current item selector `@` can be used to represent the current item being
evaluated. The initial current item defaults to the root node unless a
descendant node is given to the evaluator as initial current item.
```

Let me know if you have any further feedback, otherwise let me know you are ok for now so I can push it.

glyn:
@bettio What are your plans for this PR? Thanks.

glyn:
Closing as there was no response. Please re-open if you would like to continue working on this PR.

Issue #60, opened 2021-03-10
Determinism?


cabo:
Does applying a JSONPath query to a JSON data item always generate the same result?

Some selectors do not have a natural order of the nodelists they return, e.g., $.* applied to a JSON object

We could prescribe an order (e.g., sorting keys lexicographically or in JavaScript order) if determinism is important

We could disallow .* on JSON objects

goessner:
hmm .. depends, how deterministic JSON parsers are when iterating over 
objects ... they should work in ascending chronological order ...

https://stackoverflow.com/questions/42491226/is-json-stringify-deterministic-in-v8

JavaScript 'Object.keys' also guaranties the same order

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

So ... order should be always the same, as long the implementation is 
the same. How different implementations behave is uncertain at best.

But why should we care and heal deficiencies of JSON parsers in that case?

I can't see a benefit of always guaranteeing the same -whatever- order. 
And disallowing iterating over objects by '$.*' might break existing 
implementations.



Am 10.03.2021 um 14:02 schrieb cabo:

cabo:
(Determinism in interop testing is a major benefit for me.  But that can be added by reprocessing after the JSONpath processing, so it is not completely a killer argument.)

glyn:
Another selector which requires careful definition for ordering is `..`, particularly when this recurses over an object or objects.

goessner:
hmm ... compliance testing is a point, sure. Simple first tests of `$.obj.*` and `$..*` regarding ordering should be possible with current test suite ... ?

glyn:
> hmm ... compliance testing is a point, sure. Simple first tests of `$.obj.*` and `$..*` regarding ordering should be possible with current test suite ... ?

Anything is "possible". ;-)

Tests involving objects which have only one name/value pair are trivially easy to test because any non-determinism of object enumeration is irrelevant. But tests involving objects with two or more name/value pairs need to capture the ordering possibilities somehow. Note that the [current test suite](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/main/cts.json) is a JSON document so there would need to be a _declarative_ way of specifying the ordering possibilities.

gregsdennis:
We covered this topic over in the JSON Schema repo somewhere with quite a lengthy discussion.  I'll see if I can dig it up.

[...]

Issue #14, opened 2020-09-06
Define an algorithm


gregsdennis:
The processes between implementations may vary widely.  This is fine so long as they all have the same output for the a given input.  However it may be beneficial to outline a basic algorithm.

- guidance for new implementations
- reference point internally to the spec for describing how operators (selectors) work.

## First thoughts

Define a selector to be a function that takes an array of values and outputs an array of values:

```
selector(value[]) : value[]
```

Next, define a path as a sequence of selectors.

```
path : selector[]
```

To evaluate a path on a JSON instance, we just need to place the instance in an array and run through all of the selectors so that the output of each selector is the input of the next.  The final output is the output of the final selector.

```
current = [input]
for each selector in path
    current = selector(current)

return current
```

---

Looking at it, there's not really much to it, but it's nice to have something defined.  Now we only need to define the selectors (I'm authoring another issue).

glyn:
It's also possible to frame this in terms of a selector that takes a single value as input and produces an array of values as output. Then the above code needs a "flat map" adding to it, which might be a bit opaque in pseudocode.

gregsdennis:
Right, but that's merely iterating the function taking a single value and iterating over the values output by the previous selector.  I think it's simpler when the selector takes in the full output from the previous one.  Regardless, the important bit here is that we should define something for that the spec can reference.

(JSON Schema does this when talking about annotations.)

glyn:
> I think it's simpler when the selector takes in the full output from the previous one.

The code calling the selector is simpler because it no longer needs a "flat map", but then every selector has a more complex signature, taking an array of values as input instead of a single value. Also, I think each selector really needs to treat each value in the array independently from the rest, and that's obvious if a single value is passed to the selector instead of an array of values.

gregsdennis:
That's fine, but we need to be sure that the results of each selector are collected separately so that there is a clear definition of when to move to the next selector.

Passing full array (repeated from above):
```
current = [input]
for each selector in path
    current = selector(current)

return current
```

Passing individual items:
```
current = [input]
for each selector in path
    results = []
    for each item in current
        results.add(selector(item))
    current = results.distinct()

return current
```

It can be very easy, but wrong, to get to something like this while trying to optimize:
```
current = [input]
for each selector in path
    for each item in current
        current.add(selector(item))

return current
```

[...]

PR #79, opened 2021-03-23
Processing model

cabo:
Try to nail down the processing model some more, and raise four discussion points that we might be able to resolve now.


gregsdennis:
Seems to be related to #14.

cabo:
Indeed, this tries to reflect the great discussion in #14.

draft-ietf-jsonpath-base.md
```diff
@@ -159,6 +159,20 @@ in ABNF as `%x2318`.
 
 The terminology of {{-json}} applies.
 
+For the purposes of this specification, a JSON value as defined by
+{{-json}} is also viewed as a tree of nodes.
```
gregsdennis, line 163: This is good.  We don't need to claim that JSON itself is a tree of nodes, only that this spec views them as such when convenient.

Maybe instead of "is also viewed..." we use "can also be viewed..."

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
-If a syntactially invalid JSON value is provided, any implementation SHOULD
-return an error.
+A string to be used as a JSONPath query needs to be *well-formed* and
+*valid*.
+A string is a well-formed JSONPath query if it conforms to the syntax
+of JSONPath.
+A well-formed JSONPath query is valid if it also fulfills all semantic
+requirements posed by this document.
+
+Well-formedness and validity of JSONPath queries are independent of
```
gregsdennis, line 443:
```suggestion
The well-formedness and validity of JSONPath queries are independent of
```

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+the JSON value the query is applied to; no further errors can be
+raised during application of the query to a JSON value.
+
+(Obviously, an implementation can still fail when executing a JSONPath
+query, e.g., because of resource depletion, but this is not modeled in
+the present specification.)
+
+> Discussion: are we ready for the error model defined here, i.e. one
+> that does not have runtime errors, but only compile time errors?
```
gregsdennis, line 452: I think this needs further discussion before representing it in a PR.

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+
+A valid JSONPath query is executed against a JSON value, the *Argument*, and
+produces a list of zero or more nodes of the JSON value which are
+selected by the JSONPath.
+(Note that the term JSON value also implies that this value complies
+to the definitions in {{-json}}, i.e., is indeed a JSON value.)
+
+The JSONPath query is a sequence of zero or more *selectors*, each of
+which is applied to the result of the previous selector and leaves
+input to the next selector.
```
gregsdennis, line 464:
```suggestion
which is applied to each node in the result of the previous selector and outputs
a single list of nodes to be used as input by the next selector.
```

See [discussion](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/14#issuecomment-694687029) over representing the algorithm as iterating over the individual results of the previous selector vs taking the full result set as a single input.

The suggestion above is in line with the rest of the content of the PR so far: each selector iterates over the output from the previous selector.

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+These results and inputs take the form of a *nodelist*, i.e., a
+sequence of zero or more nodes.
+
+The nodelist going into the first selector contains a single node,
+holding the Argument.
+The nodelist resulting from the last selector is presented as the
+result of the query; depending on the specific API, it might be
+presented as an array of the JSON values at the nodes, an array of
+Output Paths referencing the nodes, or both — or some other
+representation as desired by the implementation.
```
gregsdennis, line 474: I like leaving this open to the implementation.  We _could_ (in a later PR) go so far as to define what the output looks like for each of these cases in order to give an expected API for consumers of the implementations.

It may also be beneficial to explicitly state that a node list is just a collection of nodes; it is _not_ a JSON array, although its _representation in JSON_ is.  This is to say that a node list can be _serialized_ to a JSON array (e.g. for the purposes of output), but it is not one inherently.

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+raised during application of the query to a JSON value.
+
+(Obviously, an implementation can still fail when executing a JSONPath
+query, e.g., because of resource depletion, but this is not modeled in
+the present specification.)
+
+> Discussion: are we ready for the error model defined here, i.e. one
+> that does not have runtime errors, but only compile time errors?
+
+## Processing Model
```
gregsdennis, line 454: We should add a paragraph that state that this processing model is an example for the purposes of explaining how JSON Path works and is not prescriptive of the internal workings of an implementation.  If an implementation wishes to (or needs to due to framework limitations) design a different process that yields the same results, this should be acceptable.

draft-ietf-jsonpath-base.md
```diff
@@ -480,32 +519,29 @@ operations which might be considered erroneous, such as indexing beyond the
...
-Suppose the input
-document is: `{"a":[{"b":0},{"b":1},{"c":2}]}`.
-As we will see later, the JSONPath `$.a[*].b` selects the following list of nodes: `0`, `1`.
+But a selector doesn't just act on a single node: a selector acts on
+each of the nodes in its input nodelist and selects an output
+nodelist for each of them, as follows, which are then combined using
+the combine1 algorithm to form the result nodelist of the selector.
+
+> Discussion: Do we allow selectors that have their own combine1
+> variant, or can we always use the same?
```
gregsdennis, line 528: Definitely worth discussion, but I think a common "combine" mechanism is simplest to understand.  Each selector defining its own forces the Path author to remember how each one behaves, which can get confusing fast.

draft-ietf-jsonpath-base.md
```diff
@@ -480,32 +519,29 @@ operations which might be considered erroneous, such as indexing beyond the
...
+
+For each node in the list, the selector selects zero or more nodes,
+each of which is a descendant of the node or the node itself.
+
+> Discussion: Is that a common requirement?  Or can a selector also go
+> up, or to the query Argument?
```
gregsdennis, line 534: This touches on the function of `@`, `$` within filter queries, and the idea of scope when `@` is used in nested filter queries.

I think it's better left for another PR when we've explored these ideas and how they relate.

gregsdennis (changes requested):
I think this is coherent, and I like the direction.  I've also added a few suggestions/comments.

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +433,70 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+presented as an array of the JSON values at the nodes, an array of
+Output Paths referencing the nodes, or both — or some other
+representation as desired by the implementation.
+Note that the API must be capable of presenting an empty nodelist as
+the result of the query.
+
+A selector performs its function on each of the nodes in its input
+nodelist, during such a function execution, such a node is referred to
+as "current node".  Each of these function executions produces a
+result nodelist, which are then combined by algorithm *combine1* into
```
glyn, line 481: combine1 appears to be undefined.

draft-ietf-jsonpath-base.md
```diff
@@ -201,6 +215,12 @@ Query:
 Argument:
 : Short name for the JSON data item a JSONPath expression is applied to.
 
+Nodelist:
+: Output of applying a query to an argument: a list of nodes within
```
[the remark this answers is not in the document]
cabo: There is a definition for "node" up in line 183.

I'm not entirely sure we need both "item" and "node", but I would ask for your patience to leave this for now until we figure it out.

draft-ietf-jsonpath-base.md
cabo: Well, we very actively view it as a tree of nodes in the current text, so I think we are good as is.

draft-ietf-jsonpath-base.md
cabo: Indeed.  So we could resolve this question before merging, or we could merge with the question intact, to be resolved in later PR.

draft-ietf-jsonpath-base.md
cabo: Right.  The current PR text says "depending on the specific API", so I think we are covered on the first item.  It also doesn't seem to suggest the nodelist is a JSON array, or are text changes actually required for the second suggestion?

draft-ietf-jsonpath-base.md
cabo: Good point.  I put a new paragraph at the start of 3.2.  I had to weasel word a bit around "the same", because we are saying the model isn't deterministic.

draft-ietf-jsonpath-base.md
cabo: That is a good reason to stick with a common "combine1" model.

draft-ietf-jsonpath-base.md
cabo: Sure -- it has the strange name because it is a placeholder that needs to be filled in based on the discussion points.  But we can live with this placeholder while we resolve the discussions, hence giving it a name is a good idea for these discussions.

draft-ietf-jsonpath-base.md
cabo: The changed text implies that each function invocation of the selected leaves a node list that is then independently fed to the next selector.  I think the discussion resolved to doing the combine1 step first, yielding a single nodelist, and then applying the next selector again to each node in that nodelist.

draft-ietf-jsonpath-base.md
gregsdennis: I don't think that we should rely on implications.  Explicit statements are always better.

cabo: Can you say what the wording should be?

draft-ietf-jsonpath-base.md
gregsdennis: I'm not sure I followed that.  My suggested changes align this statement with the rest of the your descriptions to say that each selector takes the nodelist (combined) output by the previous selector as a single argument and iterates over the nodes internally.

The other form is to say that the selector is fed each node individually by the processor, which doesn't line up with the rest of the PR.

draft-ietf-jsonpath-base.md
gregsdennis: Are you saying that we should leave this markdown block quote in place to spur discussion?

danielaparker (commented):
Regarding "Data Item: A structure complying to the generic data model of JSON, i.e., composed of containers, namely JSON objects and arrays, and of atomic data, namely null, true, false, numbers, and text strings. Also called a JSON value.", this seems to be related to @ttimbrays posts [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/66#issuecomment-804425507) and [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/66#issuecomment-804425507). Perhaps "Data Item" could be dropped and just use "JSON value", as @ttimbrays proposed?

Regarding "Argument: Short name for the JSON data item a JSONPath expression is applied to", this also seems to be related  to @ttimbrays post [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/66#issuecomment-804425507). Perhaps "Argument" could be replaced with "root value", as @ttimbray suggested?

Note that if a node is regarded as a value/location pair, than a JSON value cannot be considered as a "tree of nodes", or a "node" itself. We can define a "children" property on a JSON value, which can be considered its members or elements if such exist. But we cannot define a "location" property for a JSON value, it's not possible, and "location" seems to be an essential property of a node as understood in the processing model defined in this PR. See [here](https://www.w3.org/TR/xpath-datamodel-31/#Node) for a related definition of "node". JSONPath doesn't need a notion of "node" as complicated as XDM, but minimally it needs location, children, and value properties.

I don't think this observation presents a great difficulty for the PR, but I think that mentions of JSON values as "trees of nodes" and remarks that "the term _node_ has the same meaning [as JSON value]" would be better expunged.

draft-ietf-jsonpath-base.md
cabo: Yes.  Until that discussion is resolved.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: I think "node" warrants its own entry if we're to define "nodelist."  What is a "nodelist" a list of?

draft-ietf-jsonpath-base.md
cabo: The processing model actually says that the selector processing the input nodelist into the output nodelist by processing each entry of the input nodelist into a partial output nodelist and then uses combine1 to put these together into the output nodelist.  This probably needs better terminology.  Your text tells me that a selector is applied to each element of the input nodelist and does not say what we do with all those partial output nodelists.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: OK, I can make line 183 into its own entry.

draft-ietf-jsonpath-base.md
gregsdennis: Actually, it looks like lines 221-222 cover it: 

> While this list can be represented in JSON, e.g. as an array, the nodelist is an abstract concept unrelated to JSON data items.

cabo:
It seems these are all terminology comments.
Can we refrain from continuing the Brownian motion on terminology and focus on content for this PR?
Unless of course the terminology is wrong or not sufficiently sharp.
We can go back to terminology discussions after having made some progress.

draft-ietf-jsonpath-base.md
gregsdennis: If that's the convention we want to follow for authoring the spec, this is fine.

draft-ietf-jsonpath-base.md
goessner: I like that "view" trick ...

draft-ietf-jsonpath-base.md
gregsdennis: After rereading my comments, I see the confusion between what I've been saying and my addition of "each node in."  I think this can be removed from my suggestion:

```suggestion
which is applied to the result of the previous selector and outputs
a new nodelist to be used as input by the next selector.
```

Regarding how the results of each node are combined, this paragraph doesn't address it at all.  This paragraph instead focuses on the input/output stream of the selectors.  The "combining" logic is something internal to how the selectors work, which is discussed later when you're talking about the "combine1" function.  I think this is a good layout for the spec as it stands in this proposal.

cabo:
I have collected these items in #84 so we don't forget where we were.

draft-ietf-jsonpath-base.md
cabo: (Marking this as resolved -- it is however best if this is done by who raised the comment.)

danielaparker:
> It seems these are all terminology comments.
> Can we refrain from continuing the Brownian motion on terminology and focus on content for this PR?
> Unless of course the terminology is wrong or not sufficiently sharp.
> We can go back to terminology discussions after having made some progress.

@cabo, can you at least define the properties of "node" for JSONPath? I think the discussion of the processing model depends on that, just as the discussion of the processing model in XPath depends on the properties of "node" defined in the XDM. I think the essential properties are value and location, but I would like to know your view.

draft-ietf-jsonpath-base.md
goessner: Each selector can only select nodes from the subtree of the current node. So with each selector the number of potentially selectable nodes is reduced or stays constant at best. It can never increase ... so "combine" suggests "increase", is misleading and should either be "pick model" or "select model" or something similar.

cabo: Well, a selector operates on the nodes in the input node list.
The "combine1" function is about combining the outputs that each such operation provides.

So if you have `[1], [2]` as the nodelist (identifying nodes by their values here), and the selector is `.[0]`, the output nodelists from each selector function invocation are `1` and `2` respectively.  combine1 would create an output nodelist `1, 2`.  This is a bit boring, but more complex selectors can leave a nodelist for each node in the input nodelist, so we need to discuss how to combine them:

`{a: {a: 1}},  {a: {a: 2}}` as input nodelist, `..a` as selector: The individual nodelists are  `{a: 1}, 1` and `{a: 2}, 2`.  combine1 turns this into, say, `{a: 1}, 1, {a: 2}, 2`.

cabo:
> @cabo, can you at least define the properties of "node" for JSONPath? I think the discussion of the processing model depends on that, just as the discussion of the processing model in XPath depends on the properties of "node" defined in the XDM. I think the essential properties for JSONPath are value and location, but I would like to know your view.

I was thinking that line 183..185 were doing that.  But since you insist, in eefc298 I have tried to clarify this paragraph and elevate it to its own definition.

draft-ietf-jsonpath-base.md
goessner: A selector can never go up, given the information the current node is holding alone, as it lacks parent node info.

Maintaining parent node info during selection process introduces a new &ndash; hard to overview &ndash; level of complexity and should be avoided.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
goessner: Do we also need ...

```
Value:  A JSON value as defined in [RFC8259].
```

It would make your inline ...

```
(Note that the term JSON
   value also implies that this value complies to the definitions in
   [RFC8259], i.e., is indeed a JSON value.)
```
... obsolete.

danielaparker:
@cabo wrote:

> I was thinking that line 183..185 were doing that. But since you insist, in [eefc298](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/commit/eefc298f5e2711d7acb858f5798c1b59f2de3ef9) I have tried to clarify this paragraph and elevate it to its own definition.

Thank you. Your forbearance is appreciated :-)

I was hoping for something more along the lines of XQuery and XPath Data Model 3.1 wording:

"All nodes must satisfy the following general constraints:"

"Elements [Nodes] have the following properties:"

But I understand (and largely agree with) this sentence

A node can be viewed as a combination of a (1) JSON value and (2) its location in the  argument; the latter can, if desired, be represented as an Output Path.

I (and no doubt @timbray) would prefer "root value" to "argument", and I would prefer "Normalized Path" to "Output Path" (why use any other then the easily constructed canonical one?) but these then become terminological or minor issues.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: We already say:    The terminology of [RFC8259] applies.
So we have already imported all that terminology.

On the specific note:
A note is by definition redundant; it just alerts the reader to what is said in the note.  I think it is very much worth to say this again, because it reminds people that we are not taking a position on error handling with respect to broken JSON input -- this is just out of scope for the present specification.

draft-ietf-jsonpath-base.md
danielaparker: @goessner wrote:

> Oh well ... it's a simple "append" to the resulting nodelist. According to my mental model allowing only a single resulting nodelist, the sequence of ...
> 
> * `{a: 1}`
> * `1`
> * `{a: 2}`
> * `2`
> 
Just this. That corresponds to my "mental model", and directly in many implementations, some implementations.

> ... would be added, where order is implementation dependent of course (determinism).

Implementation dependent, yes, but I think the "natural order", would satisfy the minimally restrictive conditions identified by @glyn in #60.

glyn: Yes, to spell this out, there is _one_ object with more than one key/value pairs and thus one source of non-determinism in this example (assuming the visitation ordering of `..` will be tied down in the spec). So the other possible resultant nodelist would  correspond to:
* `{a: 2}`
* `2`
* `{a: 1}`
* `1`

draft-ietf-jsonpath-base.md
gregsdennis: I looked for it last night, but it seems I can't mark it as resolved.

goessner (approved):

glyn:
@cabo This needs rebasing on main. Also the text includes some discussion points and I'm not sure whether your intention is that these be addressed before merging?

draft-ietf-jsonpath-base.md
```diff
@@ -413,51 +436,77 @@ the grammar in this document, its UTF-8 form SHOULD first be decoded into
...
+A selector performs its function on each of the nodes in its input
+nodelist, during such a function execution, such a node is referred to
+as "current node".  Each of these function executions produces a
+result nodelist, which are then combined by algorithm *combine1* into
+the result of the selector.
+
+> Discussion: We haven't decided yet whether there is only one
+> *combine1* or a choice, with for instance simple concatenation and
+> concatenation with duplicate removal, based on a definition of
+> duplicate, are candidates.
```
glyn, line 497: I think a single *combine1* is sufficient and much simpler than allowing selector-specific behaviour.

draft-ietf-jsonpath-base.md
glyn: I agree a selector can never go up.

glyn (approved):
Approving with discussion points left inline. Better to make progress on the rest.

glyn:
@gregsdennis and @danielaparker I'd like to get this PR merged with the discussion points inline. That would be a nice stake in the ground. Do either of you have objections to that or are you comfortable with merging?

draft-ietf-jsonpath-base.md
cabo: The word nodelist is first discussed in the next sentence, so this is simply pulling in an unneeded forward reference.

draft-ietf-jsonpath-base.md
cabo: This thread proposes an answer to the discussion point.
Maybe we can merge the PR first and then specifically address the discussion point.

cabo:
To make this easier, I rebased on main again, and I numbered the discussion points (D1 to D4).

draft-ietf-jsonpath-base.md
gregsdennis: Still, "leaves input to..." is odd wording.

gregsdennis (approved):
I'm happy with this for its purpose.

I'd like to deal with [this wording] (https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/79#discussion_r602668320)  at some point, but it's not crucial to this PR.

draft-ietf-jsonpath-base.md
cabo: 89d7d72: leaves ➔ provides

danielaparker:
It might be helpful to consider the parse tree for selectors, which, because of the simplicity of the JSONPath grammar, has a simple representation as a list, with each element in the list containing some specific information for its type, and a "next" selector. Each selector can be thought to have a select operation that takes as input a current node, invokes select on its "next" selector one or more times if "next" is not nil, or appends to a result nodelist if at the tail.

For example, given the the JSONPath `$.a[*].b`, the "parse tree" would correspond to:
```
root selector
    identifer selector "a"
       wildcard selector
          identifier selector "b"
              nil
```
Given a root value `{"a":[{"b":0},{"b":1},{"c":2}]}`, evaluation would proceed as
```
root selector selects root value {"a":[{"b":0},{"b":1},{"c":2}]} and invokes its "next" selector once
    identifier selector "a" selects [{"b":0},{"b":1},{"c":2}]  and invokes its "next" selector once
        wildcard selector invokes its "next" selector for each of {"b":0}, {"b":1}, and {"c":2} (3 invocations)
            {"b":0} -> 0     ; identifer selector "b" at tail adds {"$['a'][0]['b']", 0} to nodelist
            {"b":1} -> 1     ; identifer selector "b" at tail adds {"$['a'][0]['b']", 1} to nodelist   
            {"c":2} -> nil    ; identifier selector "b" does nothing
```
The result is a nodelist
```
{"$['a'][0]['b']", 0}
{"$['a'][0]['b']", 1} 
```

Here there is no combining of lists, there is only concatenation into one nodelist. The result nodelist can be stripped of duplicates or sorted according to path components if desired.

gregsdennis:
@danielaparker the goal for the spec is to have some algorithm to reference internally.  It doesn't need to consider edge cases or have much detail.  It just needs to be sufficient for other parts of the spec to use.  It's not even intended as guidance.  We want to leave as much as possible up to the implementor.

PR #83, opened 2021-03-24
Avoid copying HTML to docs directory

glyn:
Also leave ABNF file around after check.


cabo (approved):
Looks good!

cabo (commented):
It is probably innocuous to leave this around in the main branch (as opposed to moving it over to gh-pages), so I think we can leave it at that.

PR #89, opened 2021-03-27
Fix formatting of Table 3: Escape Sequence Replacements

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -717,16 +717,16 @@ in the table below:
...
-| "\" "\"           | U+005C              |
-| "\" "b"           | U+0008              |
-| "\" "f"           | U+000C              |
-| "\" "n"           | U+000A              |
-| "\" "r"           | U+000D              |
-| "\" "t"           | U+0009              |
-| "\" uXXXX         | U+XXXX              |
+| "\\" %x22          | U+0022              |
+| "\\" "'"           | U+0027              |
+| "\\" "/"           | U+002F              |
```
cabo, line 722: Almost...

This needs to be "\\" "\\"

(Actually, I'd use `\` in both cases, but this works, too.)

glyn: @cabo Good eyes. I'll fix that later.

glyn: @cabo Now I come to look at this again, I can't see what's wrong - an escaped solidus (`/`) needs to be replaced with a solidus U+002F. What problem did you see?

cabo: The next line has an unescaped `\`.

glyn: @cabo Ah yes! Thanks. BTW PR comments normally refer to the line(s) above.

Issue #84, opened 2021-03-24
Terminology: next round


cabo:
Clusters of terms:

* Cluster 1

  * JSON value
  * (data) item
  * node
  * location

  These are related, but different.  We probably do not need all four.
  (Of course, JSON value is a 8259 term, so it does not go away.)
  We should decide which of these terms we want to use, and what they exactly mean.
  A node seems to be a combination of a discrete JSON value and its location.
  JSON value seems to imply equivalence by value (even though equivalence is undefined in RFC 8259),
  while node is based on equivalence by identity.
  Location seems to imply that you can get to the node (and thus the value) from it.

* Cluster 2

  * Argument: input to executing JSON query
  * root value, root node (used as Argument, may be used in nested queries)
  * current node (current item):
    * input to selector function invocation, 
    * maybe used inside filtering/indexing selectors

* Cluster 3

  * selector (part of the selector chain that makes up a JSONPath query)
  * (we probably need a word for) selector function \[invocation]

    This is needed if we stick to the model that each node in the
    input node list is treated separately and the results are then
    combined using (TBD) combine1.

  * combine1 (placeholder for what combines the outputs of the selector function invocations)

    This becomes important when we either resolve there is only one
    combine1 function or we resolve that combine1 functions are
    (within limits) implemenation specific (i.e., we give up) or that
    different selectors have different combine1 functions.

  All these can only be meaningfully decided once we have agreement on a processing model.

goessner:
... (we probably need a word for) selector function [invocation] ...

perhaps ...
* selection
* selection process 
* selection step (atomic)

danielaparker:
> Clusters of terms:
> 
> * Cluster 1
>   
>   * JSON value
>   * (data) item
>   * node
>   * location
>   
>   These are related, but different.  We probably do not need all four.
>   (Of course, JSON value is a 8259 term, so it does not go away.)

My view is that "JSON value" is enough, that "data item" is unnecessary  

>   We should decide which of these terms we want to use, and what they exactly mean.
>   A node seems to be a combination of a discrete JSON value and its location.

Agreed

>   JSON value seems to imply equivalence by value (even though equivalence is undefined in RFC 8259),
>   while node is based on equivalence by identity.
>   Location seems to imply that you can get to the node (and thus the value) from it.

Agreed

> * Cluster 2
>   
>   * Argument: input to executing JSON query
>   * root value, root node (used as Argument, may be used in nested queries)

[...]

PR #90, opened 2021-03-27
Editorial: Remove term "data item"

cabo:
See discussion in #84.

To do: What exactly doesn't $..* stand for -- the table currently says "member values" (i.e., no array elements).  Is that true?

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -418,10 +418,10 @@ constant.
 | `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
 | `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
 | `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheaper than 10                             |
-| `//*`                  | `$..*`                                    | all elements in XML document; all member values contained in JSON value |
+| `//*`                  | `$..*`                                    | all elements in XML document; all member values and array elements contained in JSON value |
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -419,8 +418,10 @@ constant.
 | `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
 | `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
 | `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheaper than 10                             |
-| `//*`                  | `$..*`                                    | all elements in XML document; all members of JSON data item  |
-{: #tbl-example title="Example JSONPath expressions applied to the example JSON data item"}
+| `//*`                  | `$..*`                                    | all elements in XML document; all member values contained in JSON value |
+{: #tbl-example title="Example JSONPath expressions applied to the example JSON value"}
+
+<!-- XXX: is $..* really just member values -->
```
glyn, line 424: There is a [consensus](https://cburgmer.github.io/json-path-comparison/results/dot_notation_with_wildcard_after_recursive_descent.html) around `..*`: it traverses descendants (including the starting node) and for each one applies `.*` to objects and `[*]` to arrays.

Let's defer fine-tuning the description until a PR which covers `..` in detail.

glyn (approved):
LGTM, thanks.

draft-ietf-jsonpath-base.md
cabo: Right.  To reduce confusion, I did add "array elements" to the text in the table; but clearly, this needs to be addressed in a technical PR.

goessner (approved):

PR #94, opened 2021-05-11
removed dashes `-` as valid characters for dot-name selectors

### LINKED ISSUE #16: Non-property-name-friendly chars
We should probably define what characters are valid for property names.  JSON makes no distinction, so we should support everything.  But this becomes especially important for the dot-property syntax.

The index-property syntax `["name"]` can support any string, so that's not a problem.  (I also think it's a good idea to mirror JSON's escaping requirements for strings.)

But for the dot-property syntax `.name`, I think it's probably a good idea to limit this to alphanumeric characters and underscores: `^[_a-zA-Z0-9]$`.  The following should also be considered:

- Non-English characters
- Starting with a number (many languages don't support starting with a number)
- Emoji and other UTF16 codepoints

gregsdennis:
Resolves #16

Also, I'm not sure what needs to be included in the PR as far as generated files go.  Any guidance would be helpful.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -592,7 +592,6 @@ selector = dot-child              ; see below for alternatives
 dot-child = "." dot-child-name / ; .<dot-child-name>
             "." "*"             ; .*
 dot-child-name = 1*(
-                   "-" /         ; -
                    DIGIT /
                    ALPHA /
                    "_" /         ; _


goessner (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -592,7 +592,6 @@ selector = dot-child              ; see below for alternatives
 dot-child = "." dot-child-name / ; .<dot-child-name>
             "." "*"             ; .*
 dot-child-name = 1*(
-                   "-" /         ; -
```
cabo, line 595: In a separate PR, we should remove all those redundant comments.

cabo (approved):

glyn (approved):

PR #97, opened 2021-05-12
84 terminology

### LINKED ISSUE #84: Terminology: next round

gregsdennis:
Resolves #84

## Alterations

- Removed `Object`, and `Array` since they are defined in 8259.  Replaced with extended verbiage at the beginning of the section that declares that 8259 applies.
- Replaced "JSON value" with "value" in line with the above.
- Relabeled "output path" as "normalized path"

## Additions

- String - a clarification and expansion on the JSON definition
- Root Node

## Other

- Reordered terms and paragraphs so that reading through doesn't use terms that haven't been defined yet (as much as possible).
- Removed a redundant parenthetical sentence in the *Processing Model* section.
- Removed some un-necessary self-references (e.g. "JSONPath query" rather than just "query").
- Made some general edits

### Existing in-place definitions

"Selector" and "slice" are defined in-place where they're needed.  Do we need to define them in the terminology section as well?

### Key

The draft currently uses "name" instead of "key" as we had agreed in the meeting.  I suspect that this is because 8259 uses "name."  Since we're trying to align with that, I've left in "name."

### Index

We discussed using "index" as a generic term for either a numeric index for arrays or a name for objects.  However in editing this section, I'm starting to second-guess that decision.

Given that we've opted to define a union selector as taking as arguments "indices," "slices," and/or "filters," I don't think it's too much of a stretch to separately include "indices" and "names."  Because of this, I left the definition of "index" as is.

This also addresses the tendency to associate "index" with arrays, as @glyn had mentioned.

I'm leaving the changes to "union selector" to include all of the various things it can accept for another PR as this is solely to clean up terminology.

### Emphasizing terms

I tried capitalizing defined terms to emphasize their special meaning.  It got weird to read, though, since they're used so frequently.  Maybe not _every_ instance of the terms needs to be emphasized.  I'm not sure.  Happy to discuss.

I've left this change out for now, but there are definitely inconsistencies present already.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -213,7 +213,7 @@ For the purposes of this specification, a value as defined by
 {{-json}} is also viewed as a tree of nodes.
 Each node, in turn, holds a value.
 Furthernodes within each value are the elements of Arrays and the
-member values of objects and are themselves Values.
+member values of objects and are themselves values.
 (The type of the value held by a node is
 may also be referred to as the type of the node.)
 
@@ -257,7 +257,7 @@ x['store']['book'][0]['title']
 ~~~~
 
 in popular programming languages such as JavaScript, Python and PHP,
-with a variable x holding the Argument.  Here we observe that
+with a variable x holding the argument.  Here we observe that
 such languages already have a fundamentally XPath-like feature built
 in.
 
@@ -270,9 +270,9 @@ The JSONPath tool in question should:
 
 ## Overview of JSONPath Expressions {#overview}
 
-JSONPath expressions always apply to a Value in the same way
+JSONPath expressions always apply to a value in the same way
 as XPath expressions are used in combination with an XML document.
-Since a Value is anonymous, JSONPath uses the abstract name `$` to
+Since a value is anonymous, JSONPath uses the abstract name `$` to
 refer to the root node of the argument.
 
 JSONPath expressions can use the *dot–notation*
@@ -354,7 +354,7 @@ JSONPath:
 # JSONPath Examples
 
 This section provides some more examples for JSONPath expressions.
-The examples are based on the simple Value shown in
+The examples are based on the simple JSON value shown in
 {{fig-example-value}}, which was patterned after a
 typical XML example representing a bookstore (that also has bicycles).
 
@@ -409,8 +409,8 @@ constant.
 | `//book[position()<3]` | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
 | `//book[isbn]`         | `$..book[?(@.isbn)]`                      | filter all books with isbn number                            |
 | `//book[price<10]`     | `$..book[?(@.price<10)]`                  | filter all books cheaper than 10                             |
-| `//*`                  | `$..*`                                    | all elements in XML document; all member values and array elements contained in Value |
-{: #tbl-example title="Example JSONPath expressions applied to the example Value"}
+| `//*`                  | `$..*`                                    | all elements in XML document; all member values and array elements contained in input value |
+{: #tbl-example title="Example JSONPath expressions applied to the example JSON value"}
 
 <!-- XXX: fine tune: is $..* really member values + array elements -->
 
@@ -436,8 +436,8 @@ A well-formed JSONPath query is valid if it also fulfills all semantic
 requirements posed by this document.
 
 The well-formedness and the validity of JSONPath queries are independent of
-the Value the query is applied to; no further errors can be
-raised during application of the query to a Value.
+the value the query is applied to; no further errors can be
+raised during application of the query to a value.
 
 (Obviously, an implementation can still fail when executing a JSONPath
 query, e.g., because of resource depletion, but this is not modeled in
@@ -494,7 +494,7 @@ selector or a set of nodes subordinate to that current node.
 ## Syntax
 
 Syntactically, a JSONPath consists of a root selector (`$`), which
-stands for a nodelist that contains the root node of the Argument,
+stands for a nodelist that contains the root node of the argument,
 followed by a possibly empty sequence of *selectors*.
 
 ~~~~ abnf
@@ -530,9 +530,9 @@ For each node in the list, the selector selects zero or more nodes,
 each of which is a descendant of the node or the node itself.
 
 > Discussion (D4): Is that a common requirement?  Or can a selector also go
-> up, or to the query Argument?
+> up, or to the query argument?
 
-For instance, with the Argument `{"a":[{"b":0},{"b":1},{"c":2}]}`, the
+For instance, with the argument `{"a":[{"b":0},{"b":1},{"c":2}]}`, the
 JSONPath `$.a[*].b` selects the following list of nodes: `0`, `1`
 (denoted here by their value).
 Let's walk through this in detail.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -156,11 +156,21 @@ their UTF-8 encoding.
 For example, the Unicode PLACE OF INTEREST SIGN (U+2318) would be defined
 in ABNF as `%x2318`.
 
-The terminology of {{-json}} applies, including definitions for
-"Value", "Object", "Array", "Number", and "String".
-Additional terms used in this specification are defined below.
+The terminology of {{-json}} applies except where clarified below.
+Definitions for "Object", "Array", and "Number" remain unchanged.
 Importantly "object" and "array" in particular do not take on a
 generic meaning, such as they would in a general programming context.
+
+Additional terms used in this specification are defined below.
+
+String:
+: A sequence of UTF-8 characters surrounded by either single quotes (`)
+  or double quotes (").
+  Escaping quote characters is required only for the specific character
+  which surrounds the string.
+  For example, a double quote character (") within a string surrounded
+  by single quotes (') does not need to be escaped.
+  Single quotes are preferred.
 
 Member:
 : A name/value pair in a JSON object.  (Not itself a value.)
@@ -197,22 +207,20 @@ Nodelist:
   While this list can be represented in JSON, e.g. as an array, the
   Nodelist is an abstract concept unrelated to JSON values.
 
-<!--
-  I feel like the below definition should be for "normalized path,"
-  but maybe that deserves its own section that also declares any
-  location included in the output Nodelist MUST be normalized.
-  (Also see line 291.) Requesting PR comments.
--->
-
-Output Path:
+Normalized Path:
 : A simple form of JSONPath expression that identifies a node by
   providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
+<!--
+  Depending on the outcome of [?()] expression support discussions,
+  we may also need to define "boolean", "number", and perhaps others.
+-->
+
 For the purposes of this specification, a value as defined by
 {{-json}} is also viewed as a tree of nodes.
 Each node, in turn, holds a value.
-Furthernodes within each value are the elements of Arrays and the
+Further nodes within each value are the elements of arrays and the
 member values of objects and are themselves values.
 (The type of the value held by a node is
 may also be referred to as the type of the node.)
@@ -420,10 +428,10 @@ constant.
 
 ## Overview {#synsem-overview}
 
-A JSONPath is a string which selects zero or more nodes of a piece of JSON.
-A valid JSONPath conforms to the ABNF syntax defined by this document.
+A JSONPath query is a string which selects zero or more nodes of a piece of JSON.
+A valid query conforms to the ABNF syntax defined by this document.
 
-A JSONPath MUST be encoded using UTF-8. To parse a JSONPath according to
+A query MUST be encoded using UTF-8. To parse a JSONPath according to
 the grammar in this document, its UTF-8 form SHOULD first be decoded into
 Unicode code points as described
 in {{RFC3629}}.
@@ -493,7 +501,7 @@ selector or a set of nodes subordinate to that current node.
 
 ## Syntax
 
-Syntactically, a JSONPath consists of a root selector (`$`), which
+Syntactically, a JSONPath query consists of a root selector (`$`), which
 stands for a nodelist that contains the root node of the argument,
 followed by a possibly empty sequence of *selectors*.
 
@@ -507,7 +515,7 @@ The syntax and semantics of each selector is defined below.
 
 ## Semantics
 
-The root selector `$` not only selects the Root node of the argument,
+The root selector `$` not only selects the root node of the argument,
 but it also produces as output a list consisting of one
 node: the argument itself.
 
@@ -533,11 +541,11 @@ each of which is a descendant of the node or the node itself.
 > up, or to the query argument?
 
 For instance, with the argument `{"a":[{"b":0},{"b":1},{"c":2}]}`, the
-JSONPath `$.a[*].b` selects the following list of nodes: `0`, `1`
+query `$.a[*].b` selects the following list of nodes: `0`, `1`
 (denoted here by their value).
 Let's walk through this in detail.
 
-The JSONPath consists of `$` followed by three selectors: `.a`, `[*]`, and `.b`.
+The query consists of `$` followed by three selectors: `.a`, `[*]`, and `.b`.
 
 Firstly, `$` selects the root node which is the argument.
 So the result is a list consisting of just the root node.
@@ -557,7 +565,7 @@ This is the concatenation of three lists, two of length one containing
 `0`, `1`, respectively, and one of length zero.
 
 As a consequence of this approach, if any of the selectors selects no nodes,
-then the whole JSONPath selects no nodes.
+then the whole query selects no nodes.
 
 In what follows, the semantics of each selector are defined for each type
 of node.
@@ -906,7 +914,7 @@ MUST raise an error.
 
 # IANA Considerations {#IANA}
 
-TBD: Define a media type for JSON Path expressions.
+TBD: Define a media type for JSONPath expressions.
 
 # Security Considerations {#Security}
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -163,36 +163,37 @@ Importantly "object" and "array" in particular do not take on a
 generic meaning, such as they would in a general programming context.
 
 Member:
-: A name/value pair in a JSON object.  (Not itself a Value.)
+: A name/value pair in a JSON object.  (Not itself a value.)
 
 Name:
 : The name in a name/value pair constituting a member.  (Also known as
   "key", "tag", or "label".)
+  This is also used in {{-json}}, but that specification does not 
+  formally define it.
+  It is included here for completeness.
 
 Element:
-: A Value in an array.  (Also used with a distinct meaning in XML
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -156,77 +156,68 @@ their UTF-8 encoding.
...
-  Never used in its generic sense, e.g., for programming language objects.
+The terminology of {{-json}} applies, including definitions for
+"Value", "Object", "Array", "Number", and "String".
+Additional terms used in this specification are defined below.
+Importantly "object" and "array" in particular do not take on a
+generic meaning, such as they would in a general programming context.
 
 Member:
-: A name/value pair in a JSON object.  (Not itself a JSON value.)
+: A name/value pair in a JSON object.  (Not itself a value.)
```
glyn, line 166:
```suggestion
: A name/value pair in an Object.  (Not itself a value.)
```

draft-ietf-jsonpath-base.md
```diff
@@ -156,77 +156,68 @@ their UTF-8 encoding.
...
-: A JSON array as defined in {{-json}}.
-  Never used in its generic sense, e.g., for programming language objects.
+  This is also used in {{-json}}, but that specification does not 
+  formally define it.
+  It is included here for completeness.
 
 Element:
-: A JSON value in an array.  (Also used with a distinct meaning in XML
+: A value in an array.  (Also used with a distinct meaning in XML
   context for XML elements.)
```
glyn, line 177: It would be good to delete this parenthetical remark if/when XML is relegated to an appendix (which could then use "XML element" instead).

cabo:
> Resolves #84
> 
> ## Alterations
> * Removed `Value`, `Object`, and `Array` since they are defined in 8259.  Replaced with extended verbiage at the beginning of the section that declares that 8259 applies.
> * Replaced "JSON value" with "value" in line with the above.

Not good.

Is there a reason you want to cause this regression?

Grüße, Carsten

gregsdennis:
> Is there a reason you want to cause this regression?

Is it a regression?  The spec already had text that it pulled these terms from 8259 and that we don't consider them in terms of their general programming usages.  It's redundant to define them again.

I think it makes sense to claim these terms as defined the same way.  We don't redefine "MUST" _et. al._, we claim them from 2119.

draft-ietf-jsonpath-base.md
gregsdennis: Will leave for a follow-up PR.

cabo:
> > Is there a reason you want to cause this regression?
> 
> Is it a regression? The spec already had text that it pulled these terms from 8259 and that we don't consider them in terms of their general programming usages. It's redundant to define them again.
> 
> I think it makes sense to claim these terms as defined the same way. We don't redefine "MUST" _et. al._, we claim them from 2119.

The current text does point to 8259 and doesn't redefine terms.  Where 8259 defines "bed" as "table", this is mentioned and possibly detailed once more, which I think is needed for readers who aren't totally immersed in 8259's terminology.
Why do you want to make it harder to use the specification?

(Geez, have you ever actually read 8259?  It doesn't actually define the terms.)

Crystallizing the definitions for use in JSONpath serves a purpose.

Why do you think it is detrimental to remind readers that object does not have the English meaning in this document?

I'd rather spend time on constructive PRs.

Grüße, Carsten

draft-ietf-jsonpath-base.md
```diff
@@ -156,8 +156,9 @@ their UTF-8 encoding.
 For example, the Unicode PLACE OF INTEREST SIGN (U+2318) would be defined
 in ABNF as `%x2318`.
 
-
-The terminology of {{-json}} applies.
+The terminology of {{-json}} applies, including definitions for
```
cabo, line 159: Please indicate where 8259 defines "string".

draft-ietf-jsonpath-base.md
```diff
@@ -163,65 +163,61 @@ Importantly "object" and "array" in particular do not take on a
...
 
 Nodelist:
-: Output of applying a query to an Argument, manifested as a list of Nodes.
+: A list of nodes.  <!-- ordered list?  Maybe TBD by issues #27 and #60 -->
+  The output of applying a query to an argument is manifested as a list of nodes.
   While this list can be represented in JSON, e.g. as an array, the
   Nodelist is an abstract concept unrelated to JSON values.
 
 <!--
   I feel like the below definition should be for "normalized path,"
```
cabo, line 201: The definitional aspect of the Output Path is that it is output.  The fact that it is normalized is certainly interesting, but not definitional.  Normalized paths could occur anywhere.  Output Paths occur at the output.

draft-ietf-jsonpath-base.md
```diff
@@ -277,7 +273,7 @@ The JSONPath tool in question should:
 JSONPath expressions always apply to a Value in the same way
 as XPath expressions are used in combination with an XML document.
 Since a Value is anonymous, JSONPath uses the abstract name `$` to
-refer to the entire Value ("root node") of the argument.
+refer to the root node of the argument.
```
cabo, line 276: The definition is not a replacement for the exposition in the text.

draft-ietf-jsonpath-base.md
```diff
@@ -459,20 +455,17 @@ the internal workings of an implementation:  Implementations may wish
...
-a valid JSONPath query is executed against a Value, the *Argument*, and
-produces a list of zero or more nodes of the Value which are
-selected by the JSONPath.
-(Note that the term Value also implies that this value complies
-to the definitions in {{-json}}, i.e., is indeed a Value.)
+a valid query is executed against a value, the *argument*, and
+produces a list of zero or more nodes of the value.
 
-The JSONPath query is a sequence of zero or more *selectors*, each of
+The query is a sequence of zero or more *selectors*, each of
```
cabo, line 461: We are still in the initial exposition, so redundancy is good here.

draft-ietf-jsonpath-base.md
gregsdennis: https://datatracker.ietf.org/doc/html/rfc8259#section-7

cabo: Yes.

```
   A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks, except for the characters that MUST be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).
```

These are not the strings JSONPath uses.

draft-ietf-jsonpath-base.md
gregsdennis: - `Node` is defined as a value and a location.
- `Nodelist` is defined as a collection of nodes.  The definition also mentions that the output of a query is a nodelist.
- Both the "Processing Model" and "Sematics" sections also mention that the output is a nodelist.  Moreover, "Processing Model" says that each selector produces a nodelist which is then used as the input to the next selector.

It seems to me that the output is already well-defined by the above.

If we also declare that the location portion of a node needs to be represented with a normalized path, then we get for free that the paths in the output are a normalized path (because the output is a list of nodes), and we don't need to define what an "output path" is.

However, "normalized path" still needs to be defined.  This is illustrated by your statement:

> Normalized paths could occur anywhere.

I don't think we need to have a special-case definition for a path that occurs in the output.

draft-ietf-jsonpath-base.md
gregsdennis: Please explain why.  The root node is by definition the entire argument.

draft-ietf-jsonpath-base.md
gregsdennis: The word "query" has no other usage within this spec other than to represent a JSONPath query.  The redundancy doesn't help clarify anything.  (It also appears I missed removing some self-references.)

draft-ietf-jsonpath-base.md
gregsdennis: I think the ones you're thinking of relate to "names" in a bracket-selector or filter expression.  These are currently discussed in the "union selector" -> "child" -> "syntax" section, where it's given effectively the same definition as 8259 except that single quotes are also allowed.

I see the argument to add a "string" definition to account for the support of single-quotes.

gregsdennis:
> Where 8259 defines "bed" as "table"... 

I think we have a confusion.  When I say "8259", mean [this RFC](https://datatracker.ietf.org/doc/html/rfc8259) (also denoted by `{{-json}}` in the raw text) which is what it appears the spec is referencing. It doesn't even _contain_ the word "bed".  So I have no idea what you're talking about.

That spec very clearly defines all of the terms that I've removed.

> Crystallizing the definitions for use in JSONpath serves a purpose.

My concern is redefinition.  It doesn't make sense to define it in two places.  The common practice (as I've seen) is to refer to existing definitions.  Since we're relying heavily on the definition of JSON anyway, it makes sense to just reference those definitions.  Anyone unfamiliar with JSON should want to read that spec as well.

> Why do you think it is detrimental to remind readers that object does not have the English meaning in this document?

I _do_ keep the callout that we're not using the English meaning of "object" and "array." I've just moved it up to the text where I claim the definitions from the JSON spec.

cabo:
On 13. May 2021, at 22:08, Greg Dennis ***@***.***> wrote:
> 
> Where 8259 defines "bed" as "table"...
> 
> I think we have a confusion.
> 

Yes.  It just didn’t occur to me that this cultural reference may be completely obvious to Germans but almost inaccessible to people from other native languages.

Here are references to the short story that should be mandatory reading for anyone working on terminology (or, actually, anyone communicating).
(Germans tend to encounter it in school, at around 8th grade.)

https://www.deutschunddeutlich.de/contentLD/GD/GT67cTischistTisch.pdf
http://www.barbarafaessler.com/files/BichselAtableisatable.pdf (English translation of unknown quality)

Grüße, Carsten

gregsdennis:
Interesting story, but I don't really get your purpose in referencing it.  We're not swapping terms between our spec and 8259.

goessner:
just to let you know ... I won't be active until Wednesday next week ...

Cheers
--
sg

glyn:
I find the definitions in 8259 rather confusing. Take _object_ for example. 8259 contains the following statements:
1. A JSON text is a serialized value.
2. `JSON-text = ws value ws`
3. A JSON value MUST be an object, array, number, or string, or one of the following three literal names: [...]
4. An object structure is represented as a pair of curly brackets surrounding zero or more name/value pairs (or members).

One way to read statement 1 is that a value is serlalized to produce JSON text and therefore a value is not itself in serial form. But statement 2 seems to imply that a value _is_ in serial form. So I think statement 1 has to be read as "A JSON text is a value in a serialized form."

If a value is in a serialized form, then statement 3 implies that an object is also in a serialized form.

But statement 4 is then confusing:
* It seems to be defining an object structure rather than an object.
* It's probably trying to define an object and say that an object is a structure (in the normal use of the word).
* It says an object (structure) "_is represented as_ a pair of curly brackets [...]", but if an object is a serialized thing, it _is_ a pair of curly brackets etc. (Or, perhaps better, an object _consists of_ a pair of curly brackets etc.)

I'm not sure whether we should attempt to clarify this in the JSONPath spec or submit an erratum to 8259 - probably the latter makes more sense.

cabo:
8259 is what it is.  It was created in a politically difficult situation, and it turned out better than many of us had feared.  But it is a rather poor document, and the errata process is not designed to fix its problems.

But the situation is actually worse.  When writing the JSONpath specification, we have to think about its audience and not about how we ourselves would like this document to be.  Most people who will read the JSONpath specification will come out of one of the JSON sub-ecosystems and will think that they know what JSON is.  They won't read 8259 just because we reference it.  It simply is not helpful to defer to 8259, we have to make sure that the terms we need are introduced properly.  (Not *defined* -- that is indeed what 8259 needs to be the authoritative reference for.). This is more important because the 8259 terminology is peculiar (its accidental or deliberate ambiguities served its purpose well in the painstaking 8259 process, but that doesn't mean it's particularly good for specifications that actually make use of JSON).

I wish we could focus on creating a good document for its audience, instead of discussing whether something theoretically needs to be brought up or not.

gregsdennis:
@cabo that was well-stated.  I will try to make another round of edits to reinsert these in such a way that clarity is provided without redefining them.

draft-ietf-jsonpath-base.md
goessner: In fact the root node is not a node initially, as it lacks a name by nature. Similar to your formulation, the root node formally becomes a node by always getting the name '$' assigned by convention. 

Your formulation

`... refer to the root node of the argument`

somehow contradicts your comment

`The root node is by definition the entire argument.` 

and also somehow confuses below  by ...

`... executed against a value, the *argument* ...`

BTW the definition above  is cristal clear, as the *argument* is the value of the *root node*.

draft-ietf-jsonpath-base.md
goessner: I think we agree with the following ...

* JSONPath is *selecting* only, not *processing*.
* Analytical post processing on the query result is heavily demanded by users.

I do like that brillant *extension points* concept a lot. So question is: Can we open the door for analytical post processing via those *extension points*? If so, the spec must provide certainty to implementors regarding *output path* format, which should be either/or ...

* list of selected values
* list of normalized pathes

So an *output path* always should have *normalized path* format in my opinion.

gregsdennis: > So an output path always should have normalized path format in my opinion.

I'm not arguing this.  I'm questioning whether we need "output path" as an explicit definition.

To me, "output path" is just a normalized path that happens to be in the output.  There's nothing really special about it that requires a definition.  We _do_ need to define the output somewhere, but it's not a definition or term; it's a feature.

"Normalized path," on the other hand, is precisely what the current definition is describing:

> A simple form of JSONPath expression that identifies a node by providing a query that results in exactly that node.

draft-ietf-jsonpath-base.md
gregsdennis: > In fact the root node is not a node initially - @goessner 

Not quite.  "Argument" is already defined as a value (not a node):

```
Short name for the value a JSONPath expression is applied to.
```

This means that "entire value of the argument" is redundant.  My change simply removes the redundancy.  "Root node of the argument" identifies the top-level location of the initial JSON value.

gregsdennis:
@cabo @goessner @glyn I've made some updates.  Just tagging you in.

draft-ietf-jsonpath-base.md
```diff
@@ -156,77 +156,81 @@ their UTF-8 encoding.
...
-  that argument.
+: A list of nodes.  <!-- ordered list?  Maybe TBD by issues #27 and #60 -->
+  The output of applying a query to an argument is manifested as a list of nodes.
   While this list can be represented in JSON, e.g. as an array, the
-  nodelist is an abstract concept unrelated to JSON values.
+  Nodelist is an abstract concept unrelated to JSON values.
 
-Output Path:
+Normalized Path:
 : A simple form of JSONPath expression that identifies a node by
```
glyn, line 216: This permits multiple possibilities. Should it be tightened up to say which types of selector may and may not be used?

glyn (approved):

cabo:
This draft is still mostly a collection of regressions.
E.g., the section about strings is seriously misleading.
There is no point in this spec explaining JSON inner workings.
(And, of course, that is similar to the symptoms of acting like the definitions in 8259, which aren't really proper definitions anyway, are useful here.)

draft-ietf-jsonpath-base.md
```diff
@@ -156,77 +156,81 @@ their UTF-8 encoding.
...
-: A JSON value, with an emphasis on its location within the
-  Argument.  I.e., a JSON value that is identical to or contained within the JSON
-  value to which the query is applied.  A node can be viewed as a
-  combination of a (1) JSON value and (2) its location in the
-  Argument; the latter can, if desired, be represented as an Output Path.
-
-Object:
-: A JSON object as defined in {{-json}}.
-  Never used in its generic sense, e.g., for programming language objects.
+String:
```
cabo, line 171: That is *NOT* what a string is for the purpose of this specification.
We do not need to confuse our readers.
(It also is completely unrelated to the JSON syntax.)

cabo (changes requested):
I really don't have time to reread the entire set of changes, but this snippet seriously discourages me from spending any time at all.

draft-ietf-jsonpath-base.md
gregsdennis: Maybe you should try offering an alternative because this is PRECISELY what a string is for JSON Path: a single- or double-quoted sequence of characters.  JSON uses UTF-8, so that's what I proposed.  If you think it can be better, how so?

draft-ietf-jsonpath-base.md
```diff
@@ -156,77 +156,81 @@ their UTF-8 encoding.
...
+: A sequence of UTF-8 characters surrounded by either single quotes (`)
```
glyn, line 172: Replace the back-tick with a single quote:
```suggestion
: A sequence of UTF-8 characters surrounded by either single quotes (')
```

glyn (changes requested):
Looks ok to me except the erroneous back-tick needs to be replaced with a single quote.

I see this PR as not necessarily the end of clarifying the terminology, but as a good step along the road.

draft-ietf-jsonpath-base.md
gregsdennis: Committed. Nice catch.

glyn (approved):

cabo:
Oops.  The commit text "slightly fixed #97" triggered the automatic close of this PR.
That was not my intention...

draft-ietf-jsonpath-base.md
```diff
@@ -168,13 +168,19 @@ Value:
...
+  which surrounds the string.
+  For example, a double quote character (") within a string surrounded
+  by single quotes (') does not need to be escaped.
+  Single quotes are preferred.
 
 Name:
 : The name in a name/value pair constituting a member.  (Also known as
   "key", "tag", or "label".)
-  This is also used in {{-json}}, but that specification does not
+  This is also used in {{-json}}, but that specification does not
```
cabo, line 183: SO do you want to keep the deletion of "member" and this and the below changes (which would unfix some nits)?

gregsdennis:
All pertinent details replicated inn #102.  Closing this.

Issue #64, opened 2021-03-10
Filter Expressions


fiestajetsam:
(This was original raised in #54 and now being split out)

> Related to that, it would be helpful to determine if JSONPath filters
apply to both JSON objects and arrays, or only to JSON arrays.
*(Daniel P)*

> I would support restricting filters to arrays, if others agree.
*(Glyn Normington)*

I tend to let implementations and their "normative force of the factual" decide here or in doubt agree to Glyn's restriction to arrays.

I am very unhappy with confusing `$..book[(@.length-1)]`, where `'@'` addresses the array itself and implies that array has a `length` property. In filter expression examples `'@'` more consistently addresses the current array element.

The invocation of 'the underlying scripting engine' wasn't meant a serious normative aspect, but rather a quick and dirty solution for JavaScript and PHP implementations at that time.

### Corner Case

> Consider this perfectly legal JSON object
>
> ```{ "ab": 0,  "'a.b": 1,  "a-b": 2, "a": { "b": 3 } }```
>
>So `$.ab` is 0, `$.a.b` is 3, `$['a.b']` is 1, `$['a-b']` is 2. You'd like to say `$.a-b` but lots of libraries will refuse it because `"a-b"` is not a legal JavaScript "name" construct, that's why you have to say `$['a-b']`.
>
> But suppose your library would accept `$.a-b`.  Then `$.a-b` and `$['a-b']` would be synonyms, but `$.a.b` and `$['a.b']` wouldn't.
*(Tim Bray)*

Hmm ... this seems to be a hint to better exclude `'-'` from dot-child-selector syntax. I think I have read more discussion about that, currently don't know where.

gregsdennis:
This actually already has an issue: #47.

danielaparker:
> This actually already has an issue: #47.

And #17

gregsdennis:
Can we close this?

goessner:
Filter expressions `[?(<expr>)]` are completely missing in the current draft.

Implementations agree, that filter expressions are an essential feature with JSONPath. They are considered much more important than "index expressions" `[(<expr>)]`, given that term "index" is an interim synonym for "name / index" ([#84](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/84#issuecomment-809245242)).

There are already some discussions about filters (historically ascending):

[...]

PR #99, opened 2021-06-10
Updated Pr 98 (selectors reworked)

cabo:
mit Fix für den YAML-nit

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -101,8 +101,8 @@ informative:
     title: >
       Proposed ECMAScript 4 Edition – Language Overview
     author:
-    - org: Adobe Systems Inc., The Mozilla Foundation, Opera Software 
-ASA, and others 
+    - org: >
+        Adobe Systems Inc., The Mozilla Foundation, Opera Software ASA, and others
     date: 2007
   SLICE:
     target: https://github.com/tc39/proposal-slice-notation
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -97,6 +97,13 @@ informative:
     seriesinfo:
       ISO/IEC 22537:2006
     date: 2006
+  E4X-overview:
+    title: >
+      Proposed ECMAScript 4 Edition – Language Overview
+    author:
+    - org: Adobe Systems Inc., The Mozilla Foundation, Opera Software 
+ASA, and others 
+    date: 2007
   SLICE:
     target: https://github.com/tc39/proposal-slice-notation
     title: Slice notation
@@ -112,6 +119,7 @@ normative:
   RFC5234: abnf
   RFC8259: json
 
+...
 --- abstract
 
 JSONPath defines a string syntax for identifying values
@@ -333,7 +341,7 @@ $.store.book[?(@.price < 10)].title
 Here is a complete overview and a side by side comparison of the JSONPath syntax elements with their XPath counterparts.
 
 | XPath | JSONPath           | Description                                                                                                                           |
-|-------+--------------------+---------------------------------------------------------------------------------------------------------------------------------------|
+|-------|--------------------|---------------------------------------------------------------------------------------------------------------------------------------|
 | `/`   | `$`                | the root element/node                                                                                                             |
 | `.`   | `@`                | the current element/node                                                                                                          |
 | `/`   | `.` or `[]`        | child operator                                                                                                                        |
@@ -510,8 +518,13 @@ stands for a nodelist that contains the root node of the Argument,
 followed by a possibly empty sequence of *selectors*.
 
 ~~~~ abnf
-json-path = root-selector *selector
-root-selector = "$"               ; $ selects document root node
+json-path = root-selector *(dot-selector        /
+                            dot-wild-selector   /
+                            index-selector      /
+                            index-wild-selector /
+                            slice-selector      /
+                            descendant-selector /
+                            filter-selector)
 ~~~~
 
 The syntax and semantics of each selector is defined below.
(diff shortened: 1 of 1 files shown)


gregsdennis:
@goessner can you rename the PR to better reflect what it's trying to do, please?  Also a summary of changes would be kind.

cabo:
This PR (actually PR #98) reverts all the fixes in 9be270d
Is this intentional?

goessner:
I understood that multiple small PR's are preferrable, which was not so 
easy with the complete rewrite of selectors. I can do that from now on.

Am 11.06.2021 um 05:42 schrieb Greg Dennis:

glyn:
> I understood that multiple small PR's are preferrable, which was not so easy with the complete rewrite of selectors. I can do that from now on.

Do you mean you will now break this PR into multiple small PRs? (That would be great!)

draft-ietf-jsonpath-base.md
```diff
@@ -911,8 +943,128 @@ size doesn't
...
+
+rel-path-val = "@" *(dot-selector / index-selector)
+calc_val     = func "(" [rel-path-val / json-path] ")"
+func         = "index"
+~~~~
+
+Notes:
+
+* Parenthesis can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `'[?(<expr)]'` is naturally contained in the current lean syntax `'[?<expr]'` as a special case.
+* Comparisons are restricted to primitive values `number`, `string`, `true`, `false`, `null`. Comparisons with complex values will fail, i.e. no selection occurs.
```
glyn, line 1030: Rather than fail silently, another option would be to avoid this situation in the syntax. Then failures will produce a syntax error, which is much clearer to the user. Have you checked for consensus of current implementations? [This issue comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/64#issuecomment-838280246) is relevant.

gregsdennis: I would actually like to be able to say `@.foo == [1,2]`.  I think this is a valid use case.

cabo: So it appears we need a new issue.

* Comparison with structured values
* Should comparison with structured values (e.g., `@.foo == [1, 2]`) be supported?
* If it is not supported, should this silently fail or the attempt cause a syntax error (in #99, it causes a syntax error, but then the text says something else).

draft-ietf-jsonpath-base.md
```diff
@@ -97,6 +97,13 @@ informative:
     seriesinfo:
       ISO/IEC 22537:2006
     date: 2006
+  E4X-overview:
```
cabo, line 100: I don't think we need this -- this seems to be a remnant from an older revision.

There are several such remnants, which make it hard to review this PR.

draft-ietf-jsonpath-base.md
```diff
@@ -577,207 +590,227 @@ of node.
...
-More general child names, such as the empty string, are supported by "Union
-Child" ({{unionchild}}{: format="default"}).
-
-Note that the `dot-child-name` rule follows the philosophy of JSON strings and is
-allowed to contain bit sequences that cannot encode Unicode characters (a
-single unpaired UTF-16 surrogate, for example).
-The behaviour of an implementation is undefined for child names which do
-not encode Unicode characters.
-
+root-selector  = %x24                 ; $
```
cabo, line 614: (Already discussed.)

draft-ietf-jsonpath-base.md
```diff
@@ -577,207 +590,227 @@ of node.
...
-
-The member name of a dot child name is the sequence of Unicode characters contained
-in that name.
-
-A dot child name consisting of a single asterisk is a wild card. It selects
-all the values of any object node.
-It also selects all the elements of any array node.
-It selects no nodes from
-number, string, or literal nodes.
+The Argument &mdash; the root JSON value &ndash; is anonymous by nature. By getting assigned the universal name `'$'` it becomes the root node.
```
cabo, line 620: Writing the actual characters instead of HTML entity references is preferred.

The same dashes should be used at the start and at the end of an interjection.

draft-ietf-jsonpath-base.md
```diff
@@ -577,207 +590,227 @@ of node.
...
-              "n" /          ; n    line feed       U+000A
-              "r" /          ; r    carriage return U+000D
-              "t" /          ; t    tab             U+0009
-              "u" 4HEXDIG )  ; uXXXX                U+XXXX
 
+### Index Selector
 
-dq-unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
+#### Syntax
+{: numbered="false" toc="exclude"}
```
cabo, line 675: If we want to use unnumbered sections more often, we might want to make an ALD for this to reduce noise.
Unnumbered sections are a bit of a boobytrap as they have to be used consistently in the XML.

draft-ietf-jsonpath-base.md
```diff
@@ -577,207 +590,227 @@ of node.
...
+index-selector      = "[" (quoted-member-name / element-index) "]"
+~~~~
 
-escape = "\"                 ; \
+Applying the `index-selector` to an object value, a `quoted-member-name` string is required. JSONPath allows it to be enclosed in _single_ or _double_ quotes.
 
-HEXDIG =  DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
-                              ; case insensitive hex digit
+~~~~ abnf
+quoted-member-name  =  %x22 *double-quoted %x22 /       ; "string"
```
cabo, line 686: I think we should have a separate set of definitions for literals.  This is particular important because JSONPath literals have a more permissive syntax than that of JSON.  (Should they?)
Literals should never be confused with the data themselves (i.e., we never should say "string" when we mean "string literal").

cabo (commented):
A few initial impressions.

draft-ietf-jsonpath-base.md
gregsdennis: > Literals should never be confused with the data themselves (i.e., we never should say "string" when we mean "string literal").

What?  When was this discussed?  Do we now have to say "boolean literal" or "numeric literal?"  Where does JSON Path ever use strings that AREN'T a string literal?

goessner:
> ------------------------------------------------------------------------
>
> In draft-ietf-jsonpath-base.md 
> <https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/99#discussion_r649782358>:
>
> > @@ -97,6 +97,13 @@ informative:
>       seriesinfo:
>         ISO/IEC 22537:2006
>       date: 2006
> +  E4X-overview:
>
> I don't think we need this -- this seems to be a remnant from an older 
> revision.
>
> There are several such remnants, which make it hard to review this PR.
>

oops ... should have been ES4 overview

http://archives.ecma-international.org/2007/misc/overview.pdf

Here array slice is documented the first time (pg.34)  ... that's all.
We can leave it out.

> >  
>   #### Semantics
>   {: numbered="false" toc="exclude"}
>   
> -A dot child name which is not a single asterisk (`*`) is considered to
> -have a member name.
> -It selects the value corresponding to the name from any object node.
> -It selects
> -no nodes from a node which is not a object.
> -
> -The member name of a dot child name is the sequence of Unicode characters contained
> -in that name.
> -
> -A dot child name consisting of a single asterisk is a wild card. It selects
> -all the values of any object node.
> -It also selects all the elements of any array node.
> -It selects no nodes from
> -number, string, or literal nodes.
> +The Argument &mdash; the root JSON value &ndash; is anonymous by nature. By getting assigned the universal name `'$'` it becomes the root node.
>
> Writing the actual characters instead of HTML entity references is 
> preferred.
>
&ndash; is not accessible per keyboard.
>
> The same dashes should be used at the start and at the end of an 
> interjection.
>
yes .. typo ..

goessner:
yes .. we need a new issue here .. let's discuss.

Am 11.06.2021 um 10:12 schrieb cabo:

cabo:
> > + E4X-overview:
> 
> oops ... should have been ES4 overview
> 
> http://archives.ecma-international.org/2007/misc/overview.pdf
> 
> Here array slice is documented the first time (pg.34)  ... that's all.
> We can leave it out.

I don’t have a strong opinion here, but I think we are better off referencing the current state in ES2020 (also applies to {{SLICE}}).

(The bibliography entry `E4X-overview` turned up in my quick glance at the PR because it is not currently being used.  So we need to add a citation or remove the entry.  This is all housekeeping…  Let’s try to extract the technical discussion points first.)

Grüße, Carsten

cabo:
> &ndash; is not accessible per keyboard.

It sure is.
This probably should be an &mdash;, however.

If you are using Windows:
https://softwareaccountant.com/em-dash-in-word/

goessner:
ahh ... learning something new .. only*alt+0151 *on numeric keyboard 
works for me.**

Am 14.06.2021 um 10:07 schrieb cabo:

cabo:
On 2021-06-14, at 10:28, Stefan Goessner ***@***.***> wrote:
> 
> ahh ... learning something new .. only*alt+0151 *on numeric keyboard 
> works for me.**

… and I think this is even easier to type than &mdash; …

(Shift-Option-minus on the Mac.
On the iPad or iPhone, long-press the minus and swipe to the character you want.)

Grüße, Carsten

glyn:
@cabo please rebase

draft-ietf-jsonpath-base.md
cabo: (Resolved in #102)

draft-ietf-jsonpath-base.md
cabo: Done in #102

cabo:
Covered by #102 now.

PR #101, opened 2021-06-12
Slightly fixed #97

cabo:


sthagen (approved):
LGTM - as a first step to „pull the cow from the ice“

goessner (commented):
well .. let's go on and asymptotically approach the optimum.

goessner (approved):

glyn (approved):

PR #102, opened 2021-06-14
Selectors reworked ...

cabo:
This is a full rebase of branch pr-98 (#99) on current main.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -520,6 +520,7 @@ json-path = root-selector *(dot-selector        /
                             dot-wild-selector   /
                             index-selector      /
                             index-wild-selector /
+                            union-selector      /
                             slice-selector      /
                             descendant-selector /
                             filter-selector)
@@ -627,12 +628,13 @@ A dot selector starts with a dot `.` followed by an object's member name.
 
 ~~~~ abnf
 dot-selector    = "." dot-member-name
-dot-member-name = 1*(
-                      DIGIT /
+dot-member-name = name-first *name-char
+name-first =
                       ALPHA /
                       "_"   /           ; _
                       %x80-10FFFF       ; any non-ASCII Unicode character
-                    )
+name-char = DIGIT / name-first
+
 DIGIT           =  %x30-39              ; 0-9
 ALPHA           =  %x41-5A / %x61-7A    ; A-Z / a-z
 ~~~~
@@ -648,6 +650,7 @@ characters — MUST NOT be used with the `dot-selector`.
 
 The `dot-selector` selects the value corresponding to the member name from any JSON object. It selects no values from any other JSON value.
 
+<!-- Not true, as JSONPath queries are UTF-8 texts -->
 Note that the `dot-selector` follows the philosophy of JSON strings and is
 allowed to contain bit sequences that cannot encode Unicode characters (a
 single unpaired UTF-16 surrogate, for example).
@@ -689,8 +692,10 @@ index-selector      = "[" (quoted-member-name / element-index) "]"
 Applying the `index-selector` to an object value, a `quoted-member-name` string is required. JSONPath allows it to be enclosed in _single_ or _double_ quotes.
 
 ~~~~ abnf
-quoted-member-name  =  %x22 *double-quoted %x22 /       ; "string"
-                       %x27 *single-quoted %x27         ; 'string'
+quoted-member-name  = string-literal
+
+string-literal      = %x22 *double-quoted %x22 /       ; "string"
+                      %x27 *single-quoted %x27         ; 'string'
 
 double-quoted       = unescaped /
                       %x27      /                       ; '
@@ -709,18 +714,26 @@ unescaped           = %x20-21 /                         ; s. RFC 8259
                       %x28-5B /                         ; omit '
                       %x5D-10FFFF                       ; omit \
 
-escapable           = (
-                          b /         ;  BS backspace U+0008
-                          t /         ;  HT horizontal tab U+0009
-                          n /         ;  LF line feed U+000A
-                          f /         ;  FF form feed U+000C
-                          r /         ;  CR carriage return U+000D
-                          / /         ;  /  slash (solidus)
-                          \ /         ;  \  backslash (reverse solidus)
-                          u 4HEXDIG   ;  uXXXX      U+XXXX
+escapable           = ( %x62 / %x66 / %x6E / %x72 / %x74 / ; \b \f \n \r \t
+                          ; b /         ;  BS backspace U+0008
+                          ; t /         ;  HT horizontal tab U+0009
+                          ; n /         ;  LF line feed U+000A
+                          ; f /         ;  FF form feed U+000C
+                          ; r /         ;  CR carriage return U+000D
+                          "/" /          ;  /  slash (solidus)
+                          "\" /          ;  \  backslash (reverse solidus)
+                          (%x75 hexchar) ;  uXXXX      U+XXXX
                       )
-HEXDIG              = %x41-46 /           ;  A-F
-                      %x61-66             ;  a-f
+
+hexchar = non-surrogate / (high-surrogate "\" %x75 low-surrogate)
+non-surrogate = ((DIGIT / "A"/"B"/"C" / "E"/"F") 3HEXDIG) /
+                 ("D" %x30-37 2HEXDIG )
+high-surrogate = "D" ("8"/"9"/"A"/"B") 2HEXDIG
+low-surrogate = "D" ("C"/"D"/"E"/"F") 2HEXDIG
+
+HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
+
+; Task from 2021-06-15 interim: update ABNF later
 ~~~~
 
 Applying the `index-selector` to an array, a numerical `element-index` is required. JSONPath allows it to be negative.
@@ -733,10 +746,10 @@ DIGIT1          = %x31-39                         ; 1-9 non-zero digit
 ~~~~
 
 Notes:
-1. `double-quoted` strings follow JSON in {{RFC8259}}.
+1. `double-quoted` strings follow JSON in {{RFC8259}};
    `single-quoted` strings follow an analogous pattern.
-2. An `element-index` is an integer (in base 10).
-3. The syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.
+2. An `element-index` is an integer (in base 10, as in JSON numbers).
+3. As in JSON numbers, the syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.
 
 #### Semantics
 {: unnumbered}
@@ -1027,12 +1040,16 @@ The current item is selected if and only if the result is `true`.
 
 
 ~~~~ abnf
-boolean-expr = *logical-expr
-logical-expr = [neg-op] ["("] comp-expr *[logical-op comp-expr] [")"]
+
+
+boolean-expr = logical-expr
+logical-expr = ([neg-op] primary-expr) / logical-or-expr
 neg-op       = "!"                                  ; not operator
-logical-op   = "||" / "&&"                          ; logical operator
-comp-expr    = (rel-path-val /
-                calc-val) [(comp-op comparable /  ; comparison
+primary-expr = "(" logical-or-expr ")"
+logical-or-expr = logical-and-expr *["||" logical-and-expr]
+logical-and-expr = comp-expr *["&&" comp-expr]
+
+comp-expr    = (rel-path-val) [(comp-op comparable /  ; comparison
                             regex-op regex     /  ; RegEx test
                             in-op container )]    ; containment test
 comp-op      = "==" / "!=" /                        ; comparison ...
@@ -1040,15 +1057,15 @@ comp-op      = "==" / "!=" /                        ; comparison ...
                "<=" / ">="
 regex-op     = "~="                                 ; RegEx match
 in-op        = " in "                               ; in operator
-comparable   = number / quoted-string /             ; primitive ...
+comparable   = number / string-literal /            ; primitive ...
                true / false / null /                ; values only
                rel-path-val /                       ; descendant value
-               calc_val /                           ; calculated value
                json-path                            ; any value
 
 rel-path-val = "@" *(dot-selector / index-selector)
-calc_val     = func "(" [rel-path-val / json-path] ")"
-func         = "index"
+
+container = <TO BE DEFINED>
+regex = <TO BE DEFINED>
 ~~~~
 
 Notes:
@@ -1097,7 +1114,28 @@ Some examples:
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
 
+# Expression Language
 
+> Task (T2): Separate out expression language.  For now, this section
+> is a repository for ABNF taken from {{-json}}.  This needs to be
+> deduplicated with definitions above.
+
+~~~~ abnf
+number = [ minus ] jsint [ frac ] [ exp ]
+decimal-point = %x2E       ; .
+digit1-9 = %x31-39         ; 1-9
+e = %x65 / %x45            ; e E
+exp = e [ minus / plus ] 1*DIGIT
+frac = decimal-point 1*DIGIT
+jsint = zero / ( digit1-9 *DIGIT )
+minus = %x2D               ; -
+plus = %x2B                ; +
+zero = %x30                ; 0
+
+false = %x66.61.6c.73.65   ; false
+null  = %x6e.75.6c.6c      ; null
+true  = %x74.72.75.65      ; true
+~~~~
 
 # IANA Considerations {#IANA}
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -841,7 +841,8 @@ The `slice-selector` consists of three optional decimal integers separated by co
 #### Semantics
 {: unnumbered}
 
-The `slice-selector` was inspired by the slice operator of ECMAScript 4 (ES4), which was deprecated in 2014 and later adopted by Python.
+The `slice-selector` was inspired by the slice operator of ECMAScript
+4 (ES4), which was deprecated in 2014, and that of Python.
 
 
 ##### Informal Introduction
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -793,7 +793,9 @@ The `index-wild-selector` behaves identically to the `dot-wild-selector`.
 #### Syntax
 {: unnumbered}
 
-The array slice selector has the form `'[<start>:<end>:<step>]'`. It selects elements starting at index `<start>`, ending at &ndash; not including &ndash; `<end>`, while incrementing by `step`.
+The array slice selector has the form `'[<start>:<end>:<step>]'`.
+It selects elements starting at index `<start>`, ending at — but
+not including — `<end>`, while incrementing by `step`.
 
 ~~~~ abnf
 slice-selector = "[" slice-index "]"
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -140,6 +140,9 @@ Comments and issues may be directed to this document's
 {:unnumbered: numbered="false" toc="exclude"}
 <!-- use as {: unnumbered} -->
 
+<!-- editorial issue: lots of complicated nesting of quotes, as in -->
+<!-- `"13 == '13'"` or `'$'`.  We probably should find a simpler style -->
+
 # Introduction
 
 This document picks up the popular JSONPath specification dated
@@ -1041,12 +1044,15 @@ Notes:
 
 * Parentheses can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `'[?(<expr>)]'` is naturally contained in the current lean syntax `'[?<expr>]'` as a special case.
 * Comparisons are restricted to primitive values `number`, `string`, `true`, `false`, `null`. Comparisons with complex values will fail, i.e. no selection occurs.
-* Implicit type conversions during comparisons are not performed. So `"13 == '13'"` selects nothing.
+<!-- issue: comparison with structured value -->
+* Types are not implicitly converted in comparisons.
+  So `"13 == '13'"` selects no value.
 * A member or element value by itself is *falsy* only, if it does not exist. Otherwise it is *truthy*, resulting in its value. To be more specific explicit comparisons are necessary. This existence test — as an exception of the general rule — also works with complex values.
 * Regular expression tests can be applied to `string` values only.
 * Containment tests work with arrays and objects.
 * Explicit boolean type conversion is done by the not operator `neg-op`.
 * The behaviour of operators is consistent with the 'C'-family of programming languages.
+<!-- need to clarify -->
 
 #### Semantics
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -780,7 +780,13 @@ index-wild-selector    = "[" "*" "]"  ;  asterisk enclosed by brackets
 #### Semantics
 {: unnumbered}
 
-An index selector combined with a single asterisk is a wild card. It selects all member values of an object as well as all elements of an array. Applying the `index-wild-selector` to primitive JSON values, as number, string, or literal, does nothing. The `index-wild-selector` behaves identically to the `dot-wild-selector`.
+An index selector combined with a single asterisk is a wild card. It
+selects all member values of an object as well as all elements of an
+array.
+Applying the `index-wild-selector` to a primitive JSON value (such as
+a number, string, or true/false/null) selects no value.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -569,206 +581,226 @@ of node.
...
-              "f" /          ; f    form feed       U+000C
-              "n" /          ; n    line feed       U+000A
-              "r" /          ; r    carriage return U+000D
-              "t" /          ; t    tab             U+0009
-              "u" 4HEXDIG )  ; uXXXX                U+XXXX
-
+index-selector      = "[" (quoted-member-name / element-index) "]"
+~~~~
 
-dq-unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
```
cabo, line 674: In the next PR, I'd prefer if we could move all the expression language parts over to an expression language section.

draft-ietf-jsonpath-base.md
```diff
@@ -569,206 +581,226 @@ of node.
...
+
+escapable           = (
+                          b /         ;  BS backspace U+0008
+                          t /         ;  HT horizontal tab U+0009
+                          n /         ;  LF line feed U+000A
+                          f /         ;  FF form feed U+000C
+                          r /         ;  CR carriage return U+000D
+                          / /         ;  /  slash (solidus)
+                          \ /         ;  \  backslash (reverse solidus)
+                          u 4HEXDIG   ;  uXXXX      U+XXXX
```
cabo, line 704: See https://www.ietf.org/archive/id/draft-bormann-cbor-cddl-freezer-07.html#name-err6527 for how to do this right.  (Again, this should be in the expression language section.)

draft-ietf-jsonpath-base.md
```diff
@@ -902,8 +933,128 @@ size doesn't
...
+               json-path                            ; any value
+
+rel-path-val = "@" *(dot-selector / index-selector)
+calc_val     = func "(" [rel-path-val / json-path] ")"
+func         = "index"
+~~~~
+
+Notes:
+
+* Parenthesis can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `'[?(<expr)]'` is naturally contained in the current lean syntax `'[?<expr]'` as a special case.
```
cabo, line 1019: What is the meaning of `<expr`?
Missing `>`?

goessner: yes ... simple typo

thanks a lot

goessner (approved):

draft-ietf-jsonpath-base.md
goessner: Wouldn't make it sense then to simple copy that into here ?

gregsdennis:
Does this incorporate any changes from the discussions from #98, or is it a simple rebase?

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +585,241 @@ of node.
...
+                    )
+DIGIT           =  %x30-39              ; 0-9
+ALPHA           =  %x41-5A / %x61-7A    ; A-Z / a-z
+~~~~
 
+Member names containing other characters than allowed by
+`dot-selector` — such as space `' '` and minus `'-'`
+characters — MUST NOT be used with the `dot-selector`.
+(Such member names can be addressed by the
+`index-selector` instead.)
```
gregsdennis, line 640: Having this as a parenthetical seems good to me. 👍

cabo:
> Does this incorporate any changes from the discussions from #98, or is it a simple rebase?

It is a simple rebase, followed by a number of edits coming from the discussions here.
Most of them need some further discussion, so I hope to cover them with my updated slides in a couple of hours.

draft-ietf-jsonpath-base.md
cabo: Yes.
(Discuss today.)

draft-ietf-jsonpath-base.md
```diff
@@ -793,7 +793,9 @@ The `index-wild-selector` behaves identically to the `dot-wild-selector`.
 #### Syntax
 {: unnumbered}
 
-The array slice selector has the form `'[<start>:<end>:<step>]'`. It selects elements starting at index `<start>`, ending at &ndash; not including &ndash; `<end>`, while incrementing by `step`.
+The array slice selector has the form `'[<start>:<end>:<step>]'`.
+It selects elements starting at index `<start>`, ending at — but
+not including — `<end>`, while incrementing by `step`.
```
gregsdennis, line 799: This doesn't quite match up with what's actually going on.  Hence [the issue](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/issues/3) in the compliance test suite I raised.

I was thinking it worked as you've written, but the referenced algorithm doesn't work exactly that way.

gregsdennis (changes requested):
There are also still some open questions from #98.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +585,241 @@ of node.
...
 ~~~~
 
 Notes:
-1. double-quoted strings follow JSON in {{RFC8259}}.
-   Single-quoted strings follow an analogous pattern.
-2. `HEXDIG` includes A-F and a-f.
+1. `double-quoted` strings follow JSON in {{RFC8259}}.
+   `single-quoted` strings follow an analogous pattern.
+2. An `element-index` is an integer (in base 10).
+3. The syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.
```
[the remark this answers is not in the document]
goessner: Glyn already suggested to phrase instead: integer base-10

draft-ietf-jsonpath-base.md
cabo: Well, slight massaging required to support both single-quoted and double-quoted strings.
I think we should leave this to a followup PR.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +585,241 @@ of node.
...
-node that has a member with that name.
-It selects no nodes from a node which is not a object.
+The `index-selector` applied with a `quoted-member-name` to an object selects the corresponding value from it, if and only if that object has a member with that name.
+Nothing is selected from a value which is not a object.
 
+Array indexing via `element-index` is a way of selecting a particular array element using a zero-based index.
+For example, selector `[0]` selects the first and selector `[4]` the fifth element of a sufficiently filled array.
 
+A negative `element-index` counts from the array end.
+For example, selector `[-1]` selects the last and selector `[-2]` selects the last but one element of an array with at least two elements.
```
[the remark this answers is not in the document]
cabo: penultimate

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: Right, but that's in point 2.  We still highlight "octal-like" integers in point 3.  Is there a reason we want to call attention to octal over other common number bases?

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +585,241 @@ of node.
...
-              "/" /          ; /    solidus         U+002F
-              "\" /          ; \    reverse solidus U+005C
-              "b" /          ; b    backspace       U+0008
-              "f" /          ; f    form feed       U+000C
-              "n" /          ; n    line feed       U+000A
-              "r" /          ; r    carriage return U+000D
-              "t" /          ; t    tab             U+0009
-              "u" 4HEXDIG )  ; uXXXX                U+XXXX
+~~~~ abnf
+quoted-member-name  =  %x22 *double-quoted %x22 /       ; "string"
```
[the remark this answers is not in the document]
gregsdennis: Suggested in meeting to use `string-literal`

draft-ietf-jsonpath-base.md
cabo: Resolved as per interim (needed a bit more surgery than I had thought, though).

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: Resolved as per interim.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: Please check as now updated.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: We didn't discuss this, so I left it, but "penultimate" really is the way to go.

draft-ietf-jsonpath-base.md
cabo: We didn't discuss this, so I left it for a future PR

draft-ietf-jsonpath-base.md
```diff
@@ -902,8 +957,135 @@ size doesn't
...
+                             index-selector       /  ; ..[<index>]
+                             index-wild-selector  /  ; ..[*]
+                             "*"                     ; ..*
+                           )
+~~~~
+
+#### Semantics
+{: unnumbered}
+
+The `descendant-selector` is inspired by ECMAScript for XML (E4X). It selects object members and array elements in a recursive descent process.
```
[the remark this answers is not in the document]
cabo: I didn't see this yet when I made a different change.  Please check...

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
 
-The child selects the value corresponding to the member name from any object
-node that has a member with that name.
-It selects no nodes from a node which is not a object.
+The `index-selector` applied with a `quoted-member-name` to an object
+selects the node of the corresponding member value from it, if and only if that object has a member with that name.
+Nothing is selected from a value which is not a object.
 
+Array indexing via `element-index` is a way of selecting a particular array element using a zero-based index.
+For example, selector `[0]` selects the first and selector `[4]` the fifth element of a sufficiently filled array.
```
glyn, line 775:
```suggestion
For example, selector `[0]` selects the first and selector `[4]` the fifth element of a sufficiently long array.
```

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
+index-wild-selector    = "[" "*" "]"  ;  asterisk enclosed by brackets
 ~~~~
 
-An array index is an integer (in base 10).
+#### Semantics
+{: unnumbered}
 
-~~~~ abnf
-array-index = integer
+An index selector combined with a single asterisk is a wild card. It
```
glyn, line 795: Same issue as before - don't put syntax in the semantics section.
```suggestion
An `index-wild-selector`
```

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
+selects all member values of an object as well as all elements of an
+array.
```
glyn, line 797: Should talk about selecting nodes rather than values.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
+step           = int       ; default: 1
+
+ws             = *( %x20 / ; Space
+                    %x09 / ; Horizontal tab
+                    %x0A / ; Line feed or New line
+                    %x0D ) ; Carriage return
 ~~~~
 
-Note: the array slices `:` and `::` are both syntactically valid, as are `:2:2`, `2::2`, and `2:4:`.
+The `slice-selector` consists of three optional decimal integers separated by colons.
```
glyn, line 827: @gregsdennis pointed out before that omitting the three integers always leaves behind _two_ separating colons when only one colon is valid. The ABNF is correct, but the description here contradicts it.

draft-ietf-jsonpath-base.md
```diff
@@ -902,11 +965,173 @@ size doesn't
...
+                             index-wild-selector  /  ; ..[*]
+                             "*"                     ; ..*
+                           )
+~~~~
+
+#### Semantics
+{: unnumbered}
+
+The `descendant-selector` is inspired by ECMAScript for XML (E4X). It
+selects the nodes of object member values and array elements in a recursive descent process.
```
glyn, line 989: The descendant selector should select all primitive values which are part of the subtree of the input node. For example, `..*` applied to the argument `true` should produce `[true]` even though there are no object members or array elements in the argument.

The descendant selector should also select _all_ object members and _all_ array elements (the current wording could permit some to be omitted).
```suggestion
It selects the node and all its descendants..
```

draft-ietf-jsonpath-base.md
```diff
@@ -902,11 +965,173 @@ size doesn't
...
+~~~~
+
+> Task (T1): This, besides slice-index, is currently one of only two
+> places in the document that mentions whitespace.  Whitespace needs
+> to be handled throughout the ABNF syntax.  Room Consensus at the
+> 2021-06-15 interim was that JSONPath generally is generous with
+> allowing insignificant whitespace throughout.  Minimizing the impact
+> of the many whitespace insertion points by choosing a rule name such
+> as "S" was mentioned.  Some conventions will probably help with
+> minimizing the number of places where S needs to be inserted.
```
glyn, line 1013: For the record, "ws" is far more memorable than "S".

draft-ietf-jsonpath-base.md
```diff
@@ -902,11 +965,173 @@ size doesn't
...
+               "<"  / ">"  /                        ; operators
+               "<=" / ">="
+regex-op     = "~="                                 ; RegEx match
+in-op        = " in "                               ; in operator
+comparable   = number / string-literal /            ; primitive ...
+               true / false / null /                ; values only
+               rel-path-val /                       ; descendant value
+               json-path                            ; any value
+
+rel-path-val = "@" *(dot-selector / index-selector)
```
glyn, line 1061: We also need absolute path values such as `$.foo`.

glyn (changes requested):
I left some comments or suggestions. I am now going on vacation, so apologies if I don't participate further for a couple of weeks.

draft-ietf-jsonpath-base.md
```diff
@@ -44,17 +44,6 @@ author:
...
-  ins: M. Mikulicic
-  name: Marko Mikulicic
-  org: InfluxData, Inc.
-  street: ''
-  city: Pisa
-  region: ''
-  code: ''
-  country: IT
-  phone: ''
-  email: mmikulicic@gmail.com
```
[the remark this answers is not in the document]
cabo: I put him back to the contributors section (sorry about that).

draft-ietf-jsonpath-base.md
```diff
@@ -157,7 +160,10 @@ For example, the Unicode PLACE OF INTEREST SIGN (U+2318) would be defined
 in ABNF as `%x2318`.
 
 The terminology of {{-json}} applies except where clarified below.
-Definitions for "Object", "Array", and "Number" remain unchanged.
+The terms "primitive" and "structured" are used to group
+the types as in {{Section 1 of -json}}.
+Definitions for "Object", "Array", "Number", and "String" remain
+unchanged.
```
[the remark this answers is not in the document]
cabo: That's my German (nouns with caps, adjectives with lower case).  Fixed.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
-
-More general child names, such as the empty string, are supported by "Union
-Child" ({{unionchild}}{: format="default"}).
-
-Note that the `dot-child-name` rule follows the philosophy of JSON strings and is
-allowed to contain bit sequences that cannot encode Unicode characters (a
-single unpaired UTF-16 surrogate, for example).
-The behaviour of an implementation is undefined for child names which do
-not encode Unicode characters.
+Every valid JSONPath query MUST begin with the root selector named `$`.
```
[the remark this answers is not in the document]
cabo: Indeed.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
 
+~~~~ abnf
+root-selector  = "$"
+~~~~
 
 #### Semantics
-{: numbered="false" toc="exclude"}
+{: unnumbered}
+The Argument — the root JSON value — is anonymous by nature.
+By getting assigned the universal name `$` it becomes the root node.
```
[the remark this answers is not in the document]
cabo: Changed this -- a bit redundant now.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +583,257 @@ of node.
...
 ~~~~
 
-
 #### Semantics
+{: unnumbered}
 
-A union selects any node which is selected by at least one of the union selectors
-and selects the concatenation of the
-lists (in the order of the selectors) of nodes selected by the union elements.<!--  TODO: define whether duplicates are kept or removed.  -->
+A dot selector combined with a single asterisk is a wild card.
```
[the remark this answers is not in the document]
cabo: Can't leave the "it" floating; slightly rephrased.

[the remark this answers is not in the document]
cabo: Oh, you have better text below.

draft-ietf-jsonpath-base.md
cabo: Fixed.

draft-ietf-jsonpath-base.md
cabo: Wait, the step is optional in the syntax, too.

draft-ietf-jsonpath-base.md
cabo: To do: Define "descendants" (making sure that member values are, but member names aren't).

draft-ietf-jsonpath-base.md
cabo: If you are stuck in an ABNF littered with them, S is sufficiently memorable...

(WS really should be for mandatory whitespace, OWS is optional whitespace, but that is too long here.)

draft-ietf-jsonpath-base.md
cabo: Todo for next PR...

draft-ietf-jsonpath-base.md
```diff
@@ -555,7 +581,7 @@ of the input node.
 The result is a list of three nodes: `{"b":0}`, `{"b":1}`, and `{"c":2}`.
 
 Finally, `.b` selects from any input node of type object with a member name
-`b` and selects the value of the input node corresponding to that name.
+`b` and selects the node of the member value of the input node corresponding to that name.
```
sthagen, line 584: I have a hard time understanding what a node of a value might be. This seems to be dimensionally wrong. Maybe there are too many objects in that sentence?

draft-ietf-jsonpath-base.md
sthagen: I never got that color thing people associate with space - maybe I should buy me a printer and do some archaeology? Just  `S` or `SP` should do nicely.

sthagen (approved):
Looks like a good next step on our path.

draft-ietf-jsonpath-base.md
glyn: The ABNF syntax is correct (last time I looked!). It's just the above sentence which is misleading.

glyn (approved):
As I'm going away, approving so I don't block progress. I think @cabo has responded to most if not all my comments.

draft-ietf-jsonpath-base.md
goessner: I am fine with "S" (minimalistic :). Is there a convention to write constants in capitals?

draft-ietf-jsonpath-base.md
goessner: this is a `json-path` then.

draft-ietf-jsonpath-base.md
```diff
@@ -569,207 +595,257 @@ of node.
...
+                    %x0D ) ; Carriage return
 ~~~~
 
-Note: the array slices `:` and `::` are both syntactically valid, as are `:2:2`, `2::2`, and `2:4:`.
+The `slice-selector` consists of three optional decimal integers separated by colons.
+
+#### Semantics
+{: unnumbered}
+
+The `slice-selector` was inspired by the slice operator of ECMAScript 4 (ES4), which was deprecated in 2014 and later adopted by Python.
```
goessner, line 845: Seems to be misleading (I know my wording). It was surely adopted by Python before 2014. So skip `later`.

draft-ietf-jsonpath-base.md
```diff
@@ -902,11 +977,175 @@ size doesn't
...
+logical-and-expr = comp-expr *["&&" comp-expr]
+
+comp-expr    = (rel-path-val) [(comp-op comparable /  ; comparison
+                            regex-op regex     /  ; RegEx test
+                            in-op container )]    ; containment test
+comp-op      = "==" / "!=" /                        ; comparison ...
+               "<"  / ">"  /                        ; operators
+               "<=" / ">="
+regex-op     = "~="                                 ; RegEx match
+in-op        = " in "                               ; in operator
```
goessner, line 1067: For the next PR:

Thinking a bit more about `in-op` leads me to the conclusion, that this operator makes sense only for arrays.
`'key' in $.obj` is equivalent to `$.obj.key` and `'key' in <literal-object>` seems to be useless. I see no other rational here. Someone else ?

goessner (approved):

draft-ietf-jsonpath-base.md
gregsdennis: I agree that this is rather verbose now.  Maybe

```suggestion
Finally, `.b` selects from any input node of type object any node corresponding to
a member named `b`.
```

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: You just added "As in JSON numbers..."

>3. As in JSON numbers, the syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.

That doesn't address my concern at all.  My question is about the callout of _octal_ and exclusion of other number bases such as binary and hexadecimal.

draft-ietf-jsonpath-base.md
```diff
@@ -902,11 +977,175 @@ size doesn't
...
+                             index-wild-selector  /  ; ..[*]
+                             "*"                     ; ..*
+                           )
+~~~~
+
+#### Semantics
+{: unnumbered}
+
+The `descendant-selector` is inspired by ECMAScript for XML (E4X). It
+It selects the node and all its descendants..
```
gregsdennis, line 1001: Repitition of "It" and extra period (ironically).

```suggestion
The `descendant-selector` is inspired by ECMAScript for XML (E4X).
It selects the node and all its descendants.
```

draft-ietf-jsonpath-base.md
gregsdennis: Do we need to make concession for multiple whitespace chars (e.g. `*ws` or `*S`), or is that implicitly covered?

gregsdennis (approved):
Added some comments that I'm happy to have resolved in a follow-up non-functional edit PR.

draft-ietf-jsonpath-base.md
cabo: That is a simple solution that we can refine later.  Put it in.

draft-ietf-jsonpath-base.md
cabo: RFC 8610 defines this as:

     S = *WS
     WS = SP / NL
     SP = %x20
     NL = COMMENT / CRLF
     COMMENT = ";" *PCHAR CRLF
     PCHAR = %x20-7E / %x80-10FFFD
     CRLF = %x0A / %x0D.0A

I'm not sure we need to support comments in JSONPath (that is a serious question though), but if you remove that, pesky details like isolated CRs are fixed, too (and then CDDL simply disallows HTs, which may not be popular here).

Next PR...

cabo: Oh, and I'm not sure these can all be called constants, but there is indeed a tendency in ABNF usage to give token-level (scanner) productions such as DIGIT, ALPHA, WS, ... upper case names.

draft-ietf-jsonpath-base.md
cabo: there is still the "enum" use case, as in `$.service_level in ["silver", "gold"]`, which is an abbreviation for `$.service_level == "silver" || $.service_level == "gold"`.

draft-ietf-jsonpath-base.md
cabo: A quick search finds it in Python in 2009, but I didn't check further.
Changed to ", and that of Python".

BTW, do we support negative steps?

draft-ietf-jsonpath-base.md
cabo: Fixed.

draft-ietf-jsonpath-base.md
goessner: well ... this is the `in <array>` use case.

draft-ietf-jsonpath-base.md
gregsdennis: The current test suite and the comparison matrix _does_ have negative step cases.

glyn: @cabo Negative steps are also supported in the spec. See for example Table 4.

cabo: Thanks.  It just occurred to me as the semantics section needs to gain more content.

PR #104, opened 2021-07-06
Close discussion D1

glyn:
Here's a chance to actually have the discussion if anyone wants to.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -463,9 +463,6 @@ raised during application of the query to a value.
 query, e.g., because of resource depletion, but this is not modeled in
 the present specification.)
 
-> Discussion (D1): are we ready for the error model defined here, i.e. one
-> that does not have runtime errors, but only compile time errors?
-
 ## Processing Model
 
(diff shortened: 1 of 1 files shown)


cabo:
Well, I'm certainly ready for this model.

gregsdennis:
I'm good with the approach described in the text.  As long as a string can be parsed as a Path, it should evaluate.  I can't see a case where processing a valid path against a value would produce an error; merely empty result sets.

This means that the only errors should be syntax errors, correct?

glyn:
> This means that the only errors should be syntax errors, correct?

I think so.

glyn:
May I encourage reviewers who are in favour of this PR to approve it?

gregsdennis:
I was awaiting a bit more discussion.

glyn:
Ok, I've requested a review from Stefan.

goessner:
> are we ready for the error model defined here, i.e. one
> that does not have runtime errors, but only compile time errors?

Yes absolutely ... I like that clean concept a lot.

goessner (approved):

gregsdennis (approved):

PR #105, opened 2021-07-06
Nail down combine1

glyn:


gregsdennis (approved):
I agree with this. Concatenation seems to always be the appropriate action here.

PR #106, opened 2021-07-06
Close discussion D4

glyn:
Here's a chance to actually have the discussion if anyone wants to.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -555,9 +555,6 @@ the combine1 algorithm to form the result nodelist of the selector.
 For each node in the list, the selector selects zero or more nodes,
 each of which is a descendant of the node or the node itself.
 
-> Discussion (D4): Is that a common requirement?  Or can a selector also go
-> up, or to the query argument?
-
 <!-- To do: Define "descendants" (making sure that member values are, but member names aren't). -->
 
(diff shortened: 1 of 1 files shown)


cabo:
One related question is whether the selector can use $ internally (still returning nodes within the current node).

glyn:
> One related question is whether the selector can use $ internally (still returning nodes within the current node).

I am in favour of supporting `$` inside filter expressions as a way to access the root node of the argument, but I don't think that should block this PR as the context of the discussion seems to be which nodes are selected. As you point out, using `$` in a filter expression doesn't select the root node - it selects nodes within the current node as usual.

gregsdennis:
I agree.  Paths like `$.foo[?(@.bar < $.baz)]` should be allowed, but wouldn't select anything that is not a descendent of `$.foo`.  Here the `$.baz` provides a basis for comparison used to select the current node `@` in the iteration.

I see no value (for an initial release anyway) in allowing selection of ancestor nodes.  If a use case and syntax for it are later defined, I'm happy to release that as a follow-up version.

goessner:
yes ... using `$` in a logical expression is valid and does no harm.

goessner (approved):

gregsdennis:
@goessner yes, but the issue at hand here is whether there is a mechanism that can select ancestor or cousin nodes.  I think we're in agreement that that there is no such mechanism.

cabo (approved):
Indeed, we seem to be in agreement.

PR #107, opened 2021-07-06
Clarify editors & authors

glyn:


cabo:
I think Stefan should be first author, and I wouldn't mind to be last.

glyn:
> I think Stefan should be first author, and I wouldn't mind to be last.

Ok, thanks. Updated.

cabo (approved):
Thank you.

PR #108, opened 2021-07-06
Fix invalid link

glyn:
Without this `make` fails with:

draft-ietf-jsonpath-base.xml(0): Error: IDREF attribute target references an unknown ID "_json", at None


cabo:
Where are you getting this?
You may want to do a `gem update` when you see an error like this.

glyn:
> Where are you getting this?

When I issue `make` locally.

> You may want to do a `gem update` when you see an error like this.

I did and the problem went away. Thanks!

Issue #110, opened 2021-07-23
Support arbitrarily structured filter predicates


glyn:
The filter syntax currently requires predicates to be expressed in (something close to) disjunctive normal form (OR of ANDs), but this is likely to be inconvenient when the intention of the filter is expressed in a different way (e.g. as an AND or ORs or as a nested predicate with ANDs and ORs at various levels).

The JSONPath comparison project doesn't yet test these more complex predicates, although some, possibly most, implementations allow them.

The standard could allow arbitrarily structured predicates via syntax along these lines:
```
filter-selector ::= "[?(" boolean-expr ")]"

boolean-expr ::= logical-and-expr |
                 logical-and-expr "||" boolean-expr

logical-and-expr ::= basic-filter |
                     basic-filter "&&" logical-and-expr

basic-filter ::= "!" basic filter |
                 "(" boolean-expr ")" |
                 comp-expr
```

For example, this would allow filters such as:
```
?(@.b && (@.c || @.d))
```
and
```
?(@.b && (@.c || @.d && @.e))
```
(where the semantics could, in line with mathematical convention and many programming languages, define `&&` to bind more tightly than `||`).

Note: the comparison project's [Proposal A](https://github.com/cburgmer/json-path-comparison/tree/master/proposals/Proposal_A) restricts the syntax so that brackets are _required_ instead of making `&&` bind more tightly than `||`. This is a clever approach but is probably not suitable for the standard because it is rather unconventional.

cabo:
I don't have a problem with requiring parentheses instead of creating an elaborate precedence hierarchy (or at least partially instead of).

I also don't have a problem with arbitrary nesting here.

gregsdennis:
> where the semantics could, in line with mathematical convention and many programming languages, define `&&` to bind more tightly than `||`

For clarity, this is because `&&` is a multiplicative operation whereas `||` is additive.  Standard order of operations then apply.

I don't know any programming language where this is reversed.  I'm not opposed to requiring `()`, however.

goessner:
If the fact that `&&`binds more tightly than `||` must be made explicite in the ABNF, it should be done. Using pairs of parentheses should be arbitrary.

glyn:
Agreed. Is there a preferred way to specify operator precedence in ABNF? The examples I've seen use a textual ordered list which won't help parser generators...

gregsdennis:
[...]

PR #111, opened 2021-07-30
Support arbitrary predicate structure

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/110

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1040,10 +1040,10 @@ boolean-expr     = logical-or-expr
 logical-or-expr  = logical-and-expr *("||" logical-and-expr)
                                                       ; disjunction
                                                       ; binds less tightly than conjunction
-logical-and-expr = basic-filter *("&&" basic-filter)  ; conjunction
+logical-and-expr = basic-expr *("&&" basic-expr)      ; conjunction
                                                       ; binds more tightly than disjunction
 
-basic-filter = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
+basic-expr   = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
 existence-expr = [neg-op] path                        ; path existence or non-existence
 paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
 neg-op       = "!"                                    ; not operator
@@ -1051,7 +1051,7 @@ neg-op       = "!"                                    ; not operator
 comp-expr    = path (comp-op comparable /             ; comparison
                      regex-op regex     /             ; RegEx test
                      in-op container)                 ; containment test
-path         = rel-path-val / json-path
+path         = rel-path / json-path
 comp-op      = "==" / "!=" /                          ; comparison ...
                "<"  / ">"  /                          ; operators
                "<=" / ">="
@@ -1059,10 +1059,10 @@ regex-op     = "~="                                   ; RegEx match
 in-op        = " in "                                 ; in operator
 comparable   = number / string-literal /              ; primitive ...
                true / false / null /                  ; values only
-               rel-path-val /                         ; relative path value
+               rel-path /                             ; relative path value
                json-path                              ; any value
 
-rel-path-val = "@" *(dot-selector / index-selector)
+rel-path     = "@" *(dot-selector / index-selector)
 
 container = <TO BE DEFINED>
 regex = <TO BE DEFINED>
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
 
 
 ~~~~ abnf
+boolean-expr     = logical-or-expr / logical-and-expr
+logical-or-expr  = logical-and-expr "||" boolean-expr ; disjunction
+                                                      ; binds less tightly than conjunction
+logical-and-expr = basic-filter /
+                   basic-filter "&&" logical-and-expr ; conjunction
+                                                      ; binds more tightly than disjunction
 
-
-boolean-expr = logical-expr
-logical-expr = ([neg-op] primary-expr) / logical-or-expr
-neg-op       = "!"                                  ; not operator
-primary-expr = "(" logical-or-expr ")"
-logical-or-expr = logical-and-expr *["||" logical-and-expr]
-logical-and-expr = comp-expr *["&&" comp-expr]
+basic-filter = comp-expr / paren-expr / ([neg-op] paren-expr)
+paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
+neg-op       = "!"                                    ; not operator
 
 comp-expr    = (rel-path-val /
-                json-path) [(comp-op comparable / ; comparison
-                            regex-op regex     /  ; RegEx test
-                            in-op container )]    ; containment test
-comp-op      = "==" / "!=" /                        ; comparison ...
-               "<"  / ">"  /                        ; operators
+                json-path) [(comp-op comparable /     ; comparison
+                            regex-op regex     /      ; RegEx test
+                            in-op container )]        ; containment test
+comp-op      = "==" / "!=" /                          ; comparison ...
+               "<"  / ">"  /                          ; operators
                "<=" / ">="
-regex-op     = "~="                                 ; RegEx match
-in-op        = " in "                               ; in operator
-comparable   = number / string-literal /            ; primitive ...
-               true / false / null /                ; values only
-               rel-path-val /                       ; descendant value
-               json-path                            ; any value
+regex-op     = "~="                                   ; RegEx match
+in-op        = " in "                                 ; in operator
+comparable   = number / string-literal /              ; primitive ...
+               true / false / null /                  ; values only
+               rel-path-val /                         ; descendant value
+               json-path                              ; any value
 
 rel-path-val = "@" *(dot-selector / index-selector)
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1043,14 +1043,15 @@ logical-or-expr  = logical-and-expr *("||" logical-and-expr)
 logical-and-expr = basic-filter *("&&" basic-filter)  ; conjunction
                                                       ; binds more tightly than disjunction
 
-basic-filter = comp-expr / paren-expr / ([neg-op] paren-expr)
+basic-filter = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
+existence-expr = [neg-op] path                        ; path existence or non-existence
 paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
 neg-op       = "!"                                    ; not operator
 
-comp-expr    = (rel-path-val /
-                json-path) [(comp-op comparable /     ; comparison
-                            regex-op regex     /      ; RegEx test
-                            in-op container )]        ; containment test
+comp-expr    = path (comp-op comparable /             ; comparison
+                     regex-op regex     /             ; RegEx test
+                     in-op container)                 ; containment test
+path         = rel-path-val / json-path
 comp-op      = "==" / "!=" /                          ; comparison ...
                "<"  / ">"  /                          ; operators
                "<=" / ">="
@@ -1058,7 +1059,7 @@ regex-op     = "~="                                   ; RegEx match
 in-op        = " in "                                 ; in operator
 comparable   = number / string-literal /              ; primitive ...
                true / false / null /                  ; values only
-               rel-path-val /                         ; descendant value
+               rel-path-val /                         ; relative path value
                json-path                              ; any value
 
 rel-path-val = "@" *(dot-selector / index-selector)
@@ -1077,7 +1078,6 @@ Notes:
 * A member or element value by itself is *falsy* only, if it does not exist. Otherwise it is *truthy*, resulting in its value. To be more specific explicit comparisons are necessary. This existence test — as an exception of the general rule — also works with complex values.
 * Regular expression tests can be applied to `string` values only.
 * Containment tests work with arrays and objects.
-* Explicit boolean type conversion is done by the not operator `neg-op`.
 * The behaviour of operators is consistent with the 'C'-family of programming languages.
 <!-- need to clarify -->
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1036,11 +1036,11 @@ The current item is selected if and only if the result is `true`.
 
 
 ~~~~ abnf
-boolean-expr     = logical-or-expr / logical-and-expr
-logical-or-expr  = logical-and-expr "||" boolean-expr ; disjunction
+boolean-expr     = logical-or-expr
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
 
 
 ~~~~ abnf
+boolean-expr     = logical-or-expr / logical-and-expr
+logical-or-expr  = logical-and-expr "||" boolean-expr ; disjunction
+                                                      ; binds less tightly than conjunction
+logical-and-expr = basic-filter /
+                   basic-filter "&&" logical-and-expr ; conjunction
+                                                      ; binds more tightly than disjunction
```
goessner, line 1045: this is very clean and seems to manage binding preference for `and` ... great.

draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
...
-
-boolean-expr = logical-expr
-logical-expr = ([neg-op] primary-expr) / logical-or-expr
-neg-op       = "!"                                  ; not operator
-primary-expr = "(" logical-or-expr ")"
-logical-or-expr = logical-and-expr *["||" logical-and-expr]
-logical-and-expr = comp-expr *["&&" comp-expr]
+basic-filter = comp-expr / paren-expr / ([neg-op] paren-expr)
+paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
+neg-op       = "!"                                    ; not operator
```
goessner, line 1049: We could select
```json
[?(@.audio && !@.video)]
```
using implicit existance test. Requiring parentheses here, i.e. `!(@.video)` would be tedious.

This brings up another issue ... see below ...

draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
...
-comparable   = number / string-literal /            ; primitive ...
-               true / false / null /                ; values only
-               rel-path-val /                       ; descendant value
-               json-path                            ; any value
+regex-op     = "~="                                   ; RegEx match
+in-op        = " in "                                 ; in operator
+comparable   = number / string-literal /              ; primitive ...
+               true / false / null /                  ; values only
+               rel-path-val /                         ; descendant value
+               json-path                              ; any value
```
goessner, line 1063: `rel-path-val` and `json-path` ...
* are in fact valid ... hmm ... `defined-expr` or something ... indicating existence.
* naming should be harmonized ... `rel-path-val` =>  `rel-path` or `rel-json-path`

draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
...
+                                                      ; binds more tightly than disjunction
 
-
-boolean-expr = logical-expr
-logical-expr = ([neg-op] primary-expr) / logical-or-expr
-neg-op       = "!"                                  ; not operator
-primary-expr = "(" logical-or-expr ")"
-logical-or-expr = logical-and-expr *["||" logical-and-expr]
-logical-and-expr = comp-expr *["&&" comp-expr]
+basic-filter = comp-expr / paren-expr / ([neg-op] paren-expr)
```
goessner, line 1046: `basic-filter` breaks the naming rule (not communicated though) of having `xxx-expr` and `yyy-op` exclusively here.

`basic-expr` or something would fit better, I think.

draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
...
-               rel-path-val /                       ; descendant value
-               json-path                            ; any value
+regex-op     = "~="                                   ; RegEx match
+in-op        = " in "                                 ; in operator
+comparable   = number / string-literal /              ; primitive ...
+               true / false / null /                  ; values only
+               rel-path-val /                         ; descendant value
+               json-path                              ; any value
 
 rel-path-val = "@" *(dot-selector / index-selector)
```
goessner, line 1065: significant improvement overall ... thanks

draft-ietf-jsonpath-base.md
glyn: Good point. I'd like to avoid confusing possibilities such as:
```
[?(!@.foo == bar)]
```
but I agree that it shouldn't be necessary to wrap existence terms in parentheses merely in order to negate them.

draft-ietf-jsonpath-base.md
glyn: Thanks for explaining the naming rule, which I hadn't noticed. I'll do some renaming.

draft-ietf-jsonpath-base.md
glyn: Let me mull that over and try to improve matters.

draft-ietf-jsonpath-base.md
```diff
@@ -1036,28 +1036,30 @@ The current item is selected if and only if the result is `true`.
...
 
-
-boolean-expr = logical-expr
-logical-expr = ([neg-op] primary-expr) / logical-or-expr
-neg-op       = "!"                                  ; not operator
-primary-expr = "(" logical-or-expr ")"
-logical-or-expr = logical-and-expr *["||" logical-and-expr]
-logical-and-expr = comp-expr *["&&" comp-expr]
+basic-filter = comp-expr / paren-expr / ([neg-op] paren-expr)
+paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
```
gregsdennis, line 1047: Eventually, we'll want other-type expressions here, such as `4*(5+6)`.

gregsdennis (changes requested):
Agree with comments posted already, and added one of my own.

draft-ietf-jsonpath-base.md
goessner: Hmm ... if we agree to existence tests and assume 

```
{ "zero": 0, "five": 5 }
``` 

then unlike in C-like languages.

`@.five` evaluates to `true`, but consistently `!@.five` to `false` and `!!@.five` to true ... and ...
`@.zero` evaluates to `true`, `!@.zero` to `false` and `!!@.zero` to true.

Testing for values needs always to be explicit, as in `@.five == 5`, `@.five > 0`, `@.five != 6` or `@.zero != 0`.

gregsdennis: So then how would `!@.foo == false` be interpreted?

goessner: if `@.foo`exists then `!@.foo` results in `false` and `!@.foo == false` will give `true` ... not to be confused with  `!(@.foo == false)` which is `true` for all values of `foo` (including `undefined`) except of  `false` ...

gregsdennis: So in summary, if `@.foo` holds a value of `false`, then `!@.foo == false` yields true, and `!(@.foo == false)` yields false.

I think this subtlety could bear some documentation at the very least.

draft-ietf-jsonpath-base.md
cabo: Why are we moving away from the standard way of expressing grammars for languages with operator precedences?  Obviously, right recursion works, but we do have *(...) in ABNF and could use it.
(Right recursion leads to a binary tree that leans to the right, so a + b + c parses as a + (b + c).
Not a problem as long as operators are [associative](https://en.wikipedia.org/wiki/Operator_associativity).)
See the ECMAScript grammars for the usual way to do this.

gregsdennis: This was asked earlier and received little response.

draft-ietf-jsonpath-base.md
goessner: So we have the alternatives ...

1. `@.foo` always indicates existence (`true / false`) if used standalone. Drawback: inconsistency with potential value.
2.  `@.foo` results in `false` if it doesn't exist, in its value otherwise. Drawback: What's about falsy values `false, 0, null`?

We need to handle the Javascript value equivalence `undefined`, which is not a valid JSON value.

Best I can see yet, is Greg's proposal slightly modified: 

For comparisons like `@.foo > 5`, always take the node value, except it does not exist, than ... hmm ... take

* `false` or
* `null`

and let the implementation do implicite type conversion.  We cannot forbid existence operations within comparative expressions, if we don't want do error handling (which is not syntactically based).

General drawback: We cannot tell, if  `!@.foo == true` is `true` because of its value `false` or inexistence. The same effect comes with  `@.foo != true`.

General result of this discussion: We cannot reliably test whether a node exists or not.

draft-ietf-jsonpath-base.md
cabo: Where was what asked?

gregsdennis: @Glyn asked if there was a standard easy for ANBF to support operator precedence.

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/110#issuecomment-887390352

You replied.

draft-ietf-jsonpath-base.md
glyn: I don't intend to include that in this PR, thanks.

glyn:
Thanks for the various comments. I've been busy with other things, but intend to come back before too long, absorb those comments, and push up some changes.

cabo:
On 2. Aug 2021, at 20:53, Greg Dennis ***@***.***> wrote:
> 
> You replied.
> 
Yes.  How is that "little response”?

I’m still trying to parse what you mean by "This was asked earlier and received little response.”.

What are you trying to say?

Grüße, Carsten

draft-ietf-jsonpath-base.md
glyn: > Why are we moving away from the standard way of expressing grammars for languages with operator precedences? 

Simply ignorance on my part.
> Obviously, right recursion works, but we do have *(...) in ABNF and could use it.
> (Right recursion leads to a binary tree that leans to the right, so a + b + c parses as a + (b + c).
> Not a problem as long as operators are [associative](https://en.wikipedia.org/wiki/Operator_associativity).)
> See the ECMAScript grammars for the usual way to do this.

I can't find an ABNF ECMAScript grammar. [This grammar](https://tc39.es/ecma262/#sec-multiplicative-operators) uses recursion, e.g.:
```
 AdditiveExpression :
AdditiveExpression + MultiplicativeExpression
AdditiveExpression - MultiplicativeExpression
```
@cabo Please could you provide a more specific reference to an ECMAScript grammar which demonstrates the "standard way" you are thinking of?

draft-ietf-jsonpath-base.md
glyn: I'm going to push a change which renders `!@.foo == false` syntactically invalid. See what you make of it.

glyn: In fact the changes I just pushed make existence and non-existence tests incapable of being used in comparisons, which I hope avoids some of the issues above.

draft-ietf-jsonpath-base.md
glyn: I went for `existence-expr` which now includes optional negation and `rel-path` as it's shorter.

draft-ietf-jsonpath-base.md
```diff
@@ -1075,7 +1078,6 @@ Notes:
 * A member or element value by itself is *falsy* only, if it does not exist. Otherwise it is *truthy*, resulting in its value. To be more specific explicit comparisons are necessary. This existence test — as an exception of the general rule — also works with complex values.
 * Regular expression tests can be applied to `string` values only.
 * Containment tests work with arrays and objects.
-* Explicit boolean type conversion is done by the not operator `neg-op`.
```
glyn, line 1078: I think this type conversion statement is no longer needed since `!` now applies only to existence terms, which give boolean values, and parenthesised boolean expressions which again give boolean values.

draft-ietf-jsonpath-base.md
glyn: @cabo I tried using repetition instead of right recursion. Please see what you make of it.

cabo: Indeed, I must have misremembered the way TC39 writes their grammars, which apparently has morphed to a simple BNF (without repetition) plus weird two-level features (not quite Algol68 though :-).

Because different parser generators react in different ways to (immediate) recursion, I try to avoid that in favor of repetition.  I just looked at the Python grammar for a potential example, but that is a weird mix of the different styles.

31c1a7f is exactly what I had in mind.  Regular and non-surprising.  Thank you.

glyn:
Ok, so I pushed up some changes which attempt to address all the issues raised. Please see the individual commit logs for more detailed comments.

draft-ietf-jsonpath-base.md
glyn: Phew, thanks! :sweat_smile:

cabo:
Maybe we want to add an operator precedence table [1] into the text and try to make sure the grammar is somewhat congruence to that table.

[1]: https://en.cppreference.com/w/c/language/operator_precedence

draft-ietf-jsonpath-base.md
glyn: I went for `basic-expr`.

glyn:
> Maybe we want to add an [operator precedence table](https://en.cppreference.com/w/c/language/operator_precedence) into the text and try to make sure the grammar is somewhat congruence to that table.

Good idea, but I'd prefer to defer as I've got enough to deal with in this PR! It will also be even more valuable if we end up supporting arithmetic expressions...

draft-ietf-jsonpath-base.md
```diff
@@ -1036,30 +1036,33 @@ The current item is selected if and only if the result is `true`.
...
+                                                      ; binds less tightly than conjunction
+logical-and-expr = basic-expr *("&&" basic-expr)      ; conjunction
+                                                      ; binds more tightly than disjunction
+
+basic-expr   = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
+existence-expr = [neg-op] path                        ; path existence or non-existence
+paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
+neg-op       = "!"                                    ; not operator
+
+comp-expr    = path (comp-op comparable /             ; comparison
```
glyn, line 1051: @goessner I'd like to support `comparable comp-op path` as well since it seems arbitrary to require the path to be on the left hand side of the comparison operator. Something for a future PR, perhaps?

glyn: I raised https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/112 to cover this.

cabo:
Having an operator precedence table will help us understand what we are trying to do here.
But I believe in incremental improvement, so completing this PR and then re-approaching it from the angle of the precedence table works for me.

goessner (approved):

glyn:
@gregsdennis Please approve if your concerns are now addressed.

glyn:
I've started the next batch of predicate improvements in draft PR https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/115 which is based on this PR.

gregsdennis (approved):

Issue #114, opened 2021-08-04
regex comparison operator should be =~


glyn:
Perl introduced `=~` as `string =~ regexp` (where regexp actually could be a substitute operator etc.).
Ruby generalized this to `regexp =~ string` and `string =~ regexp`.
(Java, JavaScript, Python, PowerShell don't have an operator at all; this is all handled using method calls.)

So if we want to have syntax for regexp matching, we probably want to go with Perl's.

_Originally posted by @cabo in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/112#issuecomment-892642185_

Issue #113, opened 2021-08-04
Operator precedence table


glyn:
Introducing an operator precedence table will help us comunicate what we are trying to achieve precedence-wise in the ABNF syntax.

_Originally posted by @cabo in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/111#issuecomment-892618964_

goessner:
Here is a first version ... taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence)

| Precedence | Operator type | Syntax | Comment |
|:--:|:--:|:--:|:--:|
|  6  | Grouping | `(...)` |  | 
|  5  | Logical NOT | `!` |  | 
|  4  | Less Than<br>Greater Than<br>Less Than Or Equal<br>Greater Than Or Equal | `<`<br>`>`<br>`<=`<br>`>=` | Relation | 
|  3  | Equal To<br>Not Equal To | `==`<br>`!=` | Comparison | 
|  2  | Logical AND | `&&` |  | 
|  1  | Logical OR | `\|\|` |  |

glyn:
Thanks for that @goessner!

I suggest we combine the third and fourth rows and give `==` and `!=` the same precedence as the other comparison operators as I don't think there is a situation where their relative precedence is necessary.

Also, having a separate row for `==` and `!=` gives the impression we want to entertain expressions such as:
```
@.foo < 2 == @.bar
```
and
```
1 < 2 == 2 < 3
```
which I'd prefer to avoid

goessner:
Yes ... so ...

| Precedence | Operator type | Syntax | Comment |
|:--:|:--:|:--:|:--:|
|  5  | Grouping | `(...)` |  | 
|  4  | Logical NOT | `!` |  | 
|  3  | Equal To<br>Not Equal To<br>Less Than<br>Greater Than<br>Less Than Or Equal<br>Greater Than Or Equal | `==`<br>`!=`<br>`<`<br>`>`<br>`<=`<br>`>=` | Comparison | 
|  2  | Logical AND | `&&` |  | 
|  1  | Logical OR | `\|\|` |  |

Issue #112, opened 2021-08-04
Generalise filter comparisons


glyn:
Currently, filter comparison require a path on the left hand side and a constant expression on the right hand side.  This seems overly restrictive. We should generalise it to allow the opposite order as well as path to path comparisons and, for completeness, constant expression to constant expression comparisons.

For example, the following comparisons should be valid:
```
@.foo == 1
1 == @.foo
@.foo == @.bar
1 == 1
```

cabo:
Should `=~` be symmetric, too?

glyn:
Did you mean `~=` (which is what we use in the current draft)?

I think not because its operands have distinct types and it's probably easier to police restrictions and give good diagnostics if we keep `~=` asymmetric. For instance, with the current syntax:
```
@.foo ~= @.bar
```
can fail with an error such as "@.bar needs to be a regular expression literal", which is a bit easier to understand than something like "one side of ~= needs to be a regular expression literal".

Having said that, other generalisations of `~=` may be worth it, such as allowing a string literal (rather than only a path) on the left hand side, for consistency with the above syntax for the other comparison operators, but even that could be give rise to confusion, e.g.:
```
"/.*/" ~= /.*/
```

cabo:
Oops.  It really should be `=~`.

Perl introduced `=~` as `string =~ regexp` (where regexp actually could be a substitute operator etc.).
Ruby generalized this to `regexp =~ string` and `string =~ regexp`.
Java, JavaScript, Python, PowerShell don't have an operator at all; this is all handled using method calls.

So if we want to have syntax for regexp matching, we probably want to go with Perl's.
We also need to discuss whether we need syntax for regexp literals.
But this is maybe leading beyond this github issue.

goessner:
Convinced ... you made the point.

PR #115, opened 2021-08-06
Filter predicate improvements

glyn:
* Generalise comparisons
* Make regex operator conform to Perl etc.
* Add operator precedence table

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/112
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/113
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/114

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1043,29 +1043,32 @@ logical-or-expr  = logical-and-expr *("||" logical-and-expr)
 logical-and-expr = basic-expr *("&&" basic-expr)      ; conjunction
                                                       ; binds more tightly than disjunction
 
-basic-expr   = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
-existence-expr = [neg-op] path                        ; path existence or non-existence
+basic-expr   = exist-expr / paren-expr / (neg-op paren-expr) / relation-expr
+exist-expr   = [neg-op] path                          ; path existence or non-existence
+path         = rel-path / json-path
+rel-path     = "@" *(dot-selector / index-selector)
 paren-expr   = "(" boolean-expr ")"                   ; parenthesized expression
 neg-op       = "!"                                    ; not operator
 
-comp-expr    = path (comp-op comparable /             ; comparison
-                     regex-op regex     /             ; RegEx test
-                     in-op container)                 ; containment test
-path         = rel-path / json-path
+relation-expr = comp-expr /                           ; comparison test
+                regex-expr /                          ; regular expression test
+                contain-expr                          ; containment test
+
+comp-expr    = comparable comp-op comparable
+comparable   = number / string-literal /              ; primitive ...
+               true / false / null /                  ; values only
+               path                                   ; path value
 comp-op      = "==" / "!=" /                          ; comparison ...
                "<"  / ">"  /                          ; operators
                "<=" / ">="
-regex-op     = "~="                                   ; RegEx match
+
+regex-expr   = regex-op regex
+regex-op     = "=~"                                   ; regular expression match
+regex        = <TO BE DEFINED>
+
+contain-expr = in-op container
 in-op        = " in "                                 ; in operator
-comparable   = number / string-literal /              ; primitive ...
-               true / false / null /                  ; values only
-               rel-path /                             ; relative path value
-               json-path                              ; any value
-
-rel-path     = "@" *(dot-selector / index-selector)
-
 container = <TO BE DEFINED>
-regex = <TO BE DEFINED>
 ~~~~
 
 Notes:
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1043,29 +1043,32 @@ logical-or-expr  = logical-and-expr *("||" logical-and-expr)
 logical-and-expr = basic-expr *("&&" basic-expr)      ; conjunction
                                                       ; binds more tightly than disjunction
 
-basic-expr   = comp-expr / existence-expr / paren-expr / (neg-op paren-expr)
-existence-expr = [neg-op] path                        ; path existence or non-existence
+basic-expr   = exist-expr / paren-expr / (neg-op paren-expr) / relation-expr
+exist-expr   = [neg-op] path                          ; path existence or non-existence
+path         = rel-path / json-path
+rel-path     = "@" *(dot-selector / index-selector)
```
gregsdennis, line 1049: Are we only supporting a single selector here, or does the `*( ... )` mean multiple?

draft-ietf-jsonpath-base.md
```diff
@@ -1043,29 +1043,32 @@ logical-or-expr  = logical-and-expr *("||" logical-and-expr)
...
-                     in-op container)                 ; containment test
-path         = rel-path / json-path
+relation-expr = comp-expr /                           ; comparison test
+                regex-expr /                          ; regular expression test
+                contain-expr                          ; containment test
+
+comp-expr    = comparable comp-op comparable
+comparable   = number / string-literal /              ; primitive ...
+               true / false / null /                  ; values only
+               path                                   ; path value
```
gregsdennis, line 1060: I'm curious, are we planning on supporting literal complex values (i.e. object / arrays) at some point later?  Maybe this needs more discussion.

draft-ietf-jsonpath-base.md
```diff
@@ -1081,10 +1084,21 @@ Notes:
 * The behaviour of operators is consistent with the 'C'-family of programming languages.
 <!-- need to clarify -->
 
+The following table lists filter expression operators in order of precedence from highest to lowest.
+
+| Precedence | Operator type | Syntax |
+|:--:|:--:|:--:|
+|  5  | Grouping | `(...)` |
+|  4  | Logical NOT | `!` |
+|  3  | Relations | `==`<br>`!=`<br>`<`<br>`>`<br>`<=`<br>`>=`<br>`=~`<br>` in ` |
```
gregsdennis, line 1093: Is it necessary to have each operator on its own line?  Seems like we could do with two or three per line.  Just a readability question.

gregsdennis (approved):
This looks good.  Just some curiosity questions from me.

draft-ietf-jsonpath-base.md
glyn: `*()` means zero or more repetitions of the contents of the brackets. See [section 3.6](https://www.rfc-editor.org/rfc/rfc5234.html#section-3.6) of RFC 5234.

draft-ietf-jsonpath-base.md
glyn: I agree it needs more discussion, but let's defer such discussion for now to keep this PR nice and small.

draft-ietf-jsonpath-base.md
glyn: Not really. I've pushed a change which groups similar operators together on single lines.

glyn: Perhaps I realise why this had one operator per line. Putting multiple on one line, separated by space or `&nbsp;` results in operators being wrapped in double quotes in the rendered text version (although not the HTML), thus:  
```
            +============+===========+===================+
              | Precedence |  Operator |       Syntax      |
              |            |    type   |                   |
              +============+===========+===================+
              |     5      |  Grouping |       (...)       |
              +------------+-----------+-------------------+
              |     4      |  Logical  |         !         |
              |            |    NOT    |                   |
              +------------+-----------+-------------------+
              |     3      | Relations |     "==" "!="     |
              |            |           | "<" "<=" ">" ">=" |
              |            |           |        "=~"       |
              |            |           |        "in"       |
              +------------+-----------+-------------------+
              |     2      |  Logical  |         &&        |
              |            |    AND    |                   |
              +------------+-----------+-------------------+
              |     1      |  Logical  |         ||        |
              |            |     OR    |                   |
              +------------+-----------+-------------------+
```
This is pretty ugly, so I'm going to revert to one operator per line until someone can suggest better markup.

glyn: Huh! The double quotes are there in the text version even when there is only one operator per line. Given I can't remove the ugliness, switching back to the multiple operators per line version.

glyn: @cabo Do you know a way of avoiding these extraneous double quotes?

glyn: The extra quotes are added by xml2rfc, but I don't know why or how to avoid this. I [posted](https://mailarchive.ietf.org/arch/msg/xml2rfc/30tTnMMcJHCIH8t8-s_NVLZYrFg/) to the xml2rfc mailing list about this.

glyn: So it seems the extra quotes are caused by a [bug](https://trac.ietf.org/trac/xml2rfc/ticket/600) which will be fixed in the upcoming release of xml2rfc. I propose to leave them be for now and pick up the new release of xml2rfc in due course.

goessner:
>   * The behaviour of operators is consistent with the 'C'-family of programming languages.
>   

I think having the precedence table now, above remark is obsolete.


> +The following table lists filter expression operators in order of precedence from highest to lowest.
> +
> +| Precedence | Operator type | Syntax |
> +|:--:|:--:|:--:|
> +|  5  | Grouping | `(...)` |
> +|  4  | Logical NOT | `!` |
> +|  3  | Relations | `==`<br>`!=`<br>`<`<br>`>`<br>`<=`<br>`>=`<br>`=~`<br>` in ` |
>

glyn:
Thanks @goessner. I raised https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/116 to remove that remark.

PR #125, opened 2021-10-04
contain-expr reworked

goessner:


draft-ietf-jsonpath-base.md
```diff
@@ -1066,9 +1066,11 @@ regex-expr   = regex-op regex
 regex-op     = "=~"                                   ; regular expression match
 regex        = <TO BE DEFINED>
 
-contain-expr = in-op container
+contain-expr = containable in-op container
+containable  = rel-path / json-path /                 ; path to primitive value 
+               number / string-literal
 in-op        = " in "                                 ; in operator
```
glyn, line 1072: Not sure why spaces need to be included in this token. Wouldn't it be better to require arbitrary whitespace either side of the token instead?

draft-ietf-jsonpath-base.md
```diff
@@ -1078,9 +1080,10 @@ Notes:
 <!-- issue: comparison with structured value -->
 * Types are not implicitly converted in comparisons.
   So `"13 == '13'"` selects no node.
-* A member or element value by itself is *falsy* only, if it does not exist. Otherwise it is *truthy*, resulting in its value. To be more specific explicit comparisons are necessary. This existence test — as an exception of the general rule — also works with complex values.
+* A member or element value by itself is *falsy* only, if it does not exist. Otherwise it is *truthy*, resulting in its value. To be more specific explicit comparisons are necessary. This existence test — as an exception of the general rule — also works with structured values.
 * Regular expression tests can be applied to `string` values only.
-* Containment tests work with arrays and objects.
+* The value of LHS `containable` of a `contain-expr` is compared to every single element of the RHS `container`. In case of a match a selection occurs. Containment tests — like comparisons — are restricted to primitive values. So even if a structured `containable` value is equal to a certain structured value in `container`, no selection is done.
```
glyn, line 1085: Why this restriction? This behaviour would probably be surprising to most users.

I think there was resistance to implementing deep equals, but I'm still not sure there's much of an issue in doing that.

glyn, line 1085: Let's avoid unnecessary acronyms:
```suggestion
* The value of the first operand (`containable`) of a `contain-expr` is compared to every single element of the RHS `container`. In case of a match a selection occurs. Containment tests — like comparisons — are restricted to primitive values. So even if a structured `containable` value is equal to a certain structured value in `container`, no selection is done.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1078,9 +1080,10 @@ Notes:
...
+* The value of RHS `container` of a `contain-expr` needs to be resolved to an array. Otherwise nothing is selected.
```
glyn, line 1086:
```suggestion
* The value of the second operand (`container`) of a `contain-expr` needs to be resolved to an array. Otherwise nothing is selected.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1066,9 +1066,11 @@ regex-expr   = regex-op regex
 regex-op     = "=~"                                   ; regular expression match
 regex        = <TO BE DEFINED>
 
-contain-expr = in-op container
+contain-expr = containable in-op container
```
glyn, line 1069: Would `content` would be a better term than `containable`?

glyn (changes requested):
My main issue with this PR is that the restrictions seem quite unintuitive.

draft-ietf-jsonpath-base.md
goessner: In fact I was thinking about using whitespace `W` here. But there are some other locations in the spec where this should be done ... perhaps within a another special whitespace PR.

draft-ietf-jsonpath-base.md
goessner: I consider `content` far too general. Perhaps `contain-val` or `contain-value` is a more specific and better term.

draft-ietf-jsonpath-base.md
goessner: Motivation here was consistency with `comp-expr`, which is restricted to primitives. So loosening this restriction should also impact comparisons. Do we really want to open this Pandora's Box?

glyn: > Motivation here was consistency with comp-expr, which is restricted to primitives. So loosening this restriction should also impact comparisons. Do we really want to open this Pandora's Box?

Perhaps; perhaps not. But I'd like to understand what issues there are, if any, specifying the behaviour of equality tests of non-primitives (essentially, deep equals). I think we agreed at the last meeting that _implementing_ deep equals is relatively straightforward.

draft-ietf-jsonpath-base.md
glyn: Yes, a whitespace cleanup is probably overdue.

draft-ietf-jsonpath-base.md
glyn: I was aiming for symmetry with `container`. If we want to be more specific, then how about something like the following?
```suggestion
contain-expr = primitive-value in-op array-value
```

PR #137, opened 2021-11-28
Address (most of) Michael McCool's comments

cabo:


sthagen (approved):
Looks good to me (should the DELETEME comment go away or the subsequent para?)

sthagen (approved):

goessner (approved):
I consider this an overall improvement ... thanks.

cabo:
On 29. Nov 2021, at 04:20, Stefan Hagen ***@***.***> wrote:
> 
> Looks good to me (should the DELETEME comment go away or the subsequent para?
> 
Both :-)
(Let’s do that consciously in the next PR.)

Grüße, Carsten

draft-ietf-jsonpath-base.md
```diff
@@ -416,17 +417,17 @@ be used with a comparable XML document; see also {{xpath-overview}}.
 ## Overview {#synsem-overview}
 
 A JSONPath query is a string which selects zero or more nodes of a piece of JSON.
-A valid query conforms to the ABNF syntax defined by this document.
 
-A query MUST be encoded using UTF-8. To parse a query according to
-the grammar in this document, its UTF-8 form SHOULD first be decoded into
+A query MUST be encoded using UTF-8.
+The grammar for queries given in this document assumes that its UTF-8 form is first be decoded into
```
glyn, line 422: Typo, should be: "... is first decoded into ..."

(Meta comment: I can no longer make suggestions. The icon has disappeared!)

draft-ietf-jsonpath-base.md
```diff
@@ -416,17 +417,17 @@ be used with a comparable XML document; see also {{xpath-overview}}.
...
 Unicode code points as described
-in {{RFC3629}}.
+in {{RFC3629}}; implementation approaches that lead to an equivalent
+result are possible.
```
glyn, line 425: The statement after the semicolon seems unnecessary. Isn't that always the case for IETF or any other specifications?

draft-ietf-jsonpath-base.md
```diff
@@ -416,17 +417,17 @@ be used with a comparable XML document; see also {{xpath-overview}}.
...
 
 A string to be used as a JSONPath query needs to be *well-formed* and
 *valid*.
 A string is a well-formed JSONPath query if it conforms to the syntax
-of JSONPath.
+of JSONPath, as defined by the ABNF syntax in this document.
```
glyn, line 430: I suggest the more concise: "A string is a well-formed JSONPath query if it conforms to the ABNF syntax in this document."

draft-ietf-jsonpath-base.md
```diff
@@ -624,17 +626,18 @@ ALPHA           =  %x41-5A / %x61-7A    ; A-Z / a-z
 ~~~~
 
 Member names containing other characters than allowed by
```
glyn, line 628: This isn't strictly in the PR, but since we are changing the sentence, better English would be "Member names containing characters other than allowed by".

draft-ietf-jsonpath-base.md
```diff
@@ -646,7 +649,8 @@ not encode Unicode characters.
 #### Syntax
 {: unnumbered}
 
-The dot wild card selector has the form `.*`.
+The dot wild card selector has the form `.*` as defined in the
+following syntax:
```
glyn, line 653: This is getting a little wordy and would probably need to be applied throughout the document for consistency. Isn't it obvious that a piece of ABNF in a syntax section is the normative syntax for the construct being discussed?

draft-ietf-jsonpath-base.md
```diff
@@ -1087,13 +1098,15 @@ Notes:
...
+<!-- FIXME: Should the syntax column be split between unary and binary operators? -->
+
 | Precedence | Operator type | Syntax |
 |:--:|:--:|:--:|
 |  5  | Grouping | `(...)` |
 |  4  | Logical NOT | `!` |
 |  3  | Relations | `==`&nbsp;`!=`<br>`<`&nbsp;`<=`&nbsp;`>`&nbsp;`>=`<br>`=~`<br>` in ` |
 |  2  | Logical AND | `&&` |
-|  1  | Logical OR | `\|\|` |
+|  1  | Logical OR | `¦¦`   |
```
glyn, line 1109: The symbol looks wrong to me. It appears to consist of two punctured vertical bars (`¦¦`, each of which is U+00A6) rather than vertical bars (`||`, each of which is U+007C).

draft-ietf-jsonpath-base.md
```diff
@@ -1184,6 +1197,11 @@ the emerging JSON community became apparent, specifically for:
   client, so the server can reduce the amount of data in its response,
   minimizing bandwidth usage.
 
+(Note that XPath has evolved since 2007, and recent versions even
+nominally support operating inside JSON values.
+This appendix only discusses the more widely used version of XPath
+that was available in 2007.)
```
glyn, line 1203: Perhaps there is a case for deleting the appendix? I don't really think it adds much value.

glyn (changes requested):

draft-ietf-jsonpath-base.md
cabo: To be fixed in next commit.

draft-ietf-jsonpath-base.md
cabo: Yes, but we also tend to always point that out.

draft-ietf-jsonpath-base.md
cabo: Fixed in next commit.

draft-ietf-jsonpath-base.md
cabo: Fixed in next commit.

draft-ietf-jsonpath-base.md
cabo: The reason for the wordiness is that the sentence ended in ".*." before -- at least in the plaintext rendering.  So we now have to contort our sentences to avoid that run-in.

draft-ietf-jsonpath-base.md
cabo: That is a markdown trick.  The backslashes are definitely wrong. ¦¦ is rendered || by xml2rfc.

draft-ietf-jsonpath-base.md
cabo: We can discuss this; I think it is still useful for people who know XPath (who by the way typically will know 1.0, not 3.0).

glyn (approved):
Thanks @cabo! Nice work.

PR #138, opened 2021-12-01
Relegate XPath details to the appendix

glyn:
XPath details are a distraction in the body of the spec.

In the future, readers are increasingly likely to come to this spec with no exposure to XPath or, for that matter, XML.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1301,8 +1301,8 @@ and their JSONPath equivalents.
 {: #tbl-xpath-equivalents title="Example XPath expressions and their JSONPath equivalents"}
 
 XPath has a lot more functionality (location paths in unabbreviated syntax,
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -188,8 +188,7 @@ Name:
   It is included here for completeness.
 
 Element:
-: A value in an array.  (Also used with a distinct meaning in XML
-  context for XML elements.)
+: A value in an array.
```
cabo, line 191: I'd probably keep a 

(not to be confused with XML element).

cabo (approved):
Looks very good, with the one nit.
Thanks!

draft-ietf-jsonpath-base.md
glyn: Ok, I grudgingly agree. ;-) I promise a beer to the first person who reads the spec and says "What's XML?"...

PR #139, opened 2021-12-01
Delete the "in" operator

glyn:
From the meeting at IETF 112 ([1]):

**Consensus**: 'in' won't be used, but it will be easy to add later if there's real demand

[1] https://datatracker.ietf.org/doc/minutes-112-jsonpath/


cabo (approved):
Thanks!  I pushed a small commit removing an "in" example.

glyn:
> Thanks! I pushed a small commit removing an "in" example.

Good eyes. Thanks!

goessner:
Yes, we agreed to remove the 'in' operator as some kind of syntactic sugar.

thanks

PR #140, opened 2021-12-01
Delete expression language

glyn:
There does not seem to be an appetite for doing the necessary spec work. This PR will test that assertion.

Filter expressions are still supported, but not "container expressions", of the form [(...)].

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -188,7 +188,7 @@ Name:
   It is included here for completeness.
 
 Element:
-: A value in an array. (Not to be confused with XML element)
+: A value in an array. (Not to be confused with XML element.)
 
 Index:
 : A non-negative integer that identifies a specific element in an array.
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1142,30 +1125,6 @@ Some examples:
...
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
-
-# Expression Language
-
-> Task (T2): Separate out expression language.  For now, this section
-> is a repository for ABNF taken from {{-json}}.  This needs to be
-> deduplicated with definitions above.
-
-~~~~ abnf
```
cabo, line 1152: We need to keep `number`, `true`, `false`, `null`.
(We also need to define slice-index.)

glyn: > We need to keep number, true, false, null.

Oops. I reintroduced them.

> (We also need to define slice-index.)

Noted. Out of scope for this PR.

goessner (approved):

gregsdennis:
I think this can be added later, similar to the consensus for `in`.

Personally I still see value in having it.

PR #141, opened 2021-12-01
Continue 137 (technical changes)

cabo:


draft-ietf-jsonpath-base.md
```diff
@@ -1113,19 +1106,17 @@ The following table lists filter expression operators in order of precedence fro
...
-| Number |  `!0`   | `true` | `false` for non-zero number  |
-| String |  `!""`<br>`!''`  | `true` | `false` for non-empty string  |
-| `null` |  `!null`| `true` | —  |
-| `true` |  `!true`| `false`| —  |
-| `false`| `!false`| `true` | —  |
-| Object | `!{}`<br>`!{a:0}` | `false`| always `false` |
-| Array | `![]`<br>`![0]` | `false`| always `false` |
+| Type    | Negation          | Result   | Comment        |
+| :----:  | :---------:       | :------: | :-------:      |
+| Number  | `!0`              | `false`  | always `false` |
```
glyn, line 1113: Would it be worth adding a row for `!1` to hammer home the rules?

glyn (commented):
Much as I continue to hate "falsiness", this is progress. One comment.

draft-ietf-jsonpath-base.md
cabo: Good idea.  !1 and !"a" like in the other rows with multiple values.

goessner: yes .. good catch.

draft-ietf-jsonpath-base.md
```diff
@@ -1113,19 +1106,17 @@ The following table lists filter expression operators in order of precedence fro
 
 The `filter-selector` works with arrays and objects exclusively. Its result might be a list of *zero*, *one*, *multiple* or *all* of their element or member values then. Applied to other value types, it will select nothing.
 
-**FIXME**: The zero number/empty string exceptions are no longer true.  Booleans work the same everywhere.
-
 Negation operator `neg-op` allows to test *falsiness* of values.
```
goessner, line 1109: So what about:

Negation operator `neg-op` allows to test for implicit boolean conversion of values resulting in `false`.

draft-ietf-jsonpath-base.md
```diff
@@ -1113,19 +1106,17 @@ The following table lists filter expression operators in order of precedence fro
...
+| Number  | `!0`              | `false`  | always `false` |
+| String  | `!""`<br>`!''`    | `false`  | always `false` |
+| `null`  | `!null`           | `true`   | —              |
+| `true`  | `!true`           | `false`  | —              |
+| `false` | `!false`          | `true`   | —              |
+| Object  | `!{}`<br>`!{a:0}` | `false`  | always `false` |
+| Array   | `![]`<br>`![0]`   | `false`  | always `false` |
 {: title="Test falsiness of JSON values" }
 
 Applying negation operator twice `!!` gives us *truthiness* of values.
```
goessner, line 1122: In the same way:

Applying negation operator twice `!!`  allows to test for implicit boolean conversion of values resulting in `true`.

glyn (approved):
LGTM, thanks.

draft-ietf-jsonpath-base.md
glyn: Applying a similar rewording to above gives us:
```suggestion
Negation operator `neg-op` applied twice to a non-boolean value implicitly converts the value to boolean.
```

I find the terms "falsiness" and "truthiness" to be fairly opaque. If we are going to use them, we'll need to define them properly. But unless they add value for some readers, we could drop them as per my suggested rewordings.

draft-ietf-jsonpath-base.md
glyn: > Negation operator `neg-op` allows to test *falsiness* of values.

The English is rather stilted, so please can we tidy it up as part of this PR?

I'd suggest something like:
```suggestion
Negation operator `neg-op` of a non-boolean value implicitly converts the value to boolean and then negates the result.
```

glyn:
@cabo It would be good to address @goessner's comments in this PR. I have made some suggestions, but feel free to pick them apart if you are keen to retain "truthiness" and "falsiness". BTW perhjaps I struggle with the latter term because of the existence of some [English slang](https://idioms.thefreedictionary.com/falsies).

glyn:
> Indeed. It is not that easy to say here what needs to be said, so I'm still in thinking mode. Thanks for the pointer to English slang; of course I was completely unaware. We don't want to use the fooiness terms. Interestingly, the remaining grammar does not seem to have a place where JSON values are implicitly converted to Boolean; this is now all explicit unless I'm missing something. So maybe we don't need this table (and its legend) at all?

You're right! The two occurrences of `neg-op` are as follows.

> exist-expr   = [neg-op S] path

which simply tests for non-existence of a value(s) selected by a path and does not depend on the actual values themselves.

> paren-expr   = [neg-op S] "(" S boolean-expr S ")"    ; parenthesized expression

which negates a _boolean_ expression. Looking at the details of `boolean-expr`, there's no way it can result in a value which might need to be converted to boolean.

So please can we scrap the table and kiss "fooiness" goodbye once and for all?

cabo:
I think we should first merge it and then delete it, in case we need to resurrect it later.

goessner:
what's about ...`$[? !(@.heaven == 'blue' && @.sun == 'shining')]` which can surely be converted to a negated form. But sometimes someone doesn't want to do that ... like in any programming language.

Nevertheless I second to kick fooiness out.

cabo:
> what's about ...`$[? !(@.heaven == 'blue' && @.sun == 'shining')]` which can surely be converted to a negated form. 

Comparisons already yield Boolean values, as does `&&`/`||`.

goessner:
Well, then in fact I misunderstood Glyn's comment above ... I agree to remove the conversion table as superfluous.

PR #143, opened 2021-12-02
Allow any regular expression syntax

glyn:
This is a deliberately provocative PR showing one option in the absence of a suitable regular expression standard.

Perhaps it will at least help us get started defining the interface between JSONPath and a regular expression standard.

This is hopefully a temporary measure until we have a standard we want to mandate.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1048,9 +1048,9 @@ true         = %x74.72.75.65                          ; true
 false        = %x66.61.6c.73.65                       ; false
 null         = %x6e.75.6c.6c                          ; null
 
-regex-expr   = (path / string-literal) S regex-op S regex
+regex-expr   = (path / string-literal) S regex-op S "/" regex "/"
 regex-op     = "=~"                                   ; regular expression match
-regex        = <TO BE DEFINED>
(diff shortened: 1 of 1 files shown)


cabo:
Two comments:

* I don't think we want to keep an open hole in here.  (I also don't know why that would be necessary.)
* I don't think we want to emulate /.../ syntax.  We have a perfectly fine string syntax and can use that for regexps as well.  This will also remind people to not just copy-paste their Perl programs into JSONPath.

glyn:
Seems like the provocation worked! ;-)

> Two comments:
> 
>     * I don't think we want to keep an open hole in here.  (I also don't know why that would be necessary.)

It would allow us to issue an interim standard JSONPath which is completely permissive about which regular expressions can be used with it. Without this, we'll need iregexp or equivalent to reach RFC status first.

> 
>     * I don't think we want to emulate /.../ syntax.  We have a perfectly fine string syntax and can use that for regexps as well.  This will also remind people to not just copy-paste their Perl programs into JSONPath.

I was looking at the consensus among JSONPath implementations that already support regexps (see [this comparison](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_regular_expression.html)). If we drop the /.../ syntax, all existing JSONPath expressions which use regexps won't work with standard JSONPath.

glyn:
Consensus is to pursue iregexp.

Issue #127, opened 2021-10-25
Should we provide for structured types in containable?


cabo:
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/125#discussion_r721614374

goessner:
After thinking a while about comparing structured types, I now see no problem to support equality of  them. I agree to handling them is straightforward:

> Two structured types are equal, if they are equal on a per-node basis. 

also see [The Abstract Equality Comparison Algorithm](https://262.ecma-international.org/5.1/#sec-11.9.3) ... not exactly, what we want though.

We need to be clear about order ... !
```
[1,2] == [1,2]
{x:1,y:2} == {x:1,y:2}
[1,2] == [2,1]  // ?
{x:1,y:2} == {y:2,x:1} // ?
[] == []
{} == {}
```
Containables may be defined on exactly that comparison rule. 

I believe, we need to add objects here again.  Also we need to discuss subsets.

### Array Examples:
```
1 in [1,2,3]             // primitive value lhs
[1,2] in [1,2,3]         // structured value lhs
[1,2] in [1,2,3,[1,2]] // subset
[] in [1,2,3]             // empty array
[] in [1,2,3,[]]          // empty array
{} in [1,2,3,{}]          // empty object
{x:1,y:2} in [{x:1,y:2}]
```
### Object Examples:
```
'x' in {x:1,y:2}
['x','y'] in {x:1,y:2}
{x:1} in {x:1,y:2}
{z:3,x:1} in {x:1,y:2,z:3}
{} in  {x:1,y:2}
```

Once having agreement about handling that edge cases (maybe there are more), we might add it to the spec. 

Btw.: JavaScript resolves all comparisons above to `false`, since objects are equal only, if they are referencing the same object (no deep equals).

glyn:
Comments inline.

On Tue, 2 Nov 2021 at 11:48, Stefan Goessner ***@***.***>
wrote:

> After thinking a while about comparing structured types, I now see no
> problem to support equality of them. I agree to handling them is
> straightforward:
>
> Two structured types are equal, if they are equal on a per-node basis.
>
> also see The Abstract Equality Comparison Algorithm
> <https://262.ecma-international.org/5.1/#sec-11.9.3> ... not exactly,
> what we want though.
>
> We need to be clear about order ... !
>
> [1,2] == [1,2]
> {x:1,y:2} == {x:1,y:2}
[...]

PR #145, opened 2022-01-17
ABNF work done

goessner:
I did some work on ABNF texts. 

Especially I tried to get equality comparisons of structured values right.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -588,11 +588,11 @@ A dot selector starts with a dot `.` followed by an object's member name.
 ~~~~ abnf
 dot-selector    = "." dot-member-name
 dot-member-name = name-first *name-char
-name-first =
+name-first      =
                       ALPHA /
                       "_"   /           ; _
                       %x80-10FFFF       ; any non-ASCII Unicode character
-name-char = DIGIT / name-first
+name-char       = name-first / DIGIT
 
 DIGIT           =  %x30-39              ; 0-9
 ALPHA           =  %x41-5A / %x61-7A    ; A-Z / a-z
@@ -640,7 +640,14 @@ string, or true/false/null) selects no node.
 An index selector `[<index>]` addresses at most one object member value or at most one array element value.
 
 ~~~~ abnf
-index-selector      = "[" S (quoted-member-name / element-index) S "]"
+index-selector = "[" S (quoted-member-name / element-index) S "]"
+
+B              =  %x20 / ; Space
+                  %x09 / ; Horizontal tab
+                  %x0A / ; Line feed or New line
+                  %x0D   ; Carriage return
+
+S              =  *B     ; optional blank space
 ~~~~
 
 Applying the `index-selector` to an object value in its input nodelist, a
@@ -652,8 +659,10 @@ the JSONPath syntax allows strings to be enclosed in _single_ or _double_ quotes
 ~~~~ abnf
 quoted-member-name  = string-literal
 
-string-literal      = %x22 *double-quoted %x22 /       ; "string"
+string-literal      = json-string /                    ; "string"
                       %x27 *single-quoted %x27         ; 'string'
+
+json-string         = %x22 *double-quoted %x22         ; "string"
 
 double-quoted       = unescaped /
                       %x27      /                       ; '
@@ -672,16 +681,14 @@ unescaped           = %x20-21 /                         ; s. RFC 8259
                       %x28-5B /                         ; omit '
                       %x5D-10FFFF                       ; omit \
 
-escapable           = ( %x62 / %x66 / %x6E / %x72 / %x74 / ; \b \f \n \r \t
-                          ; b /         ;  BS backspace U+0008
-                          ; t /         ;  HT horizontal tab U+0009
-                          ; n /         ;  LF line feed U+000A
-                          ; f /         ;  FF form feed U+000C
-                          ; r /         ;  CR carriage return U+000D
-                          "/" /          ;  /  slash (solidus)
-                          "\" /          ;  \  backslash (reverse solidus)
-                          (%x75 hexchar) ;  uXXXX      U+XXXX
-                      )
+escapable           = %x62 /         ; b  BS backspace U+0008
+                      %x66 /         ; f  FF form feed U+000C
+                      %x6E /         ; n  LF line feed U+000A
+                      %x72 /         ; r  CR carriage return U+000D
+                      %x74 /         ; t  HT horizontal tab U+0009
+                      %x47 /         ; /  slash (solidus)
+                      ESC  /         ; \  backslash (reverse solidus)
+                      (%x75 hexchar) ;  uXXXX      U+XXXX
 
 hexchar = non-surrogate / (high-surrogate "\" %x75 low-surrogate)
 non-surrogate = ((DIGIT / "A"/"B"/"C" / "E"/"F") 3HEXDIG) /
@@ -783,14 +790,6 @@ slice-selector = "[" S [start S] ":" S [end S] [":" S [step S]] "]"
 start          = int       ; included in selection
 end            = int       ; not included in selection
 step           = int       ; default: 1
-
-B              =    %x20 / ; Space
-                    %x09 / ; Horizontal tab
-                    %x0A / ; Line feed or New line
-                    %x0D   ; Carriage return
-S              = *B        ; optional blank space
-RS             = 1*B       ; required blank space
-
 ~~~~
 
 The `slice-selector` consists of three optional decimal integers separated by colons.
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1006,55 +1005,66 @@ The current item is selected if and only if the result is `true`.
...
 Notes:
 
 * Parentheses can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `[?(<expr>)]` is naturally contained in the current lean syntax `[?<expr>]` as a special case.
-* Comparisons are restricted to primitive values (such as number, string, `true`, `false`, `null`). Comparisons with complex values will fail, i.e. no selection occurs.
-<!-- issue: comparison with structured value -->
 * Types are not implicitly converted in comparisons.
   So `"13 == '13'"` selects no node.
-* A member or element value by itself in a Boolean context is
-  interpreted as `false` only if it does not exist.
+* Relational operators (`>`, `<`, `>=`, `<=`) are working on primitive number values exclusively.
```
glyn, line 1059:
```suggestion
* Relational operators (`>`, `<`, `>=`, `<=`) are intended for comparing primitive number values. 
```

draft-ietf-jsonpath-base.md
```diff
@@ -1006,55 +1005,66 @@ The current item is selected if and only if the result is `true`.
...
-logical-or-expr  = logical-and-expr *(S "||" S logical-and-expr)
-                                                      ; disjunction
-                                                      ; binds less tightly than conjunction
-logical-and-expr = basic-expr *(S "&&" S basic-expr)  ; conjunction
-                                                      ; binds more tightly than disjunction
+boolean-expr = or-expr   
+
+or-expr      = and-expr *(S "||" S and-expr)        ; logical or operator
+
+and-expr     = basic-expr *(S "&&" S basic-expr)    ; logical and operator
```
goessner, line 1013: Comply to implicit `xxx-expr` naming rule.

draft-ietf-jsonpath-base.md
```diff
@@ -1006,55 +1005,66 @@ The current item is selected if and only if the result is `true`.
...
 * Parentheses can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `[?(<expr>)]` is naturally contained in the current lean syntax `[?<expr>]` as a special case.
-* Comparisons are restricted to primitive values (such as number, string, `true`, `false`, `null`). Comparisons with complex values will fail, i.e. no selection occurs.
-<!-- issue: comparison with structured value -->
 * Types are not implicitly converted in comparisons.
   So `"13 == '13'"` selects no node.
-* A member or element value by itself in a Boolean context is
-  interpreted as `false` only if it does not exist.
+* Relational operators (`>`, `<`, `>=`, `<=`) are working on primitive number values exclusively. 
+Comparison with a non-number value(s) fails, i.e. no selection occurs.
+* Equality operators (`==`, `!=`) on primitive and/or structured values are working as expected, i.e. selection occurs in case of deep equality/inequality.
```
glyn, line 1061: What happens when a path is compared to an array and the path produces a nodelist which happens to be the same as the contents of the array, in the same order?

goessner, line 1061: Not sure, if we want to include lexical sorting also ...

draft-ietf-jsonpath-base.md
```diff
@@ -672,16 +681,14 @@ unescaped           = %x20-21 /                         ; s. RFC 8259
...
-                          "/" /          ;  /  slash (solidus)
-                          "\" /          ;  \  backslash (reverse solidus)
-                          (%x75 hexchar) ;  uXXXX      U+XXXX
-                      )
+escapable           = %x62 /         ; b  BS backspace U+0008
+                      %x66 /         ; f  FF form feed U+000C
+                      %x6E /         ; n  LF line feed U+000A
+                      %x72 /         ; r  CR carriage return U+000D
+                      %x74 /         ; t  HT horizontal tab U+0009
+                      %x47 /         ; /  slash (solidus)
```
cabo, line 689: We do not use "b" because that would be case insensitive in ABNF.
We currently use %x62, we could use %s"b" instead.
But we never need to use %x47, as "/" is unambiguous.

draft-ietf-jsonpath-base.md
goessner: I have in mind to allow only a path referencing a single array having the same content as the array to be compared. Nodelists are to be discussed ...

draft-ietf-jsonpath-base.md
goessner: ok ,,, got it.

draft-ietf-jsonpath-base.md
goessner: yes ... better

cabo:
While we are fixing the ABNF, `slice-index` is undefined; this needs to be done by opening up `slice-selector` and putting the innards in a `slice-index` production, I think.

cabo:
> While we are fixing the ABNF, `slice-index` is undefined; this needs to be done by opening up `slice-selector` and putting the innards in a `slice-index` production, I think.

Fixed in #147

draft-ietf-jsonpath-base.md
```diff
@@ -1006,55 +1005,66 @@ The current item is selected if and only if the result is `true`.
...
-               "<"  / ">"  /                          ; operators
-               "<=" / ">="
-number       = int [ frac ] [ exp ]                   ; decimal number
-frac         = "." 1*DIGIT                            ; decimal fraction
-exp          = "e" [ "-" / "+" ] 1*DIGIT              ; decimal exponent
-true         = %x74.72.75.65                          ; true
-false        = %x66.61.6c.73.65                       ; false
-null         = %x6e.75.6c.6c                          ; null
+comparable   = json-value / path                    ; JSON value or path value
+json-value   = number / string-literal /            ; primitive
```
cabo, line 1028: To me it seemed the resolution of #127 at https://datatracker.ietf.org/doc/minutes-112-jsonpath/ implied that we weren't going to entertain notating structured values in a JSONPath expression at all.  So this change surprises me a little.
Maybe this was just my perception, or if not, maybe we want to revert that aspect.

goessner: hmm ... I am quite sure that we somewhere, somehow agreed, that excluding structured values from equality ...
```
color == [12, 125, 255]
```
is kind of arbitrary and artificial.  Can't find it yet though. Glyn for help here ... `?

danielaparker: > We also have to define the semantics, which is relatively straightforward for most kinds of JSON values (lexicographic ordering). 

You're not suggesting lexicographic ordering for numbers, are you?

> But there is no natural total ordering for maps that I am aware of, so what does it mean to have a `<` in an expression (which, of course, needs to stay syntactically allowed) and then two maps turning up? Oh, and what is the value of `1 < true` (already a problem without this extension)? What does it mean that `<` does not "work"?

All of these issues have been addressed in prior work, in different ways. JMESPath, which defines ordering operators only for numbers, specifies exactly what it means for `1 < true` or `1 < {}`, see [ordering operators](https://jmespath.org/specification.html#ordering-operators). Other approaches define a complete ordering on JSON values, assuming where types differ that null < boolean < number < object < array < string (following Python.)

goessner: I am proposing:
```
* Relational operators (`>`, `<`, `>=`, `<=`) are intended for comparing primitive number values. 
Comparison with a non-number value(s) fails, i.e. no selection occurs.
``` 
So there is no need to discuss ordering of structured values. In that case no selection occurs,
We can include strings on basis of lexical ordering though.

Regarding `1 < true` ... we opposed against implicit type conversion, so no selection occurs here.

But ... as I was always in favor of comparing primitive values only, I would be happy to return to that clean solution again. I am obviously wrong in interpreting the discussion in context `contain-expr` (which is obsolete now) that way, that it is natural to include structured values in comparisons. (#127)

glyn: My recollection, which could easily be faulty, is that we didn't want to support non-primitive literal values because of the cost/benefit ratio being high. I *think* the jury was out on whether we would support ("deep") equals comparisons of non-primitive (non-literal) values, but again I may have misremembered. It's hard to keep track of past discussions...

danielaparker: > My recollection, which could easily be faulty, is that we didn't want to support non-primitive literal values because of the cost/benefit ratio being high. I _think_ the jury was out on whether we would support ("deep") equals comparisons of non-primitive (non-literal) values, but again I may have misremembered. It's hard to keep track of past discussions...

On the other hand, there's plenty of prior experience to draw on, including a few JSONPath implementations, and the very completely specified JMESPath. It's not necessary to think this all through as if from the beginning.

draft-ietf-jsonpath-base.md
glyn: Perhaps Singular Paths (https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/164) helps with this.

goessner:
Consensus on today's interim WG meeting was not to do equality comparisons of structured values yet.

timbray:
This has come up in multiple WG conversations and interim meetings and has failed to get consensus support. Difficulties include: requirements for unbounded recursion.  Cost of rejecting this PR: ?@.foo == $.bar returns false if either/both sides of the comparison return a structured value.

PR #147, opened 2022-01-23
Rename "union selector" to "list selector"

### LINKED ISSUE #21: "Union" could have more description, and maybe a new name
The term "union" will likely be unfamiliar for readers (it is for me) and could use a bit more description than "A union ~matcher~ selector consists of one or more union elements."

"Union" isn't a common term outside of set manipulation.  I would call the `[...]` an "indexer," which is a much more well-known term and is already associated with this construct in many programming languages.  The values inside are indices (as opposed to "union children" or "union elements"), of which there are multiple types:

- Integer
- Slice
- String
- Container query expression
- Item query expression

See [selector query expressions](https://github.com/jsonpath-standard/internet-draft/issues/17) for more details on the query-style indexers.

cabo:
Clarify related text; add explanation of change.
Take slice-selector apart to define slice-index.
Close #21.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -993,9 +993,10 @@ list-entry     =  ( quoted-member-name /
 {: unnumbered}
 
 A list selector selects the nodes that are selected by at least one of
-the selectors in the list and yields the concatenation of the
-lists (in the order of the selectors) of nodes selected by the list elements.
-Note that any node selected in more than one of the list selectors is kept
+the selector entries in the list and yields the concatenation of the
+lists (in the order of the selector entries) of nodes selected by the
+selector entries.
+Note that any node selected in more than one of the selector entries is kept
 as many times in the node list.
 
 To be valid, integer values in the `element-index` and `slice-index`
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -969,7 +969,7 @@ slices in a single selector.
 Note: The list selector was called "union selector" in
 {{JSONPath-orig}}, as it was intended to solve use cases addressed by
 the union selector in XPath.
-However, the term "union" implies the set operation that involves
+However, the term "union" has the connotation of a set operation that involves
 merging input sets while avoiding duplicates, so the concept was
 renamed into "list selector".
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -315,7 +315,7 @@ $.store.book[?(@.price < 10)].title
 | `..`               | nested descendants                                                           |
 | `*`                | wildcard: all member values/array elements regardless of their names/indices |
 | `[]`               | subscript operator: index current node as an array (from 0)                  |
-| `[,]`              | Union operator JSONPath allows alternate(??) names or array indices as a set |
+| `[,]`              | list operator: allows combining member names, array indices, and slices |
 | `[start:end:step]` | array slice operator                                                         |
 | `?()`              | applies a filter expression                                                  |
 | `()`               | expression, e.g., for indexing                                               |
@@ -477,7 +477,7 @@ json-path = root-selector *(S (dot-selector        /
                                dot-wild-selector   /
                                index-selector      /
                                index-wild-selector /
-                               union-selector      /
+                               list-selector       /
                                slice-selector      /
                                descendant-selector /
                                filter-selector))
@@ -556,7 +556,8 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Array slice selector `[<start>:<end>:<step>]`, where the optional
     values `<start>`, `<end>`, and `<step>` are integer literals.
   * Nested descendants selector `..`.
-  * Union selector `[<sel1>,<sel2>,...,<selN>]`, holding a comma delimited list of index, index wild card, array slice, and filter selectors.
+  * List selector `[<sel1>,<sel2>,...,<selN>]`, holding a comma
+    separated list of index and slice selectors.
   * Filter selector `[?(<expr>)]`
   * Current item selector `@` (used in expressions)
 
@@ -588,7 +589,7 @@ A dot selector starts with a dot `.` followed by an object's member name.
 ~~~~ abnf
 dot-selector    = "." dot-member-name
 dot-member-name = name-first *name-char
-name-first =
+name-first      =
                       ALPHA /
                       "_"   /           ; _
                       %x80-10FFFF       ; any non-ASCII Unicode character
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -556,7 +556,8 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Array slice selector `[<start>:<end>:<step>]`, where the optional
     values `<start>`, `<end>`, and `<step>` are integer literals.
   * Nested descendants selector `..`.
-  * Union selector `[<sel1>,<sel2>,...,<selN>]`, holding a comma delimited list of index, index wild card, array slice, and filter selectors.
+  * List selector `[<sel1>,<sel2>,...,<selN>]`, holding a comma
+    separated list of index and slice selectors.
```
gregsdennis, line 560: I think calling it that it only hold index and slice selectors is good.  During the meeting, I feared that it would merely a "collection of selectors," implying that it could itself hold a list selector, which would be recursive and confusing.

Do we need to add "member names" as is mentioned on line 966?

draft-ietf-jsonpath-base.md
```diff
@@ -958,19 +961,29 @@ In the resultant nodelist:
...
+
+The list selector allows combining member names, array indices, and
+slices in a single selector.
+
+Note: The list selector was called "union selector" in
+{{JSONPath-orig}}, as it was intended to solve use cases addressed by
+the union selector in XPath.
+However, the term "union" implies the set operation that involves
+merging input sets while avoiding duplicates, so the concept was
+renamed into "list selector".
```
gregsdennis, line 974: I'm not sure this note is necessary for the spec.  I don't think historical context is right for specifying language.  Maybe this kind of information needs its own area in the document or even externally.

draft-ietf-jsonpath-base.md
```diff
@@ -958,19 +961,29 @@ In the resultant nodelist:
...
 
 #### Syntax
 {: unnumbered}
 
-The union selector is syntactically related to the
-`index-selector`.
+The list selector is syntactically related to the
+`index-selector` and the `slice-selector`.
```
gregsdennis, line 980: What's the relationship between `index-selector` and member names? That might answer my question from above as well.  Member names are again mentioned on 986.  I just find it confusing to have two terms in some descriptions and three in others.

Maybe this could be addressed in a follow-up PR?

draft-ietf-jsonpath-base.md
```diff
@@ -958,19 +961,29 @@ In the resultant nodelist:
...
-### Union Selector
+### List Selector
+
+The list selector allows combining member names, array indices, and
+slices in a single selector.
+
+Note: The list selector was called "union selector" in
+{{JSONPath-orig}}, as it was intended to solve use cases addressed by
+the union selector in XPath.
+However, the term "union" implies the set operation that involves
```
glyn, line 972: If we keep this note (and I'm not sure we should), I suggest the following rewording:
```suggestion
However, the term "union" has the connotation of a set operation that involves
```

draft-ietf-jsonpath-base.md
```diff
@@ -958,19 +961,29 @@ In the resultant nodelist:
...
 
 ~~~~ abnf
-union-selector = "[" S union-entry 1*(S "," S union-entry) S "]"
+list-selector  = "[" S list-entry 1*(S "," S list-entry) S "]"
 
-union-entry    =  ( quoted-member-name /
+list-entry     =  ( quoted-member-name /
                     element-index      /
                     slice-index
                   )
```
glyn, line 989: It seems that lists can't contain filters (neither could unions before this PR). I wonder if this is deliberate or accidental? The comparison project shows there is [no consensus](https://cburgmer.github.io/json-path-comparison/results/union_with_filter.html) among implementations.

glyn (commented):
A couple of comments. I think the general direction  of this PR is good.

draft-ietf-jsonpath-base.md
gregsdennis: Interesting. I'm not sure that you should be forbidden to do so, either. I believe I support it already.

draft-ietf-jsonpath-base.md
cabo: index selectors can contain quoted member names, so that should not be necessary (after we have called it out already).

draft-ietf-jsonpath-base.md
cabo: I believe previous users of JSONPath are part of the target audience.
It would be confusing to just rename the thing.

draft-ietf-jsonpath-base.md
cabo: At this point, the reader should already know what an index selector is.

draft-ietf-jsonpath-base.md
cabo: Now https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/150

draft-ietf-jsonpath-base.md
```diff
@@ -1192,7 +1206,7 @@ with similar XPath concepts.
 | `*`   | `*`                | wildcard: All XML elements regardless of their names                                                                                  |
 | `@`   | n/a                | attribute access: JSON values do not have attributes                                                                                  |
 | `[]`  | `[]`               | subscript operator used to iterate over XML element collections and for predicates                                                    |
-| `¦`   | `[,]`              | Union operator (results in a combination of node sets); JSONPath allows alternate names or array indices as a set                     |
+| `¦`   | `[,]`              | Union operator (results in a combination of node sets); called list operator in JSONPath, allows combining member names, array indices, and slices |
```
goessner, line 1209: ... combining *quoted* member names, array indices, ...

cabo: Is the fact that the member names are quoted relevant here?  It is not mentioned in []...

goessner: As we explicitly distinguish between `dot-member-name` and `quoted-member-name` we should be concrete here also.

draft-ietf-jsonpath-base.md
gregsdennis: My argument is that there is no historical spec. I think this document should be written as if it is defining the syntax from the start.  Any and all references to previous materials, including XPath, should be in an appendix or a separate document.

cabo: Previous users of JSONPath did have a spec, just not a very detailed one or a very official one.  Generally, aspects of usability trump arbitrary principles that do not by themselves contribute to a good outcome.  Why do we waste time on this discussion.

gregsdennis: It was a proposal in a blog post.  Such things need to include this kind of context because users have nothing else to draw upon, and given that XPath was the inspiration, its inclusion in the post is warranted.

However a specification is intended to _define_ a thing.  Purpose is needed, but stories of inspiration are not.  If you want to put some history in an appendix, okay, I guess, but historical context should not exist in the midst of the definition (either syntactical definition or behavioral).

glyn (approved):
LGTM.

draft-ietf-jsonpath-base.md
glyn: I agree with @cabo on this. The term "union" is pretty widespread among JSONPath implementations and the renaming deserves being called out in the body of the document. If we don't explain the rationale, people will wonder why we didn't stick with "union". We could put this in an appendix, but that seems rather artificial/pedantic to me.

glyn:
I'm going to merge this PR to make forward progress. Tweaks can be added via further PRs as usual.

PR #149, opened 2022-01-23
Define application/jsonpath media type

cabo:


sthagen (approved):

glyn (approved):
Looks plausible.

cabo:
I merged this so once we have published an I-D with this change in it we can initiate the necessary review on the media-types list.

Issue #67, opened 2021-03-10
Differentiation from JSON Pointer


fiestajetsam:
(This was original raised in #54 and now being split out)

> I anticipate being asked "Why is JSON Pointer not sufficient?" Indeed its abstract says:
>
>   JSON Pointer defines a string syntax for identifying a specific value within a JavaScript Object Notation (JSON) document.
>
>... which sounds awfully similar.  If we could include a sentence about
that, or a link to an answer, that might be helpful.
*(Murray S. Kucherawy)*

> No - it's not similar in concept, they're separate things. If you really wanted to mention JSON Pointer, you could say something like "Note that while JSON Pointer (RFC xxxx) is already standardised, it is designed to provide a reference to a single, specific part of a JSON document, whereas JSONPath provides the ability to query a document and potentially return multiple values."
*(Mark Nottingham)*

>The short answer is that JSON pointer is good if you already know the structure of the JSON data item you want to point into, and you want to point to exactly one position in there.  If you need to do something that is closer to a “search” (which might also result in multiple positions), JSONPath gives you more rope.
*(Carsten Bormann)*

As agreed at the IETF 110 meeting the above-mentioned suggestions should be included into the introduction to expand further the distinction between JSON Pointer and our specification.

gregsdennis:
Mildly related to some parts of the discussion in #44.

danielaparker:
@gregsdennis 

> Mildly related to some parts of the discussion in #44.

More than mildly! A Normalized Path Expression as introduced in [Goessner](https://goessner.net/articles/JsonPath/) performs exactly the role of JSONPointer, to uniquely identify a specific value in a JSON document.

bettio:
I think we should clearly encourage users using JSON Pointer when security is a concern and they are just interested into selecting a single document sub-tree (when JSON document structure is known).
This might be a good prescription for the security considerations section.

gregsdennis:
> I would suggest "majorly" rather than "mildly". - @danielaparker 

That issue is about containing the location with the values in the output, but it was hijacked into a discussion about JSON Pointer.

danielaparker:
@gregsdennis wrote:
> 
> That issue is about containing the location with the values in the output, but it was hijacked into a discussion about JSON Pointer.

Fair enough. But it was an interesting discussion, given that JSONPointer and Normalized Path both serve to specify a specific value in a JSON document.

[...]

Issue #117, opened 2021-08-30
String normalization


cburgmer:
When looking up object member names (dot selector or index selector) normalization of strings becomes an issue. See https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization for an introduction on normalization in Unicode. 

As indicated with one test in the comparison project (https://cburgmer.github.io/json-path-comparison/#bracket_notation_with_NFC_path_on_NFD_key) it seems though that there is consensus not to respect equivalent code points. The Raku implementation is the only one that matches the equivalent key, which seems to stem from the Raku runtime already normalizing strings upfront (if my manual tests are to be believed).

Is this something we want to prescribe in a standard, or would we end up with advice possibly contradicting the implementation of JSON parsers? (A quick search turned up http://seriot.ch/parsing_json.php which complains about different implementation in parsers with respect to string normalization.)

gregsdennis:
I believe there's already a convention that bracketed syntax (`[1]` and `['foo']`) is the normalized syntax.

Secondarily, where should these normalized paths be used/enforced?  Surely input should handle any form.  However output (i.e. paths to matches) should be normalized.

danielaparker:
@cburgmer, it's an interesting issue. 

I was interested in how XPath 3.1 handles normalization, discussed [here](https://www.w3.org/TR/xpath-functions-31/#character-terminology), and it appears to have very limited support, remarking that "Unless explicitly stated, the `xs:string` values returned by the functions in this document are not normalized in the sense of [Character Model for the World Wide Web 1.0: Fundamentals](https://www.w3.org/TR/charmod/).

XPath 3.1 does have an [fn:normalize-unicode](https://www.w3.org/TR/xpath-functions-31/#func-normalize-unicode) function, which applies unicode normalization to a string. It has a two argument overload that allows the caller to set the Normalization Form, with five different kinds of normalization. I suppose we could use such a function to compare string values in filter expressions.

As for looking up property values in a JSON object, I think many JSONPath implementations use functions provided by another library, such as `JsonElement.TryGetProperty` in `.Net`. And it seems from your test that they don't generally support normalization.

gregsdennis:
Wait, are we discussing _Unicode_ normalization or _Path_ normalization?  I think my comment may have been off-topic.

cabo:
When discussing normalisation of text in this context, it is useful to have read https://www.w3.org/TR/charmod-norm/

[...]

PR #152, opened 2022-02-01
Normalized Paths and how JSONPath relates to JSON Pointer

glyn:
Specify Normalized Paths
    
Also clarify the relationship of JSONPath in general, and Normalized Paths in particular, to JSON Pointer.
    
The approach for Normal Paths is to minimise escaping. The alternative of maximising escaping was rejected because the results are unreadable. The alternative of escaping all but a particular set of characters was rejected because it would be biassed.
    
Co-authored-by: @cabo 

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/67

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1109,8 +1109,8 @@ Single quotes are used to delimit string member names. This reduces the number o
 need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
 
 The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
-given Normalized Path. Putting this another way, for any two distinct Normalized Paths, there is a JSON document
-which will yield distinct results when the Normalized Paths are applied to the document.
+given Normalized Path. Putting this another way, for any two distinct Normalized Paths, a JSON document exists
+which will yield distinct results when the Normalized Paths are applied to the it.
 
 Certain characters are escaped in one and only one way; all other characters are unescaped.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1110,9 +1110,9 @@ need escaping when Normalized Paths appear as strings (which are delimited with 
 
 The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
 given Normalized Path. Putting this another way, for any two distinct Normalized Paths, a JSON document exists
-which will yield distinct results when the Normalized Paths are applied to the it.
+which will yield distinct results when the Normalized Paths are applied to it.
 
-Certain characters are escaped in one and only one way; all other characters are unescaped.
+Certain characters are escaped, in one and only one way; all other characters are unescaped.
 
 ~~~~ abnf
 normalized-path           = root-selector *(normal-index-selector)
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -141,15 +141,8 @@ format for structured data values.
 JSONPath defines a string syntax for identifying values
 within a JSON value.
 
-JSONPath is not intended as a replacement, but as a more powerful
-companion, to JSON Pointer {{RFC6901}}. [^json-pointer-missing]
-
-[^json-pointer-missing]:
-    Insert reference to section where the relationship is detailed.
-    The purposes of the two syntaxes are different.
-    Pointer is for isolating a single location within a JSON document.
-    Path is a query syntax that can also be used to pull multiple
-    locations.
+JSONPath is not intended as a replacement for, but as a more powerful
+companion to, JSON Pointer {{RFC6901}}. See {{json-pointer}}.
 
 ## Terminology
 
@@ -279,16 +272,13 @@ JSONPath expressions can use the *dot notation*
 $.store.book[0].title
 ~~~~
 
-or the *bracket notation*
+or the more general *bracket notation*
 
 ~~~~
 $['store']['book'][0]['title']
 ~~~~
 
-to build paths that are input to a JSONPath processor.
-Bracket notation is more general than dot notation and can serve as a
-canonical form (for instance, when a JSONPath processor uses JSONPath
-expressions as output paths).
+to build paths that are input to a JSONPath implementation.
 
 JSONPath allows the wildcard symbol `*` to select any member of an
 object or any element of an array ({{wildcard}}).
@@ -450,7 +440,7 @@ the argument.
 The nodelist resulting from the last selector is presented as the
 result of the query; depending on the specific API, it might be
 presented as an array of the JSON values at the nodes, an array of
-Output Paths referencing the nodes, or both — or some other
+Normalized Paths referencing the nodes, or both — or some other
 representation as desired by the implementation.
 Note that the API must be capable of presenting an empty nodelist as
 the result of the query.
@@ -680,8 +670,8 @@ escapable           = ( %x62 / %x66 / %x6E / %x72 / %x74 / ; \b \f \n \r \t
                           ; n /         ;  LF line feed U+000A
                           ; f /         ;  FF form feed U+000C
                           ; r /         ;  CR carriage return U+000D
-                          "/" /          ;  /  slash (solidus)
-                          "\" /          ;  \  backslash (reverse solidus)
+                          "/" /          ;  /  slash (solidus) U+002F
+                          "\" /          ;  \  backslash (reverse solidus) U+005C
                           (%x75 hexchar) ;  uXXXX      U+XXXX
                       )
 
@@ -1104,6 +1094,57 @@ Some examples:
 | `[3,4,5]` | `$[?index(@)==2]`<br>`$[?index(@)==17]` | `[5]`<br>`[]` | Select array element |
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
+
+## Normalized Paths
+
+A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node. For example,
+the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`. For a given JSON document, there is a one to one correspondence between the document's
+nodes and the Normalized Paths that identify these nodes.
+
+An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
+instead of, or in addition to, the values identified by these paths.
+
+Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
+Single quotes are used to delimit string member names. This reduces the number of characters that
+need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
+
+The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
+given Normalized Path. Putting this another way, for any two distinct Normalized Paths, there is a JSON document
+which will yield distinct results when the Normalized Paths are applied to the document.
+
+Certain characters are escaped in one and only one way; all other characters are unescaped.
+
+~~~~ abnf
+normalized-path           = root-selector *(normal-index-selector)
+normal-index-selector     = "[" (normal-quoted-member-name / normal-element-index) "]"
+normal-quoted-member-name = %x27 *normal-single-quoted %x27 ; 'string'
+normal-single-quoted      = normal-unescaped /
+                            ESC normal-escapable
+normal-unescaped          = %x20-26 /                       ; omit control codes
+                            %x28-5B /                       ; omit '
+                            %x5D-10FFFF                     ; omit \
+normal-escapable          = ( %x62 / %x66 / %x6E / %x72 / %x74 / ; \b \f \n \r \t
+                                ; b /         ;  BS backspace U+0008
+                                ; t /         ;  HT horizontal tab U+0009
+                                ; n /         ;  LF line feed U+000A
+                                ; f /         ;  FF form feed U+000C
+                                ; r /         ;  CR carriage return U+000D
+                                "'" /         ;  ' apostrophe U+0027
+                                "\" /         ;  \ backslash (reverse solidus) U+005C
+                                (%x75 normal-hexchar) ;  certain values u00XX U+00XX
+                            )
+normal-hexchar            = "0" "0"
+                            (
+                              ("0" %x30-37) / ; "00"-"07"
+                              ("0" %x62) /    ; "0b"      ; omit U+0008-U+000A
+                              ("0" %x65-66) /  ; "0e"-"0f" ; omit U+000C-U+000D
+                              ("1" normal-HEXDIG)
+                            )
+normal-HEXDIG             = DIGIT / %x61-66   ; "0"-"9", "a"-"f"
+normal-element-index      = "0" / (DIGIT1 *DIGIT) ; non-negative decimal integer
+~~~~
+
 # IANA Considerations {#IANA}
 
 ##  Registration of Media Type application/jsonpath
@@ -1310,6 +1351,28 @@ JSONPath:
 * With JSONPath, square brackets operate on the *object* or *array*
   addressed by the previous path fragment. Array indices always start
   at 0.
+
+# JSON Pointer
+
+This appendix is informative.
+
+JSONPath is not intended as a replacement for, but as a more powerful
+companion to, JSON Pointer {{RFC6901}}. The purposes of the two standards
+are different.
+
+JSON Pointer is for identifying a single value within a JSON document whose
+structure is known.
+
+JSONPath can identify a single value within a JSON document, for example by
+using a Normalized Path. But JSONPath is also a query syntax that can be used
+to search for and extract multiple values from JSON documents whose structure
+is known only in a general way.
+
+A Normalized JSONPath can be converted into a JSON Pointer by converting the syntax,
+without knowledge of any JSON document. The inverse is not generally true: a numeric
+path component in a JSON Pointer may identify a member of a JSON object or may index an array.
+For conversion to a JSONPath query, knowledge of the structure of the JSON document is
+needed to distinguish these cases.
 
 # Acknowledgements
 {: numbered="no"}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1102,8 +1102,7 @@ the JSONPath expression `$.book[?(@.price<10)]` could select two values with Nor
 `$['book'][3]` and `$['book'][5]`. For a given JSON document, there is a one to one correspondence between the document's
 nodes and the Normalized Paths that identify these nodes.
 
-An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
-instead of, or in addition to, the values identified by these paths.
+A JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
 
 Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
 Single quotes are used to delimit string member names. This reduces the number of characters that


draft-ietf-jsonpath-base.md
```diff
@@ -227,7 +220,7 @@ Nodelist:
   While this list can be represented in JSON, e.g. as an array, the
   nodelist is an abstract concept unrelated to JSON values.
 
-Normalized Path:
```
goessner, line 230: I definitely prefer 'Normalized Path'.

> "Normalization is the process of organizing a database to reduce redundancy and improve data integrity.
Normalization also simplifies the database design so that it achieves the optimal structure composed of atomic elements (i.e. elements that cannot be broken down into smaller parts)." 

https://database.guide/what-is-normalization/

>  In this context, a canonical form is a representation such that every object has a unique representation.

https://en.wikipedia.org/wiki/Canonical_form#Computing

So 'normalized path' in its canonical form is much more concrete than merely 'output path' . Its syntax definition is:

```
normalized-path = root-selector *("[" (quoted-member-name / positiv-element-index) "]")

positive-element-index = ( "0" / (DIGIT1 *DIGIT) )
```

goessner: A 'normalized path' usually is an 'output path'. It is frequently used as 'json-path' for revisiting previous selections (post processing ?) and is understood natively as accessor at least in JavaScript.

glyn: Fair enough. I'm going to try using normalized path throughout. Please see the second commit.

draft-ietf-jsonpath-base.md
```diff
@@ -231,6 +224,8 @@ Normalized Path:
 : A simple form of JSONPath expression that identifies a node by
   providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
+  Note: Normalized Paths are not canonical: two distinct Normalized Paths
```
cabo, line 227: This cries for an example.
It probably would be useful to have a place where normalized paths are discussed that is outside the definitions.

draft-ietf-jsonpath-base.md
```diff
@@ -285,10 +280,10 @@ or the *bracket notation*
...
 ~~~~
 
-to build paths that are input to a JSONPath processor.
-Bracket notation is more general than dot notation and can serve as a
-canonical form (for instance, when a JSONPath processor uses JSONPath
-expressions as output paths).
+to build paths that are input to a JSONPath implementation.
+Bracket notation is more general than dot notation and can be used to construct
+Normalized Paths (for instance, when a JSONPath implementation outputs Normalized Paths
+instead of the values identified by these paths).
```
cabo, line 286: not necessarily "instead of".

```suggestion
instead of or in addition to the values identified by these paths).
```

draft-ietf-jsonpath-base.md
```diff
@@ -1311,6 +1306,22 @@ JSONPath:
...
+are different.
+
+JSON Pointer is for identifying a single value within a JSON document whose
+structure is known.
+
+JSONPath can identify a single value within a JSON document, for example by
+using a Normalized Path. But JSONPath is also a query syntax that can be used
+to search for and extract multiple values from JSON documents whose structure
+is partially known.
+
```
cabo, line 1324: I think we should say something about conversion between JSON Pointer and JSONPath.

```
A Normalized JSONPath can be converted into a JSON Pointer by converting the syntax.
The inverse is not generally possible as a path component in a JSON Pointer that looks like a number may identify a member of a JSON object or may index an array.  For this conversion, knowledge of the structure of the references JSON value is needed.

File: draft-ietf-jsonpath-base.md [hunk 5]
```diff
@@ -1311,6 +1306,22 @@ JSONPath:
...
+companion to, JSON Pointer {{RFC6901}}. The purposes of the two standards
+are different.
+
+JSON Pointer is for identifying a single value within a JSON document whose
+structure is known.
+
+JSONPath can identify a single value within a JSON document, for example by
+using a Normalized Path. But JSONPath is also a query syntax that can be used
+to search for and extract multiple values from JSON documents whose structure
+is partially known.
```
cabo, line 1323: partially sounds like one half is (fully) known and the other half is completely unknown.

```suggestion
is known only in a general way.
```

draft-ietf-jsonpath-base.md
glyn: Ok, I'll bite.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,23 @@ Some examples:
...
+An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
+instead of, or in addition to, the values identified by these paths.
+
+Since bracket notation is more general than dot notation, it is convenient in constructing
+Normalized Paths. However, Normalized Paths are not required to use bracket notation when dot notation is appropriate.
+
+Normalized Paths are not necessarily canonical since two distinct Normalized Paths may identify the same node.
+For example, `$.a` and `$['a']` are equivalent Normalized Paths. With certain JSON documents, `$[-1]` and `$[4]` may
+be equivalent Normalized Paths.
+
```
goessner, line 1113: Here
```
$['a']
$[4]
```
qualifies for a normalized Path only.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,23 @@ Some examples:
 | `[3,4,5]` | `$[?index(@)==2]`<br>`$[?index(@)==17]` | `[5]`<br>`[]` | Select array element |
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
+
+## Normalized Paths
+
+A Normalized Path identifies a node by providing a query that results in exactly that node. For example,
+the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$.book[3]` and `$.book[5]`.
+
```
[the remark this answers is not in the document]
cabo: `$['book'][3]` and `$["book"][3]` address the same node, so the Normalized Path is canonical in structure only, not in syntax.

draft-ietf-jsonpath-base.md
goessner: Please see ...
```
normalized-path = root-selector *("[" (quoted-member-name / positiv-element-index) "]")

positive-element-index = ( "0" / (DIGIT1 *DIGIT) )
```

from above ...

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
goessner: No ... even here we should state in the spec, that `$['book'][3]` is the one and only Normalized Path. Reason here is to reuse normalized pathes as strings in JSON documents without necessity of escaping ... exactly the same reason, why we support `'...'` additionally  to JSON's `"..."`.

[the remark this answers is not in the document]
cabo: I like the idea of defining the preferred (canonical) syntax.  We still need to escape backslashes and decide how certain characters are shown (e.g., \r vs. \u000d, no \/ etc.) -- should not be too hard, but needs to be done.

draft-ietf-jsonpath-base.md
cabo: s/positiv/unsigned/  (0 is allowed, but not positive)

goessner: ... I wanted to avoid `non-negative-element-index` ... now I like `unsigned` ...

draft-ietf-jsonpath-base.md
goessner: hmm ... `quoted-member-name` must be replaced by `single-quoted-member-name` above.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: I'm not sure of the goal here, or what the WG consensus is likely to be, especially since Normalized Paths don't currently feature much in the spec. Should we take this discussion to the mailing list?

draft-ietf-jsonpath-base.md
cabo: > hmm ... `quoted-member-name` must be replaced by `single-quoted-member-name` above.

Indeed, but that is not everything we need to go fully canonical (https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/152#discussion_r796785776).

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
goessner: I don't know, what the goal of a mailing list discussion would be.

IIRC normalized pathes as an alternative result (to values) of queries never have been questionable. It is also the only way to verify, that identical queries lead to identical results ... wouldn't be possible with ambiguous node values only ... it's key for tests!

But if you want ... take this discussion to the mailing list?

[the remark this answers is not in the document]
glyn: > I think this is a tangent, but a useful one. Should there be a single canonical form for a Normalized Path?

Yes, that's the question. But not everyone who can usefully contribute will be tuned in to this PR.

> 
> (The discussion about "Normalized Path" vs. "Output Path" is a bit confused, like a discussion whether somebody is an "MIT Alumnus" or "CEO of acme corp." -- these are just orthogonal, one is what it is, one is where it is used.)

I understand, but I don't see the need for "Output Path" as a separate term. We can simply talk about the output from an implementation etc.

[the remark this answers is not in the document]
cabo: > I understand, but I don't see the need for "Output Path" as a separate term. We can simply talk about the output from an implementation etc.

I don't see a need to *introduce* this term either, but we can say that the output of a query can be values and/or paths, and these are then of course output values and output paths without further definition or capitalization.

Let's open the Normalized Path determinism ("canonicalization") issue on the mailing list, though; we can finish the rest of this PR while this is being discussed.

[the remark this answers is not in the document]
goessner: 'Output Path' is not needed. Result of a JSONPath query is a list of nodes formatted as canonical Normalized Pathes.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,23 @@ Some examples:
...
+An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
+instead of, or in addition to, the values identified by these paths.
+
+Since bracket notation is more general than dot notation, it is convenient in constructing
+Normalized Paths. However, Normalized Paths are not required to use bracket notation when dot notation is appropriate.
+
+Normalized Paths are not necessarily canonical since two distinct Normalized Paths may identify the same node.
+For example, `$.a` and `$['a']` are equivalent Normalized Paths. With certain JSON documents, `$[-1]` and `$[4]` may
+be equivalent Normalized Paths.
```
gregsdennis, line 1112: I'm not sure I agree with `$[-1]` being considered normalized.  In my opinion, a normalized path should be one that distinctly identifies a node regardless of the input JSON.  `$[-1]` will identify different nodes depending on the input.

draft-ietf-jsonpath-base.md
danielaparker: > ... I wanted to avoid `non-negative-element-index` ... now I like `unsigned-element-index` ...

I think `non-negative-element-index` is better. The domain of indices is signed integers. That domain is restricted to non-negative integers for normalized paths, it's not that they've suddenly become unsigned.

draft-ietf-jsonpath-base.md
glyn: > I'm not sure I agree with `$[-1]` being considered normalized. In my opinion, a normalized path should be one that distinctly identifies a node regardless of the input JSON. `$[-1]` will identify different nodes depending on the input.

I have some reservations too, but I was trying to make clear one consequence of _not_ insisting on a canonical form of Normalized Path. However, I think Normalized Paths only make sense in the context of a known/given JSON document. If the document varies, then the node identified by _any_ path can be different (e.g. just replace the identified node in one document with a distinct node).

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: @cabo @goessner "canonical Normalized Path" is a mouthful. Why not simply "Canonical Path"?

draft-ietf-jsonpath-base.md
glyn: > Please see ...
> 
> ```
> normalized-path = root-selector *("[" (quoted-member-name / positiv-element-index) "]")
> 
> positive-element-index = ( "0" / (DIGIT1 *DIGIT) )
> ```
> 
> from above ...

Unless the mailing list pushes back, I plan to adopt this prescriptive approach.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: Normalized Path.

(Don't use the C word.)

draft-ietf-jsonpath-base.md
cabo: > > ... I wanted to avoid `non-negative-element-index` ... now I like `unsigned-element-index` ...
> 
> I think `non-negative-element-index` is better. The domain of indices is signed integers. That domain is restricted to non-negative integers for normalized paths, it's not that they've suddenly become unsigned.

I get this discussion often.  Non-negative is pedantic.  The range of actual array indexes actually is unsigned integer, so why not just say that.  (We also allow indexing with negative integers, but that means something different.)

danielaparker: > > I think `non-negative-element-index` is better. The domain of indices is signed integers. That domain is restricted to non-negative integers for normalized paths, it's not that they've suddenly become unsigned.
> 
> I get this discussion often. Non-negative is pedantic. The range of actual array indexes actually is unsigned integer, so why not just say that. (We also allow indexing with negative integers, but that means something different.)

In this context, "pedantic" is just another name for "clear". It's unclear what it means to mix "unsigned" and "negative" integers on the same domain. It doesn't translate well into programming languages.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: For the record, what's wrong with "canonical"?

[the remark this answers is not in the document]
cabo: Generally evokes enormous amounts of pain in people who have ever tried to do c14n.

Also, unnecessary here, as we already use the term "normalized".

draft-ietf-jsonpath-base.md
glyn: > Please see ...
> 
> ```
> normalized-path = root-selector *("[" (quoted-member-name / positiv-element-index) "]")
> 
> positive-element-index = ( "0" / (DIGIT1 *DIGIT) )
> ```
> 
> from above ...

`quoted-member-name` allows two sort of variability: single or double quoted strings and escapes or unescaped characters. Should we leave these open or attempt to tie them down so there is one and only one representation of any Normalized Path?

cabo: I sensed the direction of the discussion that we wanted to have deterministic Normalized Path results, i.e. two implementations would generate the same output paths for the same JSON value and query.  That would need us to pick the kind of quoting and the details of the escaping.  Deterministic results generally help in testing.  We might still decide that this is too much work.

goessner: Please see my correction above ...

> hmm ... quoted-member-name must be replaced by single-quoted-member-name above.

so it should read somewhat like ...
```
normalized-path = root-selector *("[" *(%x27 *single-quoted %x27) / positiv-element-index) "]")

unsigned-element-index = ( "0" / (DIGIT1 *DIGIT) )
```

goessner: @cabo: I think, the benefit of clean testing justifies the effort here. However, I cannot really judge the effort involved in documenting escaping.

glyn: Ok, I suggest we mandate single quotes (to align with JSON) and escapes (except for \/ and \uxxxx, instead of which / and the unicode character for U+xxxx must be used, respectively). Is that a reasonable starting position?

goessner: ... but, isn't escaping already clearly documented with `single-quoted` already in the current draft ... ?

glyn: > ... but, isn't escaping already clearly documented with `single-quoted` already in the current draft ... ?

No, because it permits at least two representations of any given character. I think we should try to nail this down so there is a unique representation.

cabo: To be clear, JSON uses double quotes, so we are using single quotes to evade further JSON escaping, except where we need to escape.  We can't use the actual characters for \u0000 to \u001f, so we should use \b\t\n\f\r or the \u00xx for the other 11 cases.  We also need `\\` and `\'`, but we don't need `\/`.  I'm sure there will be people who would want to escape anything including and above \u00ff as well.  Oh, and we need to specify the case of the hexchars.

glyn: Gotcha.

> I'm sure there will be people who would want to escape anything including and above \u00ff as well.

I guess escaping this range is safer, even if it can be criticised as ASCII-centric. It wouldn't surprise me if mandating unescaped characters for that range caused problems in some older programming languages.

draft-ietf-jsonpath-base.md
glyn: `$[-1]` is now not allowed as part of a Normalized Path.

glyn:
@cabo @goessner I've pushed a first attempt at tying down Normalized Path syntax. The approach is to use \uXXXX except for a specific set of codepoints which can appear unescaped.

draft-ietf-jsonpath-base.md
goessner: I was thinking a while about that escape thing. There is a strong relation to string comparison, see ...

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/117#issuecomment-909356265

Consensus here obviously was *code point* i.e. *byte string* comparison.

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/117#issuecomment-1014932216

In this case here we need not to deal with arbitrary text content, but rather with member names. This should be less crtitical then. 

So, if restricting to code point equality here again, is all in all consistent.

I tried to collect some common scenarios:

1. Query results are reused on the same JSON source document for revisiting nodes with the purpose of ...
2. Normalized pathes are applied to similar JSON documents (output from a single application or into a standardised JSON format) by different users with different implementations on different platforms.
3. Compare query results from different implementations for test purposes.

Discussion:
1. Should be mostly applied in practise (estimated). Is uncritical regarding escapes, since JSON document, implementation, platform remains the same.
2. Is the most critical from my point of view. But the probability of dealing with escaped characters in this scenario is considered rather low.
3. For developers or advanced users only. Should be uncritical regarding escapes.

Are there other relevant scenarios, that I might be missing here?

So Glyn's approach or code point equality seems to be a pragmatic solution here.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,34 @@ Some examples:
...
+normal-quoted-member-name = %x27 *normal-single-quoted %x27 ; 'string'
+normal-single-quoted      = normal-unescaped /
+                            ESC normal-escapable
+normal-unescaped          = %x20-26 /                       ; omit control codes
+                            %x28-5B /                       ; omit '
+                            %x5D-7E                         ; omit \
+                            %xA0-FF                         ; omit control codes
+normal-escapable           = (%x75 hexchar) ;  uXXXX U+XXXX ; TODO: exclude normal-unescaped codepoints
+normal-element-index       = "0" / (DIGIT1 *DIGIT)          ; non-negative decimal integer
+~~~~
```
goessner, line 1123: ... I like that ...

cabo: So we'd have to use \u0027 for '?

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,34 @@ Some examples:
...
+~~~~ abnf
+normalized-path           = root-selector *(S normal-index-selector)
+normal-index-selector     = "[" S (normal-quoted-member-name / normal-element-index) S "]"
+normal-quoted-member-name = %x27 *normal-single-quoted %x27 ; 'string'
+normal-single-quoted      = normal-unescaped /
+                            ESC normal-escapable
+normal-unescaped          = %x20-26 /                       ; omit control codes
+                            %x28-5B /                       ; omit '
+                            %x5D-7E                         ; omit \
+                            %xA0-FF                         ; omit control codes
```
cabo, line 1120: We probably want to exclude U+00FF (Rubout).

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,34 @@ Some examples:
 | `[3,4,5]` | `$[?index(@)==2]`<br>`$[?index(@)==17]` | `[5]`<br>`[]` | Select array element |
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
+
+## Normalized Paths
+
+A Normalized Path identifies a node by providing a query that results in exactly that node. For example,
+the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$.book[3]` and `$.book[5]`.
```
goessner, line 1102: ... `$['book'][3]` and  `$['book'][5]` ...

draft-ietf-jsonpath-base.md
cabo, line 1120: I'm not sure how basic latin is better than other beyond-ASCII parts of Unicode.

draft-ietf-jsonpath-base.md
goessner: Wouldn't we also need to omit `"` also, since we want to use normalized pathes - as they are - as potential text in JSON documents ?

glyn: > So we'd have to use \u0027 for '?

Yes.

glyn: > Wouldn't we also need to omit `"` also, since we want to use normalized pathes - as they are - as potential text in JSON documents ?

The requirement to use Normalized Paths as text in JSON documents is new to me. So far, I have only considered a JSONPath implementation outputting Normalized Paths instead of, or in addition to, values - end of story. What use case do you have in mind?

draft-ietf-jsonpath-base.md
glyn: Nice catch. I'll fix that.

draft-ietf-jsonpath-base.md
goessner: As I already commented here ...

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/152/files#r796782971

charme of using `'` delimiters within JSONPath queries and resulting pathes is the possibility of having both as unescaped values in JSON documents. Use them for referencing locations in the same or other JSON documents in JSONPath context or simply document query results in testing contexts in JSON, whatever ... thus JSONPath and JSON harmonize perfectly together.

glyn: Great. Let me build that in. Thanks!

goessner: example ... very unlikely ...

```json
{
     "j\"org" : "name of Jörg"
}
```

JSONPath result of `$['j"org']` or something similiar might / will be  `$['j"org']`. Putting them in 
```json
{
     "JSONPath_result": "$['j"org']"
}
```
will collide with JSON's  `"`.

cabo: Well, putting something into a JSON string will take care of that.

cabo: The charm of using `'` is that no escaping is needed, so the path stays readable in a JSON context.

glyn: > As I already commented here ...
> 
> https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/152/files#r796782971
> 
> charme of using `'` delimiters within JSONPath queries and resulting pathes is the possibility of having both as unescaped values in JSON documents. Use them for referencing locations in the same or other JSON documents in JSONPath context or simply document query results in testing contexts in JSON, whatever ... thus JSONPath and JSON harmonize perfectly together.

@goessner Please see the latest commit which addresses your comment.

draft-ietf-jsonpath-base.md
cabo: I don't know how to explain that we need to escape Ё, ⽔, Č, ⨸, but not ÿ or ÷ or ¶.

draft-ietf-jsonpath-base.md
```diff
@@ -1108,17 +1108,23 @@ Since bracket notation is more general than dot notation, it is used to construc
 the syntax of Normalized Paths is restricted so that there is one and only one way of representing any
 given Normalized Path.
 
+Single quotes are used to delimit string member names. This makes it possible to use Normalized Paths
+as strings in JSON documents (delimited with double quotes) without needing to escape the contents of
```
cabo, line 1112: That is not true.  You still need to escape some of the contents of the strings when putting them into JSON; you just get `'` for free (and stay reasonably readable that way).

draft-ietf-jsonpath-base.md
```diff
@@ -1108,17 +1108,23 @@ Since bracket notation is more general than dot notation, it is used to construc
...
 normalized-path           = root-selector *(S normal-index-selector)
 normal-index-selector     = "[" S (normal-quoted-member-name / normal-element-index) S "]"
 normal-quoted-member-name = %x27 *normal-single-quoted %x27 ; 'string'
 normal-single-quoted      = normal-unescaped /
                             ESC normal-escapable
-normal-unescaped          = %x20-26 /                       ; omit control codes
+normal-unescaped          = %x20-21 /                       ; omit control codes
+                            %x23-26 /                       ; omit "
                             %x28-5B /                       ; omit '
                             %x5D-7E                         ; omit \
```
cabo, line 1124: I would comment on the missing `/`, but I think we'll need to get rid of the following line anyway (or change it considerably).

draft-ietf-jsonpath-base.md
glyn: Is it simply a compression chosen because of the high frequency of latin characters in extant JSON?

cabo: But then you'd need to include all Latin characters.   I'm sure Mališa Vučinić would agree :-)
And what is better about Latin than Greek, Cyrillic, ...

draft-ietf-jsonpath-base.md
glyn: What's interesting about `/`?

cabo: There is none in the current ABNF between %x5D-7E and %xA0-FF :-)

draft-ietf-jsonpath-base.md
cabo: `"`, I think.

draft-ietf-jsonpath-base.md
glyn: What would you propose? Escape almost all characters?

glyn:
@glyn @goessner I don't like the alternatives of picking certain some characters which are not escaped or of escaping everything. Would it actually be neater to allow some freedom to escape or not escape most characters? I know this would create a bit more code where such paths are processed, e.g. in testcases testing for equivalence of two normalized paths, but that seems like a lesser evil.

draft-ietf-jsonpath-base.md
cabo: The reason I didn't propose anything yet is that I have no idea.  

Escaping *something* seems in line with https://nvd.nist.gov/vuln/detail/CVE-2021-42574 -- if we assume Normalized Paths are for humans to look at.

If this is all about programs, there is no point in escaping anything (except what is needed to stay similar enough to JSON).

timbray:
I strongly lean to the absolute minimum of required escaping.  Related: These things are going to be embedded in Java/Go/Python/whatever code - or JSON (!) - so there's already going to be some escaping at a higher level.  Generally speaking, almost everything handles UTF-8 properly these days, including non-BMP characters, if only because of the requirement to process emojis correctly.  (My unit tests for some text-intensive code I'm working on now are full of emojis.)

draft-ietf-jsonpath-base.md
goessner: Yes of course that must be done manually ... but it is very unlikely to meet member names in any JSON document with embedded `"` (bad behavior).

goessner: what about pragmatically reusing the same ABNF, that rfc8259 uses for member names ... adding unescaped `'` of course ?

cabo: I think we already do...

```
      member = string name-separator value
```

where "string" is called "string-literal".
RFC 8259 hasn't done the work to avoid unpaired surrogates, so going back to their definition of "string" would be a regression.

But maybe I don't understand what you are trying to do.

goessner: oh ... I would like to mention the fact somewhere, that normalized pathes should be naturally the most performant queries in any implementation, compared with other - more complex - JSONPath queries leading to the same result node.

goessner: > But maybe I don't understand what you are trying to do.

you did ... thanks for the clarification.

timbray: You're correct that 8259 doesn't rule out naked surrogates and suchlike because the consensus was we couldn't retroactively ban things that were previously JSON per previous specifications. That's what I-JSON is for.  So the question is, should jsonpath member names conform to JSON or I-JSON?

cabo: I-JSON.

cabo: 8259 is schizophrenic: It requires UTF-8 for unescaped, but doesn't exclude unpaired surrogates for escaped.  That clearly would have been fixed if we had been a bit more audacious at the time.

timbray: OK, whether jsonpath should be more restrictive than JSON in what names it allows is an issue we're going to have to take up on its own.

gregsdennis:
I concur with @timbray on this.  We should only escape that which _we_ need to escape. Let other formats provide secondary escaping for their own needs. Trying to accommodate other formats is just going to overcomplicate our syntax.

draft-ietf-jsonpath-base.md
glyn: Ah, gotcha. I was so focussed on the character ranges, I couldn't see the missing `/`. Fixed.

glyn:
I'm going to let this PR rest/fester over the weekend. Meanwhile, if anyone would like to submit a minimal escaping alternative, reusing the JSON Pointer text here, please be my guest.

draft-ietf-jsonpath-base.md
cabo: As a WG chair, you may have to say something like this.

From a technical point of view, I find the suggestion that we could do anything that could be understood to lend additional credibility to the Zombie that Section 8.2 of RFC 8259 is, simply untenable. That isn't JSON, it is crimes of the past that wander the unlit cemeteries waiting to finally be buried.

glyn:
I pushed a commit to minimise escaping.

glyn:
In line with @timbray's and @gregsdennis's comments above, I pushed a further commit to clarify the need for escaping when a Normalized Path is embedded in a JSON string.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,62 @@ Some examples:
...
+A Normalized Path identifies a node by providing a query that results in exactly that node. For example,
+the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`.
+
+An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
+instead of, or in addition to, the values identified by these paths.
+
+Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
+
+Single quotes are used to delimit string member names. This reduces the number of characters which
```
cabo, line 1109:
```suggestion
Single quotes are used to delimit string member names. This reduces the number of characters that
```

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,62 @@ Some examples:
...
+need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
+
+The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
+given Normalized Path. Certain characters are escaped in one and only one way; all other
+characters are unescaped.
+
+~~~~ abnf
+normalized-path           = root-selector *(S normal-index-selector)
```
cabo, line 1117: Do we actually want to allow spaces here?
That makes the output non-deterministic again.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,62 @@ Some examples:
...
+                                ; n /         ;  LF line feed U+000A
+                                ; f /         ;  FF form feed U+000C
+                                ; r /         ;  CR carriage return U+000D
+                                "'" /         ;  ' apostrophe U+0027
+                                "\" /         ;  \ backslash (reverse solidus) U+005C
+                                (%x75 normal-hexchar) ;  certain values u00XX U+00XX
+                            )
+normal-hexchar            = "0" "0"
+                            (
+                              ("0" "0") /
```
cabo, line 1137: This could be shortened as ("0" %x30-37)

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,62 @@ Some examples:
...
+                              ("0" "1") /
+                              ("0" "2") /
+                              ("0" "3") /
+                              ("0" "4") /
+                              ("0" "5") /
+                              ("0" "6") /
+                              ("0" "7") /
+                              ("0" "B") / ; omit U+0008-U+000A
```
cabo, line 1145: These are case-insensitive.  We should decide on lower case or upper case.  Since we'll need to do a %x then, we can also %x65-66 again.

draft-ietf-jsonpath-base.md
```diff
@@ -1311,6 +1357,27 @@ JSONPath:
...
+
+JSONPath can identify a single value within a JSON document, for example by
+using a Normalized Path. But JSONPath is also a query syntax that can be used
+to search for and extract multiple values from JSON documents whose structure
+is known only in a general way.
+
+A Normalized JSONPath can be converted into a JSON Pointer by converting the syntax,
+without knowledge of any JSON document. The inverse is not generally true: a numeric
+path component in a JSON Pointer may identify a member of a JSON object or may index an array.
+For this conversion, knowledge of the structure of the JSON document is needed.
```
cabo, line 1379:
```suggestion
For conversion to a JSONPath query, knowledge of the structure of the JSON document is needed to distinguish these cases.
```

draft-ietf-jsonpath-base.md
glyn: Nope. Good catch.

draft-ietf-jsonpath-base.md
glyn: Agreed, but I think the longer form may be easier to read, so I'm going to punt on this for now.

draft-ietf-jsonpath-base.md
goessner: I do agree to **not** allow them.

draft-ietf-jsonpath-base.md
```diff
@@ -141,15 +141,8 @@ format for structured data values.
...
-companion, to JSON Pointer {{RFC6901}}. [^json-pointer-missing]
-
-[^json-pointer-missing]:
-    Insert reference to section where the relationship is detailed.
-    The purposes of the two syntaxes are different.
-    Pointer is for isolating a single location within a JSON document.
-    Path is a query syntax that can also be used to pull multiple
-    locations.
+JSONPath is not intended as a replacement for, but as a more powerful
+companion to, JSON Pointer {{RFC6901}}. See {{json-pointer}}.
```
goessner, line 146: Overall a great approach ... thanks.

draft-ietf-jsonpath-base.md
glyn: Agreed. I'll plump for lower case. (Also, the use of `\b` etc. rather than `\B` is a mild precedent.)

draft-ietf-jsonpath-base.md
glyn: I'll include this in the next commit, thanks.

draft-ietf-jsonpath-base.md
```diff
@@ -1097,21 +1097,22 @@ Some examples:
 
 ## Normalized Paths
 
-A Normalized Path identifies a node by providing a query that results in exactly that node. For example,
+A Normalized Path is a JSONPath which identifies a node by providing a query that results in exactly that node. For example,
```
cabo, line 1100:
```suggestion
A Normalized Path is a JSONPath that identifies a node by providing a query that results in exactly that node. For example,
```

glyn: Ok. Must get the hang of writing English some time.

glyn:
I took the liberty of squashing the 17 or so commits to make a readable commit log.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,56 @@ Some examples:
 | `[3,4,5]` | `$[?index(@)==2]`<br>`$[?index(@)==17]` | `[5]`<br>`[]` | Select array element |
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
+
+## Normalized Paths
+
+A Normalized Path is a JSONPath that identifies a node by providing a query that results in exactly that node. For example,
```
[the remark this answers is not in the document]
glyn: Hmmm. Usually the word "unique" (and its variants) introduces confusion into specs. The questions it raises are:

* What thing is unique?
* What is the set of things among which the thing is unique?

In this context, "uniquely identifies a node" could be taken a couple of ways:

1. A Normalized Path is the only JSONPath (with certain characteristics) that identifies a particular node. This is true.
2. A Normalized Path identifies a unique node. This is not true if we are allowed to vary the JSON document.

As for "eineindeutig", I understand this to refer to the mathematical concept of a one-one correspondence. For a given JSON document, there is indeed a one-one correspondence between its nodes and the Normalized Paths that identify these nodes. But the above proposal doesn't talk about a given JSON document, in which case the one-one correspondence doesn't make much sense.

I think we need an informal definition of Normalized Path so readers have a chance of engaging. We could follow it up with some more text along the lines of:

> For a given JSON document,  there is a one to one correspondence between the document's nodes and the Normalized Paths that identify these nodes.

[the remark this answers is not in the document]
goessner: Yes ... this one to one thing is perfect ...

thanks again

goessner (approved):

timbray:
BTW, the Amazon States Language, for which I was design lead, has the notion of a “Reference Path” which I think (?) means the same thing.  See https://states-language.net/#ref-paths

No, I'm not suggesting adopting that name or anything, just thought a look at the descriptive language might be helpful in crafting ours.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: Adopted. Thanks.

cabo:
The squashing/force-pushing makes it a bit laborious to see what you changed in your recent commit.  If you want to squash, maybe do it right before merging?

glyn:
> The squashing/force-pushing makes it a bit laborious to see what you changed in your recent commit. If you want to squash, maybe do it right before merging?

Apologies. I overlooked that.

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,57 @@ Some examples:
...
+
+## Normalized Paths
+
+A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node. For example,
+the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`. For a given JSON document, there is a one to one correspondence between the document's
+nodes and the Normalized Paths that identify these nodes.
+
+An example application of Normalized Paths is a JSONPath implementation which outputs Normalized Paths
+instead of, or in addition to, the values identified by these paths.
```
goessner, line 1106: A JSONPath implementation may output Normalized Paths
instead of, or in addition to, the values identified by these paths

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,57 @@ Some examples:
...
+
+Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
+Single quotes are used to delimit string member names. This reduces the number of characters that
+need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
+
+The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
+given Normalized Path. Putting this another way, for any two distinct Normalized Paths, there is a JSON document
+which will yield distinct results when the Normalized Paths are applied to the document.
```
goessner, line 1114: The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
given node.

goessner: Putting this another way, any two distinct Normalized Paths will yield distinct results when applied to the any document.
...
Is this true? Both pathes may return '5' or 'undefined'.

draft-ietf-jsonpath-base.md
glyn: Adopted. Thanks.

draft-ietf-jsonpath-base.md
glyn: I deliberately phrased that sentence that way to avoid dealing with the case where a Normalized Path doesn't identify a node in a particular document.

May I suggest we merge the PR as-is and take up any further improvements separately as I think it represents the current consensus and we are mainly talking about editorial and descriptive tweaks.

cabo: (Ok, maybe s/is/exists/ makes this clearer as a mathematically phrased statement.)

draft-ietf-jsonpath-base.md
```diff
@@ -1104,6 +1094,56 @@ Some examples:
...
+nodes and the Normalized Paths that identify these nodes.
+
+A JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+
+Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
+Single quotes are used to delimit string member names. This reduces the number of characters that
+need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
+
+The syntax of Normalized Paths is restricted so that there is one and only one way of representing any
+given Normalized Path. Putting this another way, for any two distinct Normalized Paths, there is a JSON document
```
cabo, line 1112:
```suggestion
given Normalized Path. Putting this another way, for any two distinct Normalized Paths, a JSON document exists such that
it will yield distinct results when the Normalized Paths are applied to it.
```

glyn: I pushed a commit with a variation of that suggestion.

draft-ietf-jsonpath-base.md
glyn: Yeah, that will clue in anyone familiar with predicate logic.

Issue #93, opened 2021-03-29
"Index-selection query" in bracket notation


gregsdennis:
This discussion branches off of #88 primarily but the idea has also been discussed elsewhere (#17 & #57).

This issue covers the parenthetical bracket notation without the `?`, e.g. `$[(<expr>)]`.  Specifically, what is `<expr>` and what does this syntax do?

There have been many references to confusion and questions of usefulness regarding this syntax.  But I think this is a syntax that is underutilized due to lack of understanding.

## Explanation by way of JSON Schema

In JSON Schema, there have been many discussions around the idea of comparing instance data against other instance data.  You can perform a search for "$data" on the schema spec site to read up on this, but I'll summarize.

Suppose we have these instances:

```json
// valid
{ "value": 5, "lower-bound": 3 }

// invalid
{ "value": 5, "lower-bound": 10 }
```

With Schema in its current state, we can't validate these.  That's what the proposed `$data` keyword was supposed to solve.  It would provide a mechanism to reference other data for use in other keywords.  In this example, it could use the value in `lower-bound` for the `minimum` keyword.

It was never adopted, however, because no one could figure out how to make it work within the existing mechanisms of JSON Schema. (Beside the point.)

(It's also similar to how `discriminator` works in the OpenAPI 3.1 specification.)

## Back to JSON Path

I think this operation is quite useful in that it allows the path author to use data within the Argument.

Let's look at an example path: `$.x[($.id)]`.

This, to me, says select the value at `$.id`.  So long as this value contains a valid index, this should evaluate.  We have different cases.

(Sorry, I can't make a pretty table as I'm on mobile.)

- `{ "id": 2, "x": [ 5, 10, 15 ] }` -> `[15]`
- `{ ”id": "foo", "x": { "foo": 10, "bar": 15 } }` -> `[10]`

Maybe even this could be supported, though we run into "interpretation" issues of converting a string into something more meaningful to us.

- `{ "id": "0:1", "x": [ 5, 10, 15 ] }` -> `[5, 10]`

I don't think that it make sense for `$.id` to contain a union, only a single index.

I think that if people understood this functionality better, it might get more usage.  Path authors could devise some fairly complex queries that couldn't otherwise be built.  I think this syntax _with this function_ is worth including.

[...]

PR #153, opened 2022-02-01
Remove examples with invalid syntax

glyn:
Index expressions of the form [(...)] are not supported.

@.length, to obtain the length of an array, is not supported.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -376,7 +376,7 @@ constant.
 | `$.store.*`                               | all things in store, which are some books and a red bicycle  |
 | `$.store..price`                          | the prices of everything in the store                        |
 | `$..book[2]`                              | the third book                                               |
-| `$..book[(@.length-1)]`<br>`$..book[-1]`  | the last book in order                                       |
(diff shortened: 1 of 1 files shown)


gregsdennis:
> `@.length`, to obtain the length of an array, is not supported.

We'll need to come back to this eventually.  Finding the length is useful in other scenarios like `$[?(@.length < 10)]`.

I'm not saying we should keep this syntax, though.  I'd prefer a more function-like syntax: `$[?(length(@) < 10)]`, but we haven't defined functions either.

gregsdennis (approved):

glyn:
> > `@.length`, to obtain the length of an array, is not supported.
> 
> We'll need to come back to this eventually. Finding the length is useful in other scenarios like `$[?(@.length < 10)]`.
> 
> I'm not saying we should keep this syntax, though. I'd prefer a more function-like syntax: `$[?(length(@) < 10)]`, but we haven't defined functions either.

@gregsdennis We have https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/93 to cover non-filter index expressions, but I am not aware of an issue which captures the "length" feature above. Would you care to add it?

PR #158, opened 2022-02-21
s/which/that/ in restrictive places

cabo:


glyn (approved):
LGTM

glyn:
Editorial change only - merged.

PR #163, opened 2022-03-07
Add proposal for handling regexps.

### LINKED ISSUE #70: Regular expressions in filters
Issue 17 [proposed](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/17#issuecomment-797044615) regular expressions in filters, which is supported by [several implementations](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_regular_expression.html) of JSONPath. This raises two questions:

1. What is the precise syntax of regular expressions?
2. How should [ReDOS attacks](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/25#issuecomment-700801725) be mitigated?

One approach which addresses both these questions would be to adopt [RE2 syntax](https://github.com/google/re2/wiki/Syntax) as discussed in [ReDOS attacks](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-jsonpath/issues/25#issuecomment-700801725), although not all languages may yet have support for RE2.

cabo:
Close #70

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1066,7 +1066,7 @@ Notes:
   side of `=~` are as defined in {{-iregexp}}.
   Regular expression tests can be applied to JSON string values
   ({{Section 7 of -json}}) only
-  (on the left-hand side of `=~`).
+  (on the left-hand side of `=~`); they yield false otherwise.
 * Alphabetic characters in ABNF are case-insensitive, so "e" can be either "e" or "E".
 * false, null, true are lower-case only (case-sensitive).
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -401,6 +401,9 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1]).
 
+To be valid, strings on the right hand side of the `=~` regex matching
+operator need to conform to {{-iregexp}}.
+
 The well-formedness and the validity of JSONPath queries are independent of
 the JSON value the query is applied to; no further errors can be
 raised during application of the query to a value.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -113,6 +113,7 @@ normative:
   RFC8259: json
   RFC7493: i-json
   RFC6838: media-types-reg
+  I-D.draft-bormann-jsonpath-iregexp: iregexp
 
 venue:
   group: JSON Path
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1047,21 +1048,25 @@ null         = %x6e.75.6c.6c                          ; null
...
 * A member or element value by itself in a Boolean context is
   interpreted as `false` only if it does not exist.
   Otherwise it is interpreted as `true`.
   To be more specific about the actual value, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with structured values.
-* Regular expression tests can be applied to `string` values only.
+* The regular expressions in the string-literals on the right-hand
+  side of `=~` are as defined in {{-iregexp}}.
+  Regular expression tests can be applied to JSON string values
+  ({{Section 7 of -json}}) only
+  (left-hand side of `=~`).
```
glyn, line 1069:
```suggestion
  (on the left-hand side of `=~`).
```

draft-ietf-jsonpath-base.md
```diff
@@ -1047,21 +1048,25 @@ null         = %x6e.75.6c.6c                          ; null
...
-* Types are not implicitly converted in comparisons.
+* Data types are not implicitly converted in comparisons.
   So `"13 == '13'"` selects no node.
 * A member or element value by itself in a Boolean context is
   interpreted as `false` only if it does not exist.
   Otherwise it is interpreted as `true`.
   To be more specific about the actual value, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with structured values.
-* Regular expression tests can be applied to `string` values only.
+* The regular expressions in the string-literals on the right-hand
+  side of `=~` are as defined in {{-iregexp}}.
```
glyn, line 1066: Given that the iregexp syntax is not specified in this document, should we clarify what happens if the provided regex doesn't conform to iregexp syntax? I would expect a syntax error rather than some kind of non-matching behaviour.

draft-ietf-jsonpath-base.md
glyn, line 1069: What happens when there is a path on the LHS which does not identify a string? I presume the regex result is `false`, but the spec needs to be explicit.

glyn (changes requested):
Thanks for starting this. I suggested a few improvements.

draft-ietf-jsonpath-base.md
cabo: Indeed.  See 
d674175

draft-ietf-jsonpath-base.md
cabo: Good point.  See 
f07f023

cabo:
Anything else we expect to merge for this I-D deadline?
If not, I'd like to submit.

goessner:
great ,,, thanks.

Issue #155, opened 2022-02-15
Normalized Paths and normalization


benkehoe:
The [current definition for Normalized Paths](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/ca83d54f420576b03d0700e201c5447a2b96aa0b/draft-ietf-jsonpath-base.md#normalized-paths) is a separate grammar that happens to be a subset of the general path grammar. Would it make sense to make it more directly a subset (e.g., as a set of constraints on selectors), which could then include a "normalization" operation that takes a standard path and returns a Normalized Path or an error if it cannot be normalized?

A use case: if I am a developer intending to provide an interface that accepts a JSON Path that must refer to a single element, is it reasonable for me to expect that a user of that interface knows what Normalized Paths are such that I can require a Normalized Path (for example, they know they must use single quotes, not double quotes), or would it be better to allow them to specify a standard path and then attempt to normalize it as part of input validation?

gregsdennis:
Speaking as another developer, I think you'll want to support translation to normalized paths.  Most people will want to write the dot syntax if they can. So long as they specify a single node, your code should work.

I don't think it's a matter of what we suggest rather than one of usability for your consumers.

glyn:
> The [current definition for Normalized Paths](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/ca83d54f420576b03d0700e201c5447a2b96aa0b/draft-ietf-jsonpath-base.md#normalized-paths) is a separate grammar that happens to be a subset of the general path grammar.

`normalized-path` is indeed a subset of `json-path`. The ABNF was written to make sure this is true and the spec says so: "A Normalized Path is a JSONPath with restricted syntax that [...]".

> Would it make sense to make it more directly a subset (e.g., as a set of constraints on selectors), which could then include a "normalization" operation that takes a standard path and returns a Normalized Path or an error if it cannot be normalized?

I don't think ABNF has any way of defining a rule which is expressed as a constrained form of another rule. Implementations are free to do this if they can and are also free to provide a "normalize()" operation.

cabo:
> 
> I don't think ABNF has any way of defining a rule which is expressed as a constrained form of another rule. Implementations are free to do this if they can and are also free to provide a "normalize()" operation.

Right, ABNF is a grammar, not a transformation language.
I think we are best off with making sure there indeed is a transformation that always works.
But we can describe that in English.

[...]

PR #164, opened 2022-03-08
Singular paths

glyn:
Introduce the term Singular Path.

Restrict paths in filter expressions to be Singular Paths so that,
for instance, existence tests and the left and right side of comparisons
do not involve more than one node.

Ref: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/155

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -227,7 +227,7 @@ Normalized Path:
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
 Singular Path:
-: A JSONPath expression that identifies at most one node.
+: A JSONPath expression built from selectors which each select at most one node.
 
 For the purposes of this specification, a value as defined by
 {{-json}} is also viewed as a tree of nodes.
@@ -1129,8 +1129,8 @@ that will yield distinct results when the Normalized Paths are applied to it.
 
 Certain characters are escaped, in one and only one way; all other characters are unescaped.
 
-Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-1]`, for example, is a Singular
-Path but not a Normalized Path.
+Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-3]`, for example, is a Singular
+Path, but not a Normalized Path.
 
 ~~~~ abnf
 normalized-path           = root-selector *(normal-index-selector)
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -226,6 +226,9 @@ Normalized Path:
   providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
+Singular Path:
+: A JSONPath expression that identifies at most one node.
+
 For the purposes of this specification, a value as defined by
 {{-json}} is also viewed as a tree of nodes.
 Each node, in turn, holds a value.
@@ -1025,14 +1028,15 @@ logical-or-expr  = logical-and-expr *(S "||" S logical-and-expr)
 logical-and-expr = basic-expr *(S "&&" S basic-expr)  ; conjunction
                                                       ; binds more tightly than disjunction
 
-basic-expr   = exist-expr /
-               paren-expr /
-               relation-expr
-exist-expr   = [neg-op S] path                          ; path existence or non-existence
-path         = rel-path / json-path
-rel-path     = "@" *(S (dot-selector / index-selector))
-paren-expr   = [neg-op S] "(" S boolean-expr S ")"    ; parenthesized expression
-neg-op       = "!"                                    ; not operator
+basic-expr        = exist-expr /
+                    paren-expr /
+                    relation-expr
+exist-expr        = [neg-op S] singular-path          ; path existence or non-existence
+singular-path     = rel-singular-path / abs-singular-path
+rel-singular-path = "@" *(S (dot-selector / index-selector))
+abs-singular-path = root-selector *(S (dot-selector / index-selector))
+paren-expr        = [neg-op S] "(" S boolean-expr S ")" ; parenthesized expression
+neg-op            = "!"                               ; not operator
 
 relation-expr = comp-expr /                           ; comparison test
                 regex-expr                            ; regular expression test
@@ -1040,7 +1044,7 @@ relation-expr = comp-expr /                           ; comparison test
 comp-expr    = comparable S comp-op S comparable
 comparable   = number / string-literal /              ; primitive ...
                true / false / null /                  ; values only
-               path                                   ; path value
+               singular-path                          ; Singular Path value
 comp-op      = "==" / "!=" /                          ; comparison ...
                "<"  / ">"  /                          ; operators
                "<=" / ">="
@@ -1051,7 +1055,7 @@ true         = %x74.72.75.65                          ; true
 false        = %x66.61.6c.73.65                       ; false
 null         = %x6e.75.6c.6c                          ; null
 
-regex-expr   = (path / string-literal) S regex-op S regex
+regex-expr   = (singular-path / string-literal) S regex-op S regex
 regex-op     = "=~"                                   ; regular expression match
 regex        = string-literal                         ; I-Regexp
 ~~~~
@@ -1067,6 +1071,7 @@ Notes:
   interpreted as `false` only if it does not exist.
   Otherwise it is interpreted as `true`.
   To be more specific about the actual value, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with structured values.
+* Paths in filter expressions are Singular Paths, each of which selects at most one node.
 * The regular expressions in the string-literals on the right-hand
   side of `=~` are as defined in {{-iregexp}}.
   Regular expression tests can be applied to JSON string values
@@ -1123,6 +1128,9 @@ given Normalized Path. Putting this another way, for any two distinct Normalized
 that will yield distinct results when the Normalized Paths are applied to it.
 
 Certain characters are escaped, in one and only one way; all other characters are unescaped.
+
+Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-1]`, for example, is a Singular
+Path but not a Normalized Path.
 
 ~~~~ abnf
 normalized-path           = root-selector *(normal-index-selector)


draft-ietf-jsonpath-base.md
```diff
@@ -1023,22 +1026,23 @@ logical-or-expr  = logical-and-expr *(S "||" S logical-and-expr)
...
-               relation-expr
-exist-expr   = [neg-op S] path                          ; path existence or non-existence
-path         = rel-path / json-path
-rel-path     = "@" *(S (dot-selector / index-selector))
-paren-expr   = [neg-op S] "(" S boolean-expr S ")"    ; parenthesized expression
-neg-op       = "!"                                    ; not operator
+basic-expr        = exist-expr /
+                    paren-expr /
+                    relation-expr
+exist-expr        = [neg-op S] singular-path          ; path existence or non-existence
```
cabo, line 1032: Why does an exist expression need to be singular?

draft-ietf-jsonpath-base.md
```diff
@@ -226,6 +226,9 @@ Normalized Path:
   providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
+Singular Path:
```
cabo, line 229: The question is whether we define this via the outcome or by saying how it is constructed.  The ABNF of course does the latter.

draft-ietf-jsonpath-base.md
```diff
@@ -1049,7 +1053,7 @@ true         = %x74.72.75.65                          ; true
 false        = %x66.61.6c.73.65                       ; false
 null         = %x6e.75.6c.6c                          ; null
 
-regex-expr   = (path / string-literal) S regex-op S regex
+regex-expr   = (singular-path / string-literal) S regex-op S regex
```
cabo, line 1056: Hmm, so I can't use a regexp to search in multiple elements.
I think that is something we can allow (~ matches at least one).

draft-ietf-jsonpath-base.md
glyn: Normalized Path is defined via the outcome, so I think we should do the same here for consistency.

glyn:
Rebased after merging https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/165.

draft-ietf-jsonpath-base.md
goessner: Relative pathes always should be singular.
```
During iteration process each array element or object member is
visited and its value -- accessible via symbol @ -- or one of its
descendants -- uniquely defined by a relative path -- is tested
against a boolean expression boolean-expr.
```

```
 rel-path     = "@" *(S (dot-selector / index-selector))
```
They seem to be naturally singular.

draft-ietf-jsonpath-base.md
goessner: I vote for the `singular-path` case here for clarity.

Remember ... filters are iterating over members or elements one by one and apply tests with each in turn ... so allowing multiple elements arising somehow would confuse ,,, at least me.

draft-ietf-jsonpath-base.md
glyn: But what about absolute paths?

glyn: Does the following example help to put you off non-singular absolute paths in filters?
```
$[?(@.a==$.*)]
```

goessner: hmm ... this should select nothing, since a singular value can't be equal to a nodelist ... implying RHS resulting not in a single value.

So only singular absolute pathes survive in practice ... so why restrict them artificially in ABNF ?

goessner (approved):

gregsdennis (changes requested):
I'm not sure we need to restrict `@`-paths to single values.

Suppose we want to select all children that contain a value.  Then we need something like `?('foo' in @.*)`, which can't be done if it's only allowed to produce single values.

For that expression, we definitely need to say that the LHS of `in` must be singular, but that doesn't imply that ALL `@`-paths need to be singular.

I think the definition is worth keeping for this purpose, but I don't think it needs to be tied to `@`-paths.

glyn:
> I'm not sure we need to restrict `@`-paths to single values.
> 
> Suppose we want to select all children that contain a value. Then we need something like `?('foo' in @.*)`, which can't be done if it's only allowed to produce single values.
> 
> For that expression, we definitely need to say that the LHS of `in` must be singular, but that doesn't imply that ALL `@`-paths need to be singular.
> 
> I think the definition is worth keeping for this purpose, but I don't think it needs to be tied to `@`-paths.

We scrapped `in`, so please could you re-frame your concern in terms of features in the spec.

gregsdennis:
Then why are we defining this?  If we don't have any operators that allow an operand to be a nodelist, then we get this for free.

gregsdennis:
I think I'm more responding to @goessner's

> Relative pathes always should be singular

Any path should be able to return a nodelist with multiple items as long as the operator supports it.  We don't _have_ such operators at this point, so I think this is moot.

danielaparker:
> Suppose we want to select all children that contain a value. Then we need something like `?('foo' in @.*)`, which can't be done if it's only allowed to produce single values.

Do you mean something like applying
```
$[?(1 in @.*)]
```
to
```json
[{"a":1,"b":2},{"c":3,"d":4}]       
```
? In Jayway, probably the most important of the implementations that invented their own query language, that results in
```json
[
   {
      "a" : 1,
      "b" : 2
   }
]
```
Note that in Jayway `@.*` in a filter expression _does_ evaluate to a single value, in this case an array of child values.

draft-ietf-jsonpath-base.md
danielaparker: > hmm ... this should select nothing, since a singular value can't be equal to a nodelist ... implying RHS resulting not in a single value.
> 
Note though that JSONPath implementations divide into (1) those that use JavaScript, Python, PHP for expressions, and (2) those that invented their own expression language. Considering only the second group, you'll find that the important ones including Jayway allow expressions starting with `@` to be full JSONPath expressions in their own right, and that the result of evaluating these JSONPath expressions within filter expressions are _not_ result lists, rather, they are single values that are resolved according to well defined rules. What you're describing is different than what is found in prior experience, in prior experience, JSONPath expressions appearing in filter expressions do evaluate to single values.

danielaparker:
> Then why are we defining this? If we don't have any operators that allow an operand to be a nodelist, then we get this for free.

Agreed. There is no reason to interpret the right hand side of
```
rel-path     = "@" *(S (dot-selector / index-selector))
```
as a path expression.

draft-ietf-jsonpath-base.md
glyn: > > hmm ... this should select nothing, since a singular value can't be equal to a nodelist ... implying RHS resulting not in a single value.
> 
> Note though that JSONPath implementations divide into (1) those that use JavaScript, Python, PHP for expressions, and (2) those that invented their own expression language. Considering only the second group, you'll find that the important ones including Jayway allow expressions starting with `@` to be full JSONPath expressions in their own right, and that the result of evaluating these JSONPath expressions within filter expressions are _not_ result lists, rather, they are single values that are resolved according to well defined rules. What you're describing is different than what is found in prior experience, in prior experience, JSONPath expressions appearing in filter expressions do evaluate to single values.

If we automatically converted node sets to arrays in filter expressions:

1. This would break existing implementations, including Jayway, in examples such as [$[?(@.key>42)]](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_greater_than.html) and [$[?(!@.key)]](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_negation_and_without_value.html) which expect singular paths to produce at most one _value_. (`@.key` would produce a node set with at most one node and this would convert to an empty array or an array with one value.)
2. Objects could produce non-deterministic arrays, which would be tricky to use in comparisons.

danielaparker: > If we automatically converted node sets to arrays in filter expressions:
> 
> 1. This would break existing implementations, including Jayway, in examples such as [$[?(@.key>42)]](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_greater_than.html) and [$[?(!@.key)]](https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_negation_and_without_value.html) which expect singular paths to produce at most one _value_. (`@.key` would produce a node set with at most one node and this would convert to an empty array or an array with one value.)

I don't know of _any_ implementation that automatically converts "results lists" to arrays inside filter expressions. 

Jayway doesn't produce result lists at all when evaluating a JSONPath expression inside a filter. It evaluates them as a single value according to the following rules: The slice, recursive descent and union selectors, which can evaluate to zero, one, or many items, are wrapped in a Json array. The object wildcard selector is wrapped in an array if the object has two or more members, otherwise a single value is returned. The array wildcard operator always returns an array. The others - root, current node, parent node, identifier, and index selectors - evaluate to a single value if matched, otherwise a null value.

As far as I can tell, the important JSONPath implementations that invent their own filter expressions generally allow expressions beginning with '@' to be first class JSONPath expressions. And they evaluate them as single values within filter expressions, they need to produce single values for comparisons with JSON literals to make sense. So they have rules for that, but they don't all follow the same rules as Jayway. Nonetheless, Jayway is probably the most important and influential of these.

The alternative you find in prior experience are the implementations that use JavaScript or Python or PHP for filter expressions. In these, expressions beginning with `@` aren't JSONPath expressions at all. Here `@` is simply a JavaScript or Python or PHP variable, with supported indexing operations. I believe this experience is what Stefan has in mind.

glyn: > Jayway doesn't produce result lists at all when evaluating a JSONPath expression inside a filter. It evaluates them as a single value according to the following rules: The slice, recursive descent and union selectors, which can evaluate to zero, one, or many items, are wrapped in a Json array. The object wildcard selector is wrapped in an array if the object has two or more members, otherwise a single value is returned. The array wildcard operator always returns an array. The others - current node, identifier, and index selectors - evaluate to a single value if matched, otherwise a null value.

I find this paragraph confusing.

"The slice, recursive descent and union selectors, which can evaluate to zero, one, or many items, are wrapped in a Json array. " -- I don't know what it means to wrap a selector in a JSON array. I presume it should say the node sets produced by these selectors  are wrapped in JSON arrays, which sounds very much like automatically converting node sets into arrays.

"The object wildcard selector is wrapped in an array if the object has two or more members, otherwise a single value is returned." -- I presume you mean the node set produced by such a selector is wrapped (in a non-deterministic order?) in a JSON array if it has two or more members, that if the node set has a single value, this value is returned, and if the node set is empty, a null value is returned. (The conditional behaviour in this case is intriguing: if there is a single value in the node set which is an array, that array will be returned and there is nothing to distinguish the return value from an array produced from a node set with two or more members.)

glyn: At today's JSONPath WG interim meeting, the consensus was to allow only singular paths in comparisons and regex tests. This leave open whether we should allow non-singular paths for existence tests. My position is that it is more consistent and more memorable for users to require singular paths across the board in filter expressions than to have a different rule for existence tests.

@cabo may wish to jump in here.

draft-ietf-jsonpath-base.md
glyn: I think I've understood @cabo's point better now. See 46ce76c95cd631faba3b33602f9d0e7175a73d13 for a possible solution.

draft-ietf-jsonpath-base.md
glyn: Keeping `singular-path` on the LHS of regex expressions was the consensus as today's interim WG meeting.

draft-ietf-jsonpath-base.md
danielaparker: > I find this paragraph confusing.
> 
> "The slice, recursive descent and union selectors, which can evaluate to zero, one, or many items, are wrapped in a Json array. " -- I don't know what it means to wrap a selector in a JSON array. I presume it should say the node sets produced by these selectors are wrapped in JSON arrays, which sounds very much like automatically converting node sets into arrays.
> 
> "The object wildcard selector is wrapped in an array if the object has two or more members, otherwise a single value is returned." -- I presume you mean the node set produced by such a selector is wrapped (in a non-deterministic order?) in a JSON array if it has two or more members, that if the node set has a single value, this value is returned, and if the node set is empty, a null value is returned. (The conditional behaviour in this case is intriguing: if there is a single value in the node set which is an array, that array will be returned and there is nothing to distinguish the return value from an array produced from a node set with two or more members.)

Jayway's query function provides options to evaluate a JSONPath expression either as a result list or as a single value, the user can choose. Within filter expressions, though, an embedded JSONPath expression is always evaluated as a single value. For evaluating a JSONPath expression as a single value, the basic idea is that a selector that returns at most one value returns either that value or a JSON null (if none), and a selector that potentially can return zero or many values returns them as an array (rather than appending them to a result list). This isn't particularly well documented, but if you find my explanation unclear, you can refer to [JSONPath Reference](https://support.smartbear.com/readyapi/docs/testing/jsonpath-reference.html), the section on "Usage notes and considerations." Or if you're familiar with JMESPath, it's similar to how JMESPath evaluates expressions.

My main point though is as follows. 

Currently in the grammar, you have, for filter selectors
```
path         = rel-path / json-path
rel-path     = "@" *(S (dot-selector / index-selector))
```

If you allow a `json-path` in an expression, I think you, like Jayway and other implementations, have to answer the question: how do I evaluate the `json-path` as a single value? And if you answer that question (your answer may be different than Jayway's), why limit `rel-path`? Why not let expressions beginning with '@' be full JSONPath expressions in their own right, like they are in Jayway (and other implementations that invented their own expression language)? Why not allow JSONPath expressions to begin with either '$' or '@', as Jayway does, and Bettio  proposed? This is eminently doable, there is proven prior experience.

On the other hand, if you prefer the simplicity of limiting expressions beginning with `@` to 
```
"@" *(S (dot-selector / index-selector))
```
which is consistent with the implementations that dovetail on JavaScript, Python and PHP, why not interpret them in the same way as JavaScript, Python, and PHP do? As evaluating a variable '@' against identifier and index operators, which always return a single value? Where they are most emphatically _not_ JSONPath expressions, "singular" or otherwise?  Why not just drop `json-path` from the allowed elements in filters, so you don't have to tackle the problem of how to evaluate a JSONPath expression as a single value at all?

glyn: > Why not just drop json-path from the allowed elements in filters, so you don't have to tackle the problem of how to evaluate a JSONPath expression as a single value at all?

I believe that's consistent with the direction we are taking.

glyn:
@gregsdennis wrote:
> I think I'm more responding to @goessner's
> 
> > Relative pathes always should be singular
> 
> Any path should be able to return a nodelist with multiple items as long as the operator supports it. We don't _have_ such operators at this point, so I think this is moot.

If we allow non-singular paths (which can result in a nodelist with more than one member) in comparisons, for example, then we have to define the semantics. It's not clear whether comparison should hold for all members of the node list or at least one member. For example, consider `?@.* > 3` where @ corresponds to the array `[2,4,6]`.

The approach taken here was the consensus at today's WG meeting (except perhaps for existence tests which @cabo would like to think through).

Does this help explain the point of this PR?

gregsdennis:
> If we allow non-singular paths (which can result in a nodelist with more than one member) in comparisons...

I agree with this, but the proper solution isn't to limit the path to single values.  We should define the _comparison operator_ itself to only be valid for single values.  Then, when an operator gets multiple values on either side, it can just return false.

By doing this, it'll be easier later when we want to define `in` or how complex values in comparisons work.

glyn:
> > If we allow non-singular paths (which can result in a nodelist with more than one member) in comparisons...
> 
> I agree with this, but the proper solution isn't to limit the path to single values. We should define the _comparison operator_ itself to only be valid for single values. Then, when an operator gets multiple values on either side, it can just return false.
> 
> By doing this, it'll be easier later when we want to define `in` or how complex values in comparisons work.

I disagree for two reasons:

1. Making equality silently result in `false` in the non-singular path case is more surprising/obscure for the users than a syntax error on a non-singular path.
2. It's easier to extend error cases into non-error cases. Once we define comparisons of non-singular paths to be `false`, we'll have to maintain that behaviour.

gregsdennis (approved):
🤔 I still don't agree with this approach, but I don't really have an argument against it.

glyn:
@cabo If you would like to revisit the use of singular paths in existence tests, please could you raise an issue.

Issue #150, opened 2022-01-24
Should list selectors allow filter selectors?


cabo:
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/147#discussion_r790396618

https://cburgmer.github.io/json-path-comparison/results/union_with_filter.html

cabo:
(To me it is inexplicable they aren't already.  But changing that was not the remit of #147.)

goessner:
I am fine with the status quo of excluding filters within list selectors.

* One is *either* building a list of explicite indices/indexes *or* a boolean expression for filtering.
* Two or more filter expressions in a list are equivalent to have a single one or'ed them together.
* Most cases in practises could be solved by a sequence
    * list selector following filter selector
    * filter selector following list selector

Any striking use cases  I am not aware of?

gregsdennis:
As mentioned in my comment, I see no reason why a filter can't be included in a list.  Some people may prefer to have multiple simple expressions for readability.

My implementation already supports this (I'll have to disable this functionality or hide it behind an option if we don't include it).

cabo:
I don't have a strong opinion on this question.

But not only does it seem natural to include filter selectors in list selectors: the comparison project also shows that most implementations that do provide this, also agree on what the result should be.  So this would not be a risky addition.

goessner:
I do have a strong opinion here, as ...
* index expressions (member names, index and slices) are strongly explicit and single level only.
* list-selector is merely a list of those index expressions.

Example above
```
$[?(@.key<3),?(@.key>6)]
```
is easily - and more consistently - reformulated as a pure filter-selector
```
$[?(@.key<3 || @.key>6)]
```
So I would like to see real use cases to be convinced here.

If filter selectors are natural to be included in list selectors, then so are

* index-wild-selector
* descendant-selector
* and even list-selector

goessner:
> (The sequence of data items is different between the two.)

 ???

> That already works :-) If you have the list selector [1, 2, 3], you can include that in [4, 5, 6] by unpacking the brackets, so you have 1, 2, 3, and appending that to the comma-separated list, so you have [4, 5, 6, 1, 2, 3] 

`:-)`

> (Note that all the embedded selectors have their brackets removed before including them in a list-selector.)

... even with filter selectors ?

cabo:
> On 2022-01-24, at 17:34, Stefan Goessner ***@***.***> wrote:
> 
[...]

PR #165, opened 2022-03-08
Allow filters in list selectors

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/150


cabo (approved):

goessner:
I can remember, that I was finally convinced with allowing all other selectors in list-selectors ... even list-selectors IIRC.

Issue #136, opened 2021-11-10
In-line examples


cabo:
Apart from an appendix with examples, let's have in-line examples

glyn:
To do list:

- [ ] Root selector
- [ ] Dot selector
- [ ] Dot wild card selector
- [ ] Index selector
- [ ] Index wild card selector
- [ ] Array slice selector
- [ ] Descendant selector
- [ ] List selector
- [ ] Filter selector
- [ ] Normalized paths

goessner:
where should examples go in the meantime ?

cabo:
I'd say:

ideas: right here

full proposals: PRs

glyn:
Yes, drop examples in here and I'll try to include them in PRs. The only problem will be if an example turns out to generate a lot of discussion, as issues aren't great at threading. But let's cross that bridge when we get to it...

Or feel free to submit PRs. :-)

goessner:
Format could be something like that ...

### `descendent-selector` 

| JSON | Query | Result | Comment| 
| --- | --- | --- | --- |
| `{ "o": {"key":7}}` | `$..o` | `$['o']` | Child node  |
|  | `$..['key']` | `$['o']`<br>`$['o']['key']` | Descendent node  | 
|  | `$..*` | `$['o']`<br>`$['o']['key']` |All descendent nodes  | 
|  | `$..[*]` | `$['o']`<br>`$['o']['key']` |All descendent nodes  | 

What's about invalid queries? Should they be mixed in or in a separate table / section ?

glyn:
I quite like that format. Thanks! I anticipate each selector having three sections: "Syntax", "Semantics", and "Examples". It should be pretty obvious that the examples are non-normative without saying so every time.

It seems more manageable and readable to have the JSON document available alongside the examples rather than having to "jump around" the spec, even if this does lead to some duplication.

I like showing the result as Normalized Paths rather than values, since the former are unambiguous in the presence of duplicate values. I wonder if readers will be as comfortable? I guess they'll soon get the hang of it.

goessner:
Thanks ... I will be back here earliest on Sunday ...

glyn:
Inline examples are now complete.

PR #168, opened 2022-03-09
Inline examples (batch 1)

glyn:
Let's use this PR to hash out the formatting preferences. We can come back and tweak later, but please suggest any changes now before I replicate this format in other sections.

See the [rendered HTML](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/inline-examples-root/draft-ietf-jsonpath-base.html#name-root-selector) and the [rendered plain text](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/inline-examples-root/draft-ietf-jsonpath-base.txt).

Ref: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/136

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1015,6 +1015,23 @@ When `step = 0`, no elements are selected and the result array is empty.
 To be valid, the slice expression parameters MUST be in the I-JSON
 range of exact values, see {{synsem-overview}}.
 
+#### Examples
+{: unnumbered}
+
+JSON document:
+
+    ["a", "b", "c", "d", "e", "f", "g"]
+
+Queries:
+
+| Query | Result | Result Paths | Comment |
+| :---: | ------ | :----------: | ------- |
+| `$[1:3]` | `"b"` <br> `"c"` | `$[1]` <br> `$[2]` | Slice with default step |
+| `$[1:5:2]` | `"b"` <br> `"d"` | `$[1]` <br> `$[3]` | Slice with step 2 |
+| `$[5:1:-2]` | `"f"` <br> `"d"` | `$[5]` <br> `$[3]` | Slice with negative step |
+| `$[::-1]` | `"g"` <br> `"f"` <br> `"e"` <br> `"d"` <br> `"c"` <br> `"b"` <br> `"a"` | `$[6]` <br> `$[5]` <br> `$[4]` <br> `$[3]` <br> `$[2]` <br> `$[1]` <br> `$[0]` | Slice in reverse order |
+{: title="Array slice selector examples"}
+
 ### Descendant Selector
 
 #### Syntax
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -573,6 +573,13 @@ root-selector  = "$"
 The Argument — the root JSON value — becomes the root node, which is
 addressed by the root selector `$`.
 
+#### Examples
+{: unnumbered}
+
+| JSON | Query | Result | Comment|
```
cabo, line 579: Shouldn't we give the value in addition to the path for the result?
This is a bit tautological at the moment.

glyn: Yeah, that's probably more friendly to the users. I'll do it.

goessner: Drawbacks with Markdown tables are:
* Column sizes are not explicitly adjustable; but implicitly by Header length.
* Multiline codeblocks are not supported inside. HTML `<br>`s are necessary here.

Especially rendered plain text leaves only few horizontal space.

So I suggest:

* Widening / Shortening Heading-text where appropriate.
* Minimize number of columns.
* Use short comments only.

| JSON document  | Query | Value | Result | Comment| 
| --- | --- | --- | --- | --- |
| `{ "o": {"key":7}}` | `$..o` | `{"key":7}` | `$['o']` | Child node  |
|  | `$..['key']` | 7 | `$['o']['key']` | Descendent node  |
|  | `$..*`  | `{"key":7}`<br> 7  | `$['o']`<br>`$['o']['key']` |All descendents  | 
|  | `$..[*]`  | `{"key":7}`<br> 7  | `$['o']`<br>`$['o']['key']` |All descendents  | 

Maybe with that space restrictions we might consider to show either result values or normalized pathes.
Another possibility would be to write the JSON before the table; then restricting all queries to that document.

cabo: Not commenting on the content of the table:  The RPC (RFC production center) often re-writes tables that have become unwieldy into definition lists.

(Note that most of the limitations you perceive here are those of RFCXML; markdown is just a way to input that RFCXML.)

The table here could be made more readable by leaving out the common value "JSON document" and just mentioning that before the start of the table.

goessner: Well ... then we have ...

### descendant-selector

```json
{ 
  "o": {
    "key":7
  }
}
```
| Query | Value | Result | Comment| 
| --- | --- | --- | --- |
|`$..o` | `{"key":7}` | `$['o']` | Child node  |
| `$..['key']` | 7 | `$['o']['key']` | Descendent node  |
| `$..*`  | `{"key":7}`<br> 7  | `$['o']`<br>`$['o']['key']` |All descendents  | 
| `$..[*]`  | `{"key":7}`<br> 7  | `$['o']`<br>`$['o']['key']` |All descendents  |

glyn: Please see 5df7f520de3dbce58f289636779233017cc340d5. The text rendering is still far from perfect. I don't know how to adjust column widths reliably.

glyn:
Since this PR has quite a sprawling diff, I'm going to merge immediately. We can come back and tweak the examples as needed, of course.

Issue #166, opened 2022-03-09
Reorganise and refactor list and filter selector sections


glyn:
Filter selectors should come before list selectors to avoid a forward reference.

The following comment should be deleted:
```
<!-- To do: Move list selector after filter selector to avoid a forward reference. -->
```

Also, as noted by @cabo at today's interim WG meeting, we should factor out a separate production to avoid repeating `"?" S boolean-expr` here:
```
list-entry     =  ( quoted-member-name /
                    element-index      /
                    slice-index /
                    ("?" S boolean-expr) ; filter expression
                  )
```
and here:
```
filter-selector    = "[" S "?" S boolean-expr S "]"
```
Possible names for this production: `filter-selector` or `filter`.

goessner:
I think we proposed something like ...

```
list-entry     =  ( quoted-member-name /
                    element-index      /
                    slice-index /
                    filter-expr
                  )
filter-expr = "?" S boolean-expr
```
and here:
```
filter-selector    = "[" S filter-expr S "]"
```
thanks.

PR #170, opened 2022-03-14
Move list select after filter selector

glyn:
To do:

- [x]  Consider reorganising selectors in section 3.3 to match the order of the document
- [x] Consider reorganising table 1 to match the order of the document

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/166

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -475,10 +475,10 @@ json-path = root-selector *(S (dot-selector        /
                                dot-wild-selector   /
                                index-selector      /
                                index-wild-selector /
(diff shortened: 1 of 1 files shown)


glyn:
> To do:
> 
>     * [x]   Consider reorganising selectors in section 3.3 to match the order of the document
> 
>     * [x]  Consider reorganising table 1 to match the order of the document
> 
> 
> Fixes #166

Reorganised selectors in section 3.3 to match the order of the document. Did not reorganise table 1 as it flows better as is.

goessner (approved):

PR #173, opened 2022-03-18
(housekeeping): YAML header cleanup

cabo:


glyn (approved):
I don't understand any of this PR, but since it deletes stuff I don't understand, that can't be bad, can it?

cabo:
:-)  yes, the idea is that v: 3 provides some defaults for the more arcane settings.
Thanks for the merge.

PR #174, opened 2022-03-18
Add content to Security Considerations section

### LINKED ISSUE #25: Security items
Let's use this issue to capture possible items for the Security Considerations section of the spec.

As discussed in https://github.com/jsonpath-standard/internet-draft/issues/14, CPU exploits are possible using selectors such as:
```
$[?($)][?($)][?($)][?($)]...
$[?(@==$[?(@==$[?(@==$[?(@==$[?(@==$[?(@==$[?(@==$[?(@==$[?(@==$[?(@=="b")])])])])])])])])])]
$[?(@.a || @.b || @.c ...)]
```
One mitigation would be to run the implementation in an environment, such as a Linux cgroup, which can restrict the CPU consumed.

cabo:
Close #25

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -77,7 +77,7 @@ contributor:
   email: esurov.tsp@gmail.com
 
 informative:
-  RFC3552: seccons
+#  RFC3552: seccons
 #  RFC8126: ianacons
   RFC6901: pointer
   RFC6901: pointer
@@ -106,9 +106,10 @@ informative:
     author:
     - org: Ecma International
     date: 1999-12
+  RFC8949: cbor
 
 normative:
-  RFC3629:
+  RFC3629: utf8
   RFC5234: abnf
   RFC8259: json
   RFC7493: i-json
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
 
 # Security Considerations {#Security}
 
-This section gives security considerations, as required by {{RFC3552}}.
+Security considerations for JSONPath can stem from
 
+* attack vectors on JSONPath implementations, and
+* the way JSONPath is used in security-relevant mechanisms.
 
+## Attack vectors on JSONPath Implementations
+
+Historically, JSONPath has often been implemented by feeding parts of
+the query to an underlying programming language engine, e.g.,
+JavaScript.
+This approach is well known to lead to injection attacks and would
+require perfect input validation to prevent these attacks (see
+{{Section 12 of -json}} for similar considerations for JSON itself).
+Instead, JSONPath implementations need to implement the entire syntax
+of the query without relying on the parsers of programming language
+engines.
+
+Attacks on availability may attempt to trigger unusually expensive
+runtime performance exhibited by certain implementations in certain
+cases.
+(See {{Section 10 of -cbor}} for issues in hash-table implementations,
+and {{Section 8 of -iregexp}} for performance issues in regular
+expression implementations.)
+Implementers need to be aware that good average performance is not
+sufficient as long as an attacker can choose to submit specially
+crafted JSONPath queries that trigger surprisingly high, possibly
+exponential, CPU usage.
+
+## Attacks on Security Mechanisms that Employ JSONPath
+
+Where JSONPath is used as a part of a security mechanism, attackers
+can attempt to evoke unexpected behavior, or take advantage of
+differences in behavior between implementations that implement some
+protection and the implementations that are being protected by these.
+
+This also applies to underlying technologies such as UTF-8 (see
+{{Section 10 of -utf8}}), the Unicode character set, and JSON.
+A characteristic of JSON that can lead to diverging results is the
+fact that JSON objects are unordered; therefore, the order in which
+results of a JSONPath query reflect the presence of JSON object
+members can vary with implementations.
 
 --- back
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1394,8 +1394,7 @@ exponential, CPU usage.
 
 Where JSONPath is used as a part of a security mechanism, attackers
 can attempt to evoke unexpected behavior, or take advantage of
-differences in behavior between implementations that implement some
-protection and the implementations that are being protected by these.
+differences in behavior between JSONPath implementations.
 
 This also applies to underlying technologies such as UTF-8 (see
 {{Section 10 of -utf8}}), the Unicode character set, and JSON.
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
...
+sufficient as long as an attacker can choose to submit specially
+crafted JSONPath queries that trigger surprisingly high, possibly
+exponential, CPU usage.
+
+## Attacks on Security Mechanisms that Employ JSONPath
+
+Where JSONPath is used as a part of a security mechanism, attackers
+can attempt to evoke unexpected behavior, or take advantage of
+differences in behavior between implementations that implement some
+protection and the implementations that are being protected by these.
```
glyn, line 1398: I don't understand the contrast which is being made here: "implementations that implement some
protection" vs "implementations that are being protected by these [protections?]". Was the following the intended sense?
```suggestion
protection and the implementations that do not.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
...
+
+This also applies to underlying technologies such as UTF-8 (see
+{{Section 10 of -utf8}}), the Unicode character set, and JSON.
+A characteristic of JSON that can lead to diverging results is the
```
glyn, line 1402: "diverging" carries connotations of failure. Suggestion:
```suggestion
A characteristic of JSON that can lead to varying results is the
```

glyn: Another option:
```suggestion
A characteristic of JSON that can lead to non-deterministic results is the
```

draft-ietf-jsonpath-base.md
cabo: The intention was to distinguish the protector from the protected.

draft-ietf-jsonpath-base.md
cabo: Well, it *is* failure :-)  But varying is fine, too.  Didn't want to spend "non-deterministic" here, because for some people that sounds like someone is throwing a coin (which, however, can figuratively exactly what's happening).

draft-ietf-jsonpath-base.md
cabo: ... if the protector is trying to protect the protected from the presence of certain characteristics, but the protected can be made to see different ones than the protector sees, the protection isn't going to work well.  Tons of CVEs of this class, often called check-use (but different from TOCTOU which is a race condition).

glyn: Hmmm. It's very confusing. Is the "difference in behaviour" due to the protector and the protected using different implementations of JSONPath?

draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
 
 # Security Considerations {#Security}
 
-This section gives security considerations, as required by {{RFC3552}}.
-
-
+Security considerations for JSONPath can stem from
+
+* attack vectors on JSONPath implementations, and
+* the way JSONPath is used in security-relevant mechanisms.
```
glyn, line 1368: The following is simpler, if it's accurate:
```suggestion
* the way JSONPath is used in security checks.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
...
+require perfect input validation to prevent these attacks (see
+{{Section 12 of -json}} for similar considerations for JSON itself).
+Instead, JSONPath implementations need to implement the entire syntax
+of the query without relying on the parsers of programming language
+engines.
+
+Attacks on availability may attempt to trigger unusually expensive
+runtime performance exhibited by certain implementations in certain
+cases.
+(See {{Section 10 of -cbor}} for issues in hash-table implementations,
```
timbray, line 1385: I think it would be simpler to point to the security-considerations section or I-Regexp, which describes the use of toxic regexps to accomplish denial-of-service, and how the use of I-Regexp somewhat addresses that problem.

draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
...
+
+Historically, JSONPath has often been implemented by feeding parts of
+the query to an underlying programming language engine, e.g.,
+JavaScript.
+This approach is well known to lead to injection attacks and would
+require perfect input validation to prevent these attacks (see
+{{Section 12 of -json}} for similar considerations for JSON itself).
+Instead, JSONPath implementations need to implement the entire syntax
+of the query without relying on the parsers of programming language
+engines.
```
timbray, line 1380: This is related to the common security command with JSON-formatted protocol elements: Thou shalt not use eval().

But… I'm thinking of evaluating JSONpaths in JS and it's not obvious where it'd be attractive to feed fragments to eval() or equivalent. So I'm not 100% convinced this needs to be said.  Or is there an obvious example that I'm missing?

draft-ietf-jsonpath-base.md
```diff
@@ -1361,9 +1362,47 @@ Provisional registration? (standards tree only):
...
+can attempt to evoke unexpected behavior, or take advantage of
+differences in behavior between implementations that implement some
+protection and the implementations that are being protected by these.
+
+This also applies to underlying technologies such as UTF-8 (see
+{{Section 10 of -utf8}}), the Unicode character set, and JSON.
+A characteristic of JSON that can lead to varying results is the
+fact that JSON objects are unordered; therefore, the order in which
+results of a JSONPath query reflect the presence of JSON object
+members can vary with implementations.
```
timbray, line 1405: Fair enough, but why is this a security consideration?

cabo: In any check/use situation.
(A protective application [that checks data to protect another application] cannot rely on the protected application [that uses the data] to make exactly the same recognition decisions.)

timbray: Does this presuppose that the protected application needs to be protected from objects with members that appear in a particular order?  I'm really having trouble imagining a situation where this makes a difference that could constitute a vulnerability.

cabo: Let's say the protective application needs to avoid data that say {"rocket": "launch"} ever reaching the protected application.  Let's say the two differ in the way they handle duplicate keys.

Or let's say the protected handles the JSON data as it comes in, with a streaming parser, giving the attacker way more leverage to create check/use situations.

(Generally speaking, the problem with "proof by lack of imagination" in security considerations is that the set of attackers collectively have way more imagination than we do.)

draft-ietf-jsonpath-base.md
glyn, line 1398: @cabo How about this more radical simplification which generalises the protector/protected scenario and IMO is much easier to digest?
```suggestion
Where JSONPath is used as a part of a security mechanism, attackers
can attempt to evoke unexpected behavior, or take advantage of
differences in behavior between JSONPath implementations.
```

goessner (approved):

PR #175, opened 2022-03-19
Filter and list inline examples

glyn:
Ref: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/136

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1181,18 +1181,6 @@ The following table lists filter expression operators in order of precedence fro
 
 The `filter-selector` works with arrays and objects exclusively. Its result might be a list of *zero*, *one*, *multiple* or *all* of their element or member values then. Applied to other value types, it will select nothing.
 
-Some examples:
-
-| JSON |  Query | Result | Comment |
-|:----|:---------:|:------:|:-------|
-| `{"a":1,"b":2}`<br>`[2,3,4]` | `$[?@]` | `[1,2]`<br>`[2,3,4]` | Same as `$.*` or `$[*]`  |
-| `./.` | `$[?@==2]` | `[2]`<br>`[2]` | Select by value.  |
-| `{"a":{"b":{"c":{}}}` | `$[?@.b]`<br>`$[?@.b.c]` | `[{"b":{"c":{}}]` | Existence  |
-| `{"key":false}` | `$[?index(@)=='key']`<br>`$[?index(@)==0]` | `[false]`<br>`[]` | Select object member |
-| `[3,4,5]` | `$[?index(@)==2]`<br>`$[?index(@)==17]` | `[5]`<br>`[]` | Select array element |
-| `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
-
-
 #### Examples
 {: unnumbered}
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1195,10 +1195,10 @@ Queries:
 
 | Query | Result | Result Paths | Comment |
 | :---: | ------ | :----------: | ------- |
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1193,6 +1193,23 @@ Some examples:
...
+    [3, 5, 1, 2, 4, 6, {"a": "ij"}, {"a": "ik"}]
+
+Queries:
+
+| Query | Result | Result Paths | Comment |
+| :---: | ------ | :----------: | ------- |
+| `$[?@>3.5]` | `5` <br> `4` <br> `6` | `$[1]` <br> `$[4]` <br> `$[5]` | Comparison |
+| `$[?@.a]` | `{"a": "ij"}` <br> `{"a": "ik"}` | `$[6]` <br> `$[7]` | Existence |
+| `$[?@<2 || @.a == "ik"]` | `1` <br> `{"a": "ik"}` | `$[2]` <br> `$[7]` | Logical OR |
+| `$[?@.a =~ "i.*"]` | `{"a": "ij"}` <br> `{"a": "ik"}` | `$[6]` <br> `$[7]` | Regular expression |
```
goessner, line 1210: Regex should be `"i[j,k]"` or more generally  `"i[a-z]"` iirc ... thanks for your effort.

glyn: Why? Suppose my _intention_ was to pick any value beginning with `i`. Is there some value in having one regex rather than another?

draft-ietf-jsonpath-base.md
```diff
@@ -1193,6 +1193,23 @@ Some examples:
 | `{"a":{"b":{5},c:0}}` | `$[?@.b==5 && !@.c]` | `[{"b":{5},c:0}]` | Existence  |
 
 
+#### Examples
+{: unnumbered}
+
+JSON document:
+
+    [3, 5, 1, 2, 4, 6, {"a": "ij"}, {"a": "ik"}]
+
```
goessner, line 1202: We probably also need another set of examples based on a JSON object ... perhaps ...
```
JSON document

{ "a": 1, "b": 2, "c": 3, "d": 5, "e", 7, "u": {"a": 1, "b": 2, "c": 3, "d": 5, "e", 7} }

Some Queries:
   [?@ >1 && @ < 4]
   [?@*.a || @.x]
   ...
```
Note, that we can access first five members via their values only. Index access would require ...

```
[?@index =~ "[b,c,d]"
```
or something like that ...

draft-ietf-jsonpath-base.md
goessner: ahh ... `.*` ... means zero or more arbitrary characters  ... I missed that ... yes you are right.

glyn: No problem. I thought you had some objection to `.*`.

draft-ietf-jsonpath-base.md
glyn: Ok, good suggestion. However `[?@*.a || @.x]` looks (a) syntactically invalid because of `@*` (did you mean `@.*?`) and (b) a bit strange because the `|| @.x` doesn't really add anything of interest (or was that the point?).

goessner: ... no simple typo. `[?@.a || @.x]` is meant to address existence only.

glyn: @goessner See b8d73680cc579ed757ee142c84364b1d3fae231e.

goessner (approved):

goessner:
we  need to discuss the descendant selector's behavior by including itself to the node list.

glyn:
> we need to discuss the descendant selector's behavior by including itself to the node list.

Yes. Let's do that in a PR for https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/177.

goessner:
yes ... it's illustratively shown, what a good thing inline examples are ...

PR #176, opened 2022-03-19
Normalized Path inline examples

glyn:


timbray (commented):
In the first two examples, the Path and Normalized path are equivalent in what they identify. Not so for the third because the Path produces a plural result and thus cannot be normalized. This should probably be noted?

gregsdennis (approved):

glyn:
> In the first two examples, the Path and Normalized path are equivalent in what they identify. Not so for the third because the Path produces a plural result and thus cannot be normalized. This should probably be noted?

`$.a.b[1:2]` can only produce a singular result since the slice `[1:2]` is equivalent to the index `[1]`. Agreed?

timbray:
> 

D'oh of course

Issue #171, opened 2022-03-15
Specify handing of null


glyn:
The current spec is silent about how `null` is handled in many situations:

1. What is the result of applying the JSONPath `$` to the JSON document `null`?
2. What is the result of applying the JSONPath `$.*` to the JSON document `{"a": null}`? The [comparison consensus](https://cburgmer.github.io/json-path-comparison/results/dot_notation_on_null_value.html) implies a node set with the single value `null`.
3. What is the result of applying the JSONPath `$[*]` to the JSON document `[null]`? The [comparison consensus](https://cburgmer.github.io/json-path-comparison/results/bracket_notation_with_wildcard_on_null_value_array.html) implies a node set with the single value `null`.
4. What is the result of applying the JSONPath `$[0]` to the JSON document `[null]`? Presumably a node set with the single value `null`.
5. What is the result of applying the JSONPath `$[1:2]` to the JSON document `["a", null, "b"]`? Presumably a node set with the values `"a"` and `null`.
6. What is the result of applying the JSONPath `$[?@]` to the JSON document `["a", null, "b"]`?
7. What is the result of applying the JSONPath `$[?@.a]` to the JSON document `[{"a": null}]`?
8. What is the result of applying the JSONPath `$.null` to the JSON document `{"null": "value"}`? The [comparison consensus](https://cburgmer.github.io/json-path-comparison/results/dot_notation_with_key_named_null.html) implies a node set with the single value `"value"`.

cburgmer:
To 1.: Similar is https://cburgmer.github.io/json-path-comparison/results/root_on_scalar_false.html
To 6.: https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_current_object.html
To 7.: https://cburgmer.github.io/json-path-comparison/results/filter_expression_without_value.html

gregsdennis:
I have historically handles `null` as a valid value (contrary to most of .net) separate from "undefined" or "missing."  As a result, I agree with these points.

Specifically for [7], this is an existence test. Since `a` exists as a property in the object, that object would be returned.  I believe we decided that existence testing does not consider falsiness.

Issue #92, opened 2021-03-29
bracket notation


goessner:
Also addressing Glyn's point 

> 3. "script" expressions (`()`) - I'd hope we could offer minimal support for these without too much contention.

from #78 here, obviously common agreement exists about the "bracket child operator", selecting a child value
* either from a JSON object by `"name"`
* or from a JSON array by `index`.

So in `$..[<expr>]` the result of expression `<expr>` will be interpreted as a member name or element index depending on the parent value type. While accepting this important statement, let's have a look at @danielaparker's JSON example in [#88](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/88#issuecomment-808522662): 

```
{
  "firstName": "John",
  "address"  : {
    "city" : "Nara",
  }
}
```
Applying following queries gives us

| expression | result |
|:--:|:--:|
|`$.firstName`| `["John"]` |
|`$['firstName']`| `["John"]` |
|`$[@.firstName] === $['John']`| `[]` |
|`$.address.city`| `["Nara"]` |
|`$['address']['city']`| `["Nara"]` |
|`$['address.city']`| `[]` |
|`$[@.address.city] === $['Nara']`| `[]` |

according to the *member name* or *element index* interpretation principle.

In general any query expression ala `$.x[@.a.b.c]` means:

"Find *member name* or *element index* of JSON value named `"x"` as the value of its descendant node location `'@.a.b.c'`". 

From my point of view it is of very low practical use to bury *member name* or *element index* of any JSON value deep down in its subtree.

My `[@.length-1]` index arithmetic is also generally commented of being of low practical use. 

Using root value symbol `$` in `$.x[$.id]` meaning *member name* or *element index* referencing might be of some practical use, but potentially leads to security issues.

As a consequence of this I propose to completely dispense use of `$` and `@` in other expressions than filter expressions.

goessner:
... leaving out the parenthesis in `[()]` above, was intended here for better readability only ...

glyn:
> As a consequence of this I propose to completely dispense use of `$` and `@` in other expressions than filter expressions.

+1

gregsdennis:
A few things.  First,

> ... leaving out the parenthesis in [()] above, was intended here for better readability only ...

I don't think we should shortcut notation in our discussions.  Doing so will end up with shortcutted notation becoming actual notation.  Please use the full notation intended.

[...]

PR #181, opened 2022-04-22
Semantics of `null`

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/171

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1264,6 +1264,32 @@ Queries:
 | `$[0, 0]` | `"a"` <br> `"a"` | `$[0]` <br> `$[0]` | Duplicated entries |
 {: title="List selector examples"}
 
+## Semantics of `null`
+
+JSON `null` is treated as a valid value, rather than "undefined" nor "missing."
+
+### Examples
+{: unnumbered}
+
+JSON document:
+
+    {"a": null, "b": [null], "c": [{}], "null": 1}
+
+Queries:
+
+| Query | Result | Result Paths | Comment |
+| :---: | ------ | :----------: | ------- |
+| `$.a` | `null` | `$['a']` | Object value |
+| `$.a[0]` | | | `null` used as array |
+| `$.a.d` | | | `null` used as object |
+| `$.b[0]` | `null` | `$['b'][0]` | Array value |
+| `$.b[*]` | `null` | `$['b'][0]` | Array value |
+| `$.b[?@]` | `null` | `$['b'][0]` | Existence |
+| `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
+| `$.c[?(@.d==null)]` | | | Comparison with "missing" value |
+| `$.null` | `1` | `$['null']` | Object key |
+{: title="Examples involving null"}
+
 ## Normalized Paths
 
 A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node. For example,
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1266,7 +1266,7 @@ Queries:
 
 ## Semantics of `null`
 
-JSON `null` is treated as a valid value, rather than "undefined" nor "missing."
+Note that JSON `null` is not  treated differently from any other JSON value (there is no special treatment as "undefined" or "missing").
 
 ### Examples
 {: unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1266,7 +1266,7 @@ Queries:
 
 ## Semantics of `null`
 
-Note that JSON `null` is not  treated differently from any other JSON value (there is no special treatment as "undefined" or "missing").
+Note that JSON `null` is treated the same as any other JSON value: it is not taken to mean "undefined" or "missing".
 
 ### Examples
 {: unnumbered}
@@ -1287,7 +1287,6 @@ Queries:
 | `$.b[?@]` | `null` | `$['b'][0]` | Existence |
 | `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
 | `$.c[?(@.d==null)]` | | | Comparison with "missing" value |
-| `$.null` | `1` | `$['null']` | Not JSON null at all, just a string as object key |
 {: title="Examples involving null"}
 
 ## Normalized Paths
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1264,6 +1264,32 @@ Queries:
...
+| :---: | ------ | :----------: | ------- |
+| `$.a` | `null` | `$['a']` | Object value |
+| `$.a[0]` | | | `null` used as array |
+| `$.a.d` | | | `null` used as object |
+| `$.b[0]` | `null` | `$['b'][0]` | Array value |
+| `$.b[*]` | `null` | `$['b'][0]` | Array value |
+| `$.b[?@]` | `null` | `$['b'][0]` | Existence |
+| `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
+| `$.c[?(@.d==null)]` | | | Comparison with "missing" value |
+| `$.null` | `1` | `$['null']` | Object key |
```
cabo, line 1290:
```suggestion
| `$.null` | `1` | `$['null']` | Not JSON null at all, just a string as object key |
```

draft-ietf-jsonpath-base.md
```diff
@@ -1264,6 +1264,32 @@ Queries:
...
+{: title="Examples involving null"}
```
cabo, line 1291: (which is not strictly true as the last one doesn't...)

cabo (commented):
Good addition!  Please see my detailed comments.

draft-ietf-jsonpath-base.md
```diff
@@ -1264,6 +1264,32 @@ Queries:
 | `$[0, 0]` | `"a"` <br> `"a"` | `$[0]` <br> `$[0]` | Duplicated entries |
 {: title="List selector examples"}
 
+## Semantics of `null`
+
+JSON `null` is treated as a valid value, rather than "undefined" nor "missing."
```
danielaparker, line 1269: It's unclear what is meant by JSON null being a "valid" value, valid in what sense? as opposed to an "invalid" value?

This is about the semantics of the _JSONPath spec_, not the semantics of JSON `null`. For the example given of a query that doesn't evaluate to any results, `$.a[0]`, the spec writers could reasonably choose from several alternatives: (1) stay within the JSON type system, and have it evaluate to one of the JSON values, possibly `null` (like JMESPath), or even `false`, (2) extend the JSON type system with additional values such as `undefined` (following JavaScript), or `optional`, and have it evaluate to that, or (3) treat it as an error (like Jayway JSONPath).  None of these alternatives makes JSON `null` more valid or less valid.

For comparison, in original Goessner JsonPath, it is specified that JSONPath expressions return either an `array` or `false`, "false in case of no match."

Would you write "Unlike in original Goessner JsonPath, in IETF JsonPath `false` is treated as a valid value rather than undefined or missing"? I wouldn't think so.

cabo: Yes.  See my comment above.
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/181#discussion_r856064412

draft-ietf-jsonpath-base.md
glyn: I'll scrap that example as it doesn't add much value.

cabo: I actually like it (students would certainly appreciate falling into the trap once and being helped out of it right away); it just requires being careful with the text around it.

draft-ietf-jsonpath-base.md
glyn: Thanks @danielaparker and @cabo. This is fixed with an edited form of @cabo's suggestion.

draft-ietf-jsonpath-base.md
glyn: Yes, happy to add it back with a better description.

cabo:
```suggestion
{: title="Examples involving (or not involving) null"}
```

timbray (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1264,6 +1264,32 @@ Queries:
...
+
+| Query | Result | Result Paths | Comment |
+| :---: | ------ | :----------: | ------- |
+| `$.a` | `null` | `$['a']` | Object value |
+| `$.a[0]` | | | `null` used as array |
+| `$.a.d` | | | `null` used as object |
+| `$.b[0]` | `null` | `$['b'][0]` | Array value |
+| `$.b[*]` | `null` | `$['b'][0]` | Array value |
+| `$.b[?@]` | `null` | `$['b'][0]` | Existence |
+| `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
```
gregsdennis, line 1288: I believe these need parentheses

```suggestion
| `$.b[?(@)]` | `null` | `$['b'][0]` | Existence |
| `$.b[?(@==null)]` | `null` | `$['b'][0]` | Comparison |
```

We had discussed making them optional somewhere, but I don't think we elected to allow it.

gregsdennis (dismissed):

draft-ietf-jsonpath-base.md
glyn: @gregsdennis These examples merely follow the [current ABNF](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#name-filter-selector) in which those parentheses are still optional. Happy to add the parentheses if we agree the ABNF should be changed to mandate the parentheses, but I think it would be best to discuss that separately to this PR (which may have a rather small audience).

cabo (approved):

draft-ietf-jsonpath-base.md
gregsdennis: If that's the case, the ABNF needs to be fixed.  We never agreed to allow these parentheses to be omitted.

The closest related discussion on this I could find is https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/92#issuecomment-809331011 and https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/92#issuecomment-809601525.  These comments clearly indicate that the parentheses _should_ be present.

glyn: @gregsdennis  I think you should take this up on the mailing list or via a new issue.

glyn: @gregsdennis If you are comfortable with these examples reflecting the current spec, please approve so I can merge.

Issue #167, opened 2022-03-09
Improve filter selector section


glyn:
The notes need separating out into paragraphs.

The semantics section needs to be expanded.

PR #182, opened 2022-04-22
Improve filter selector section

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/167

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1212,7 +1212,7 @@ filter selector iterates over the array or object.
 Comparisons using one of the operators `<`, `<=`, `>`, and `>=` are between numeric values only.
 Using these operators to compare other types of values produces a "false" comparison result.
 
-The semantics of regular expressions is as defined in {{-iregexp}}.
+The semantics of regular expressions are as defined in {{-iregexp}}.
 
 #### Examples
 {: unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1119,15 +1119,40 @@ basic-expr        = exist-expr /
                     paren-expr /
                     relation-expr
 exist-expr        = [neg-op S] singular-path          ; path existence or non-existence
+~~~~
+
+Paths in filter expressions are Singular Paths, each of which selects at most one node.
+
+~~~~ abnf
 singular-path     = rel-singular-path / abs-singular-path
 rel-singular-path = "@" *(S (dot-selector / index-selector))
 abs-singular-path = root-selector *(S (dot-selector / index-selector))
+~~~~
+
+Parentheses can be used with `boolean-expr` for grouping. So filter selection syntax in the original proposal `[?(<expr>)]` is naturally contained in the current lean syntax `[?<expr>]` as a special case.
+
+~~~~ abnf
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1115,54 +1115,75 @@ basic-expr        = exist-expr /
...
 comp-expr    = comparable S comp-op S comparable
 comparable   = number / string-literal /              ; primitive ...
                true / false / null /                  ; values only
                singular-path                          ; Singular Path value
 comp-op      = "==" / "!=" /                          ; comparison ...
                "<"  / ">"  /                          ; operators
                "<=" / ">="
+~~~~
+
+Alphabetic characters in ABNF are case-insensitive, so "e" can be either "e" or "E".
```
timbray, line 1161: With RFC 7405 you can say %s and get case-sensitive matching.  But I think this is OK?

draft-ietf-jsonpath-base.md
```diff
@@ -1171,15 +1192,23 @@ The following table lists filter expression operators in order of precedence fro
...
-The `filter-selector` works with arrays and objects exclusively. Its result might be a list of *zero*, *one*, *multiple* or *all* of their element or member values then. Applied to other value types, it will select nothing.
+The `filter-selector` works with arrays and objects exclusively. Its result is a list of *zero*, *one*, *multiple* or *all* of their array elements or member values, respectively. Applied to other value types, it will select nothing.
+
+A relative path, beginning with `@`, refers to the current array element or member value as the
+filter selector iterates over the array or object.
+
+Comparisons using one of the operators `<`, `<=`, `>`, and `>=` are between numeric values only.
+Using these operators to compare other types of values produces a "false" comparison result.
+
+The semantics of regular expressions is as defined in {{-iregexp}}.
```
timbray, line 1211: "is" should be "are", semantics is plural

cabo: se·man·tics | səˈmæn(t)ɪks |
plural noun [usually treated as singular]

(But I won't contradict native speakers here...)

glyn: https://www.collinsdictionary.com/dictionary/english/semantics claims that "semantics" functions as singular in British English, is unclear about American English, but then goes on to cite the example "They are also logically meaningful, they have well-defined truth conditions, and their semantics are unambiguous." This sounds plain wrong to my British English ear, but the spec should be in American English, so I'll change it.

draft-ietf-jsonpath-base.md
glyn: I think it's ok. We don't require a normalized form of `number`. For example, I think the following all have the same meaning in this context: `100`, `100.0`, `100.00`, etc., `1e2`, `1.0e2`, `1.00e2`, etc., `1e02`, `1e002`, etc., `1000e-1`, `10000e-2`, etc. (and then the forms using `E` instead of `e`).

draft-ietf-jsonpath-base.md
glyn: Fixed (after rebasing) with the observation that "semantics" is already plural elsewhere in the spec.

goessner (approved):

goessner (approved):

goessner (approved):

PR #187, opened 2022-04-27
String matching is codepoint by codepoint, no normalization

timbray:
Per discussion at recent interim.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -555,6 +555,14 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Filter selector `[?(<expr>)]`
   * Current item selector `@` (used in expressions)
 
+Note that processing the dot selector and string-valued index selectors
+requires matching strings from the JSONPath against member names and
+string values from the JSON to which it is being applied.  Two strings
+MUST be considered equal if and only if they are identical sequences
+of Unicode code points. In other words, normalization operations MUST
+NOT be applied to either the string from the JSONPath or from the JSON
+prior to comparison.
+
 ### Root Selector
 
 #### Syntax
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -555,12 +555,13 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Filter selector `[?(<expr>)]`
   * Current item selector `@` (used in expressions)
 
-Note that processing the dot selector and string-valued index selectors
-requires matching strings from the JSONPath against member names and
-string values from the JSON to which it is being applied.  Two strings
-MUST be considered equal if and only if they are identical sequences
-of Unicode code points. In other words, normalization operations MUST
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -555,6 +555,14 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Filter selector `[?(<expr>)]`
   * Current item selector `@` (used in expressions)
 
+Note that processing the dot selector and string-valued index selectors
```
gregsdennis, line 558: Do we want to include comparisons in expressions?

timbray: Right, as in filter selectors. Should definitely mention them.  Now, I haven't used them so I have a basic question. All our examples are of the form 

`$.foo == "bar"`

so the statement that you're comparing a string from the JSON to one from the JSONPath is still true.  But is it also legal to have a filter like `$.foo == $.bar`?  In that case you're comparing two strings from the JSON.  Which doesn't change the code points rule of course.

glyn: Either side of `==` is a `comparable` which, for strings, could be either a string literal or a path beginning with `$` or `@`. So string comparisons include, for example:
* `"x" == "x"`
* `"x" == $.foo`
* `$.foo == "x"`
* `$.foo == $.bar`
* `@.foo == "x"`
* `@.foo == @.bar`
* `@.foo == $.bar`
* `$.bar == @.foo`

(There are also examples which would evaluate to false because a string is being compared to another type, such as `"true" == true`.)

goessner: Despite `$.foo == $.bar` is valid in filters, it hardly makes sense. The current value `@` during iteration process plays no role.

goessner (approved):

glyn (changes requested):
Filter comparisons need to be included.

draft-ietf-jsonpath-base.md
glyn: > Despite `$.foo == $.bar` is valid in filters, it hardly makes sense. The current value `@` during iteration process plays no role.

It could make some kind of sense, e.g. to select an array element if and only if two values in the document happen to be equal. Comparisons such as `"x" == "x"` make less sense because they are constant predicates, true or false.

timbray: Actually, I can imagine something like `$.submittedBy == $.resourceOwner`

(actually more likely != but the point stands)

Will update the PR

glyn (commented):
Removing veto since I am going on vacation.

Please include filter comparisons before merging.

glyn (approved):

glyn (changes requested):
Filter comparisons need to be included.

@timbray ok?

draft-ietf-jsonpath-base.md
cabo: > Will update the PR

Time for that update, I'd say...

timbray:
Oops yes thanks done.

glyn (approved):

Issue #190, opened 2022-05-30
Typos in index selector examples


glyn:
The following result path of two examples is missing a `j`:
```
$['o']['j ']['k.k']
```

Issue #189, opened 2022-05-26
Queries of properties containing special characters


csuwildcat:
I'm seeking clarification on whether the current spec takes into account queries targeting properties that contain characters that may overlap with characters used for various purposes within JSONPath. The clearest example of this that we have run into in other spec work that seeks to rely on JSONPath is the case of properties that contain the `@` sign. This often presents when attempting to target the `@context` property in JSON-LD documents.

I'm hoping the spec 1) does or will allow a means to query by properties that contain such characters, and 2) includes examples/tests showing how such queries are correctly constructed. If this is already allowed by the spec, thank you in advance, and I'd love a pointer to the example/test. If this isn't yet permitted, I'd be happy to help do the work to account for it in spec text and tests.

cabo:
`$['@context']`

cabo:
That said, JSONPath is of somewhat limited value for querying JSON-LD as it works on the JSON level, not on the RDF level.  A proper query to an JSON-LD document would need to perform the JSON-LD processing rules before operating on the resulting data model level items (triples).

csuwildcat:
> That said, JSONPath is of somewhat limited value for querying JSON-LD as it works on the JSON level, not on the RDF level. A proper query to an JSON-LD document would need to perform the JSON-LD processing rules before operating on the resulting data model level items (triples).

Well aware of the processing phase of JSON-LD, but in this case the checks are for context URI presence to identify that the general object is inclusive of a general type of data. The fact the standard spec clarifies and allows for this is great, and means the class of issues folks have seen with some libraries are more conformance issues at the lib level than anything indicative of a spec concern. Thank you for your answer, I'd consider this good to close (unless there's a desire to include a specific example or reference of what you clarified in the spec).

gregsdennis:
I believe there are tests for this already.  I'm pretty sure it's in the comparison project, at least.

glyn:
I added an example with the characters `'` and `@` in member names. In doing so, I observed that some implementations didn't cope very well, at least with `'`. In particular, [Jayway](https://jsonpath.herokuapp.com/) when presented with the JSON document:
```
    {
      "'": 1
    }
``` 
and the JSONPath `$["'"]` gave the correct matching value (`1`) but gave an invalid normalized path expression (`$[''']` - the inner `'` is not escaped).

glyn:
> I believe there are tests for this already. I'm pretty sure it's in the comparison project, at least.

[...]

PR #191, opened 2022-05-30
Add example of member name containing @

glyn:
Also fix a couple of typos.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/190
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/189


gregsdennis (approved):

cabo (approved):
As the "j j" fix demonstrates, we really need to auto-check these.
Eyeballing-wise, this change looks good.

PR #192, opened 2022-05-31
Editorial changes to Normalized Paths

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -1386,6 +1386,8 @@ normal-element-index      = "0" / (DIGIT1 *DIGIT) ; non-negative decimal integer
 | `$.a` | `$['a']` | Object value |
 | `$[1]` | `$[1]`  | Array index |
 | `$.a.b[1:2]` | `$['a']['b'][1]` | Nested structure |
+| `$["\u000B"]`| `$['\u000b']` | Unicode escape |
+| `$["\u0061"]`| `$['a']` | Unicode character |
```
glyn, line 1390: I would have preferred a special symbol (e.g. `±`) or unusual character (e.g. `å`), but I couldn't find one which rendered correctly in text and HTML.

draft-ietf-jsonpath-base.md
```diff
@@ -1365,7 +1365,7 @@ normal-escapable          = ( %x62 / %x66 / %x6E / %x72 / %x74 / ; \b \f \n \r \
                                 ; r /         ;  CR carriage return U+000D
                                 "'" /         ;  ' apostrophe U+0027
                                 "\" /         ;  \ backslash (reverse solidus) U+005C
-                                (%x75 normal-hexchar) ;  certain values u00XX U+00XX
+                                (%x75 normal-hexchar) ;  certain values u00xx U+00XX
```
glyn, line 1368: `u00XX` gave the impression that the hex is in upper case, which it is not.

draft-ietf-jsonpath-base.md
cabo: See my fix (this is probably as good as it gets with RFCXMLv3)

goessner (approved):

glyn:
@cabo and @goessner Thanks for your help. I verified cut'n'past works correctly.

PR #206, opened 2022-06-25
Index wild selector examples fix

graza:


graza:
Forgive me if I'm going about this the wrong way.  I don't use GitHub a great deal.

gregsdennis (approved):
Nice catch.

glyn (approved):
Good catch - thanks!

glyn:
> Forgive me if I'm going about this the wrong way. I don't use GitHub a great deal.

This PR is a great way to contribute.

Issue #123, opened 2021-09-17
Is absent equal to absent?


cabo:
If yes,

`@.doesnotexist != @.alsodoesnotexist` 

is false.

Need to discuss impact on >, >=, etc.

See also:
https://notes.ietf.org/notes-ietf-interim-2021-jsonpath-03-jsonpath

cburgmer:
I've extended the query which compares two paths via equals to include absent keys: https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_equals_with_path_and_path.html. No consensus, not even a majority at this point.

cabo:
... which is not necessarily a problem: I gives us more leeway to nail down something sane (if we want to).

goessner:
... with JavaScript ...
```js
const o = {};
o.a === o.b;  // true
```
due to ` undefined === undefined`.

danielaparker:
> I've extended the query which compares two paths via equals to include absent keys: https://cburgmer.github.io/json-path-comparison/results/filter_expression_with_equals_with_path_and_path.html. No consensus, not even a majority at this point.

Isolating the empty key behavior, I think the results are consistent with what we understand about the implementations. By my count, 12 implementations select the `{}` item. It depends on whether the implementation returns a value (such as `undefined`) for an absent key, or whether it handles it as a recoverable error. The Javascript ones that return `undefined` for an absent key select `{}`, as would be expected. Jayway, which handles an absent key as a recoverable error that takes out the term, does not select `{}`, again as expected.     

None select `{"key1": null}` or `{"key2": null}`. I think it would be interesting to add `{"key1": false}` to the document.

cburgmer:
> I think it would be interesting to add {"key1": false} to the document.

Done :)

gregsdennis:
I would like to add here, just to ensure that it's documented, that C# recently added `x is {}` as an expression equivalent to `x != null`.

Therefore returning `{}` could be quite confusing for some users.

danielaparker:
> > I think it would be interesting to add {"key1": false} to the document.
> 
> Done :)

Okay, from the data (checked cells correspond to selected values), it appears that twelve implementations consider absent is equal to absent, while one, PHP (galbar-jsonpath), considers absent is equal to false but not to absent.

The ones that consider absent is equal to absent do so for different reasons. Cpp (jsoncons) and dotNET (JsonCons.JsonPath)
[...]

PR #207, opened 2022-06-25
Comparison of absent values

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/123


draft-ietf-jsonpath-base.md
```diff
@@ -1226,7 +1226,11 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
 A relative path, beginning with `@`, refers to the current array element or member value as the
 filter selector iterates over the array or object.
 
-Comparisons using one of the operators `<`, `<=`, `>`, and `>=` are between numeric values only.
+Comparison using one of the operators `==` or `!=` are between primitive values or singular paths. A singular path that selects no nodes compares unequal to a primitive value or a singular path which selects a node and compares
+equal to a singular path that selects no nodes. Two singular paths that each select a node compare equal if and only
+if these nodes hold equal primitive values.
+
+Comparisons using one of the operators `<`, `<=`, `>`, or `>=` are between numeric values only.
```
gregsdennis, line 1233: Do we not want to include strings here?  (Note that we would need to define some kind of character ordering, e.g. ordinal, case-insensitive, etc.)

glyn: @gregsdennis  I think we agreed to avoid "ordering comparisons" of strings. Anyway, this is existing text, so if you feel strongly, it deserves a separate issue.

draft-ietf-jsonpath-base.md
```diff
@@ -1226,8 +1221,22 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+When a path resulting in an empty nodelist appears on either side of a comparison, the result of the comparison is
+true if and only if the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
+resulting in an empty nodelist.
+
+When no path resulting in an empty nodelist appears on either side of a comparison, any path which appears on either
+side of the comparison and results in a nodelist consisting of a single node is replaced by the value of the node
+and then:
+
+* comparison using one of the operators `==` or `!=` produce a "true" comparison result if and only if the comparison
+is between primitive values which satisy the comparison.
```
gregsdennis, line 1233: "satisfy"

draft-ietf-jsonpath-base.md
```diff
@@ -1156,11 +1156,6 @@ relation-expr = comp-expr /                           ; comparison test
                 regex-expr                            ; regular expression test
 ~~~~
 
-Comparisons are restricted to Singular Path values and primitive values (such as number, string, `true`, `false`, `null`).
```
gregsdennis, line 1159: Is the "singular path" requirement mentioned elsewhere?  I see it removed here and not added back anywhere.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
glyn: Fixed, thanks.

draft-ietf-jsonpath-base.md
glyn: The semantics section actually implies that if a non-singular path was permissible, the result of the comparison would be "false", but the above sentence was actually talking about the syntax, so I've put it back. Thanks for spotting this.

gregsdennis (approved):

glyn:
Merging to unblock other editorial work.

Issue #208, opened 2022-06-25
Move filter semantics out of syntax section


glyn:
Currently the syntax section of the "Filter Selector" section describes some semantics. These descriptions should be folded into the semantics section.

PR #214, opened 2022-06-29
Move filter semantics out of syntax section

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/208

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1222,10 +1222,8 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
 A relative path, beginning with `@`, refers to the current array element or member value as the
 filter selector iterates over the array or object.
 
-A member or element value by itself in a Boolean context is
-interpreted as `false` only if it does not exist.
-Otherwise it is interpreted as `true`.
-To be more specific about the actual value, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with structured values.
+A singular path by itself in a Boolean context is an existence test with result `true` if the path selects a node and result `false` if the path does not select a node.
+To be more specific about the actual value of a node selected by a path, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with nodes with structured values.
 
 When a path resulting in an empty nodelist appears on either side of a comparison, the result of the comparison is
 true if and only if the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1130,7 +1130,7 @@ filter             = "?" S boolean-expr
 
 During iteration process each array element or object member is visited and its value — accessible via symbol `@` — or one of its descendants — uniquely defined by a relative path — is tested against a boolean expression `boolean-expr`.
 
-The current item is selected if and only if the result is `true`.
+The current item is selected if and only if the boolean expression yields true.
 
 
 ~~~~ abnf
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1224,6 +1212,11 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
 A relative path, beginning with `@`, refers to the current array element or member value as the
 filter selector iterates over the array or object.
 
+A member or element value by itself in a Boolean context is
```
cabo, line 1215: That doesn't work for me.
Once something is a value, it is that value; no further conversion happens.
This sentence was originally about queries that deliver a node list.
A value that doesn't exist isn't a value, no?

glyn: Yes, I've added a commit with better wording.

timbray (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1253,6 +1244,13 @@ the comparison is between numeric values which satisfy the comparison.
 Note that comparisons between structured values, even if the values are equal, produce a "false" comparison result.
 <!-- issue: comparison with structured value -->
 
+Data types are not implicitly converted in comparisons.
+So `"13 == '13'"` selects no node.
+
+Regular expression tests can be applied to JSON string values
+({{Section 7 of -json}}) only
+(on the left-hand side of `=~`); they yield false otherwise.
```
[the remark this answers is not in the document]
glyn: Yes, that's better thanks. Adopted.

I also like the phrase "yields true" and have adopted it throughout for results of boolean expressions.

draft-ietf-jsonpath-base.md
```diff
@@ -1234,30 +1222,39 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
 A relative path, beginning with `@`, refers to the current array element or member value as the
 filter selector iterates over the array or object.
 
-When a path resulting in an empty nodelist appears on either side of a comparison, the result of the comparison is
+A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
```
cabo, line 1225: Did we ever clarify that we don't want existence tests on general paths?

draft-ietf-jsonpath-base.md
```diff
@@ -1234,30 +1222,39 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+To be more specific about the actual value of a node selected by a path, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with nodes with structured values.
```
cabo, line 1226: s/be more specific about the actual/make use of the JSON/

draft-ietf-jsonpath-base.md
```diff
@@ -1234,30 +1222,39 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+
+When a path resulting in an empty nodelist appears on either side of a comparison, the comparison yields
 true if and only if the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
```
cabo, line 1229: So != is always false, while == can be true?
Doesn't look right.

cabo (commented):
I'm not sure we are done with this yet.  See comments.

draft-ietf-jsonpath-base.md
glyn: I was under the impression that the consensus was to restrict existence tests to singular paths. That seems to be clear in the spec. Did you want further clarification in the spec or to revisit the question?

draft-ietf-jsonpath-base.md
glyn: Yes, I'll put that in a new PR, thanks.

draft-ietf-jsonpath-base.md
glyn: I tried fixing this in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/224.

draft-ietf-jsonpath-base.md
glyn: Please see https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/224.

PR #218, opened 2022-07-06
descendant ordering and clarification

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -1073,7 +1071,7 @@ The `descendant-selector` selects certain descendants of a node:
 In the resultant nodelist:
 
 * nodes occur before their children, and
-* nodes of an array occur in array order.
+* nodes of a given array, and descendants of nodes of the given array, occur in array order.
```
gregsdennis, line 1074: Is this specific enough to clarify between the two cases I mentioned in the issue?

- complete each item and its descendants before moving to the next item, in array order, vs
- complete the items in the array, in order, then cycle through descendants of members (queue-style)

gregsdennis: It might be worth including the `[[0],[1]]` example.

glyn: > Is this specific enough to clarify between the two cases I mentioned in the issue?
> 
>     * complete each item and its descendants before moving to the next item, in array order, vs
> 
>     * complete the items in the array, in order, then cycle through descendants of members (queue-style)

Yes. The second implementation is unlikely to work without some clever/complex handling of the nodelist. (The reason I say this is that the traversal order doesn't necessarily determine the order of the resultant nodelist. For example, after traversal is complete, the nodelist could be sorted into an order consistent with the spec rules. Not a good implementation, but possible.)

draft-ietf-jsonpath-base.md
```diff
@@ -305,7 +305,7 @@ $.store.book[?(@.price < 10)].title
 | `@`         | the current node: within [filter selectors](#filter-selector)                                                                     |
 | `.name`     | child selectors for JSON objects: [dot selector](#dot-selector)                                                                     |
 | `['name']`  | child selectors for JSON objects: [index selector](#index-selector)                                                                 |
-| `..`        | descendants: [descendant selector](#descendant-selector)                                                                     |
+| `..name` <br> `..[3]` | descendants: [descendant selector](#descendant-selectors)                                                                     |
```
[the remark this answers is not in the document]
glyn: > I'm not sure I agree with this being unified with the name/bracket notation. I get it, in that you can't just have `..` by itself, though. Maybe it's an implementation detail.

What spec change would you like to see? It would require an issue as the change is outside the scope of this PR.

> Secondly, _any_ bracket notation query will work: you could have `$..[?(@.foo)]` to recursively search for objects with `foo` properties.

The spec doesn't currently support filters as part of a descendant selector. Some implementations may support that. An issue would be needed as it's also outside the scope of this PR.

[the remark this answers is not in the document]
gregsdennis: > What spec change would you like to see?

I think what you have is fine to describe the syntax, but my implementation doesn't (and likely won't) actually do it that way.

> The spec doesn't currently support filters as part of a descendant selector.

[looks up spec]  yep.  I'll open a new issue for this.  I think any `[ ... ]` selector should be supported after `..`.

gregsdennis (approved):

Issue #215, opened 2022-07-05
descendant selector non-determinism


glyn:
Currently, the descendant selector spec allows some non-determinism which may be unexpected.

The spec currently says that in the resultant nodelist from the descendant selector:
* nodes occur before their children, and
* nodes of an array occur in array order.

This sometimes allows multiple orderings when the result includes children of a node, but not the node itself.

For example, with the document:
```
[0, [1]]
```
the selector `$..[0]` could, according to the spec, produce the nodelist:
```
[0, 1]
```
or the nodelist:
```
[1, 0]
```

This is surprising given that there are no objects in the document, since object ordering is the usual source of non-determinism.

We could change the spec to constrain the order of traversal rather than the order of the resultant nodelist, but this seems to be getting into implementation detail.

Is the current approach acceptable?

glyn:
> Have I missed something? Is it a matter of updating the semantics to say the entirety of the current node is looked at first?

I think something like that is probably desirable, but without unduly constraining implementations.

> (Side note - the spec is not as clear on the element-index only applying to arrays. It's implied, but not stated as clearly as it is for quoted-member-names and objects.)

Thanks. I'll take a look at that with an editorial PR.

glyn:
I think rewording the second condition ("nodes of an array occur in array order") achieves the desired ordering:
* nodes of a given array, and descendants of nodes of the given array, occur in array order.

gregsdennis:
> I don't see how the result could be `[1,0]`.

I'm still here. I'm not sure that anything needs to be updated. I think the current text (node then descendants) is sufficient to guarantee the correct ordering.

In this case,

- `[0, [1]]` is processed, producing `0`
- then `[1]` is processed, producing `1`

It must be this way because of array ordering.

`..` gives the (ordered) node list `[ [0,[1]], 0, [1], 1 ]`.  Of those, we take the first elements of arrays, which is `0` from `[0,[1]]` and `1` from `[1]`.  I'm really struggling to see how the current wording can yield anything else.

glyn:
The behaviour of [four existing implementations](https://jsonpath.herokuapp.com/) is interesting. With the document:
```
[["b"], ["c"]]
```
and the selector `$..*`, Nebhale produces a result obeying the reworded conditions (but starts the nodelist with `$`):
```
[
   [
      [
         "b"
      ],
      [
         "c"
      ]
   ],
   [
      "b"
   ],
   "b",
   [
[...]

PR #221, opened 2022-07-09
Sharpen up descendant ordering terminology

glyn:
This is probably a bit too wordy, but is a starting point for a sharper definition of the nodelist orders which can be produced by the descendant selector.

Ref: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/215

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1068,7 +1068,7 @@ Note that there is no "bald" descendant selector `..`.
 #### Semantics
 {: unnumbered}
 
-The `descendant-selector` selects certain descendants of a node:
+A `descendant-selector` selects certain descendants of a node:
 
 * the `..<name>` form (and the `..[<index>]` form where `<index>` is a `quoted-member-name`) selects those descendants that are member values of an object with the given member name.
 * the `..[<index>]` form, where `<index>` is an `element-index`, selects those descendants that are array elements with the given index.
@@ -1076,14 +1076,14 @@ The `descendant-selector` selects certain descendants of a node:
 * the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
 * the `..[*]` and `..*` forms select all the descendants.
 
-The resultant nodelist is ordered as if:
+An _array preorder_ of the descendants of a node is a sequence of all the descendants in which:
 
-* nodes are visited before their children, and
-* nodes of an array are visited in array order.
+* nodes of an array appear in array order, and
+* nodes appear before their children.
 
-Children of an object may be visited in any order, since JSON objects are unordered.
+Children of an object may appear in any order, since JSON objects are unordered.
 
-Implementations may visit descendants in any way providing the resultant nodelist could have been produced by visiting descendants as above.
+The resultant nodelist of a `descendant-selector` applied to a node must be a sub-sequence of an array preorder of the descendants of the node.
 
 #### Examples
 {: unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1063,7 +1063,7 @@ descendant-selector = ".." ( dot-member-name      /  ; ..<name>
                            )
 ~~~~
 
-Note that there is no "bald" descendant selector `..`.
+Note that there is no "bald" descendant selector (`..` on its own).
 
 #### Semantics
 {: unnumbered}
@@ -1076,14 +1076,14 @@ A `descendant-selector` selects certain descendants of a node:
 * the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
 * the `..[*]` and `..*` forms select all the descendants.
 
-An _array preorder_ of the descendants of a node is a sequence of all the descendants in which:
+An _array-sequenced preorder_ of the descendants of a node is a sequence of all the descendants in which:
 
-* nodes of an array, immediately followed by their descendants appear in array order, and
-* nodes appear before their children.
+* nodes of any array appear in array order,
+* nodes appear immediately before all their descendants.
 
-Children of an object may appear in any order, since JSON objects are unordered.
+This definition does not stipulate the order in which the children of an object appear, since JSON objects are unordered.
 
-The resultant nodelist of a `descendant-selector` applied to a node must be a sub-sequence of an array preorder of the descendants of the node.
+The resultant nodelist of a `descendant-selector` applied to a node must be a sub-sequence of an array-sequenced preorder of the descendants of the node.
 
 #### Examples
 {: unnumbered}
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1068,22 +1068,22 @@ Note that there is no "bald" descendant selector `..`.
...
+A `descendant-selector` selects certain descendants of a node:
 
 * the `..<name>` form (and the `..[<index>]` form where `<index>` is a `quoted-member-name`) selects those descendants that are member values of an object with the given member name.
 * the `..[<index>]` form, where `<index>` is an `element-index`, selects those descendants that are array elements with the given index.
 * the `..[<slice-index>]` form selects those descendants that are array elements selected by the given slice.
 * the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
 * the `..[*]` and `..*` forms select all the descendants.
 
-The resultant nodelist is ordered as if:
+An _array preorder_ of the descendants of a node is a sequence of all the descendants in which:
```
cabo, line 1079: Hmm, the "preorder" part is universal and not just for arrays.  So qualifying "preorder" with "array" doesn't strike me as helpful.  Simply calling it "preorder" is fine, but I wanted to add an adjective that indicates that this conventional order is one of many possible (JSON objects), which the implementation is free to choose, and that it is the order we have chosen for JSONPath.

glyn: Yeah, I was trying to capture an additional constraint on the order: it is both a preorder (parents before children) and arrays are traversed in array order (not implied by preorder).

So I tried to introduce a new term "array preorder" which is a special kind of preorder.

I'm not sure about an adjective conveying the lack of ordering of objects. Even if we found a suitable adjective, we'd still need to build in the array condition, ideally into the term, somehow.

cabo: After some more thinking, I don't get a better result than "array-ordered pre-order".  Maybe  bit redundant and wordy, but we are not using this term all over the place.

glyn: That's certainly explicit, but the ordered/order part doesn't sound good. Do any of the following grab you?

* array-based preorder
* arraywise preorder
* array-respecting preorder
* arrayful preorder
* linear preorder
* vector preorder

cabo: Surprisingly hard.

* array-sequenced pre-order
* array-guided pre-order

draft-ietf-jsonpath-base.md
```diff
@@ -1068,22 +1068,22 @@ Note that there is no "bald" descendant selector `..`.
...
 
-* nodes are visited before their children, and
-* nodes of an array are visited in array order.
+* nodes of an array appear in array order, and
+* nodes appear before their children.
```
cabo, line 1082: hmm, this doesn't actually say depth first.

..* on
[1, 2, 3, [4, 5], 6]

could be shown as [1, 2, 3, [4, 5], 6, 4, 5] and satisfy the two rules.

draft-ietf-jsonpath-base.md
```diff
@@ -1068,22 +1068,22 @@ Note that there is no "bald" descendant selector `..`.
...
 * the `..[<slice-index>]` form selects those descendants that are array elements selected by the given slice.
 * the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
 * the `..[*]` and `..*` forms select all the descendants.
 
-The resultant nodelist is ordered as if:
+An _array preorder_ of the descendants of a node is a sequence of all the descendants in which:
 
-* nodes are visited before their children, and
-* nodes of an array are visited in array order.
+* nodes of an array appear in array order, and
```
cabo, line 1081:
```suggestion
* nodes of an array, immediately followed by their descendants appear in array order, and
```

Need to say the same for objects and their descendants.

draft-ietf-jsonpath-base.md
glyn: "array-sequenced preorder" seems to be the best term so far. (I think preorder is usually unhyphenated. See, for example, [wikipedia](https://en.wikipedia.org/wiki/Preorder).)

draft-ietf-jsonpath-base.md
glyn: Good point. I'll address that shortly.

sthagen (approved):
In case `a` and `an` Indicate `any` or `for every` this change looks good to me.

draft-ietf-jsonpath-base.md
```diff
@@ -1063,27 +1063,27 @@ descendant-selector = ".." ( dot-member-name      /  ; ..<name>
                            )
 ~~~~
 
-Note that there is no "bald" descendant selector `..`.
+Note that there is no "bald" descendant selector (`..` on its own).
```
cabo, line 1066: I'm not sure we need to introduce that term (even though, informally, I find it amusing).

```suggestion
Note that `..` on its own is not a valid selector.
```

cabo (approved):
(Optional comment.)

draft-ietf-jsonpath-base.md
```diff
@@ -1063,27 +1063,27 @@ descendant-selector = ".." ( dot-member-name      /  ; ..<name>
...
 
 #### Semantics
 {: unnumbered}
 
-The `descendant-selector` selects certain descendants of a node:
+A `descendant-selector` selects certain descendants of a node:
```
sthagen, line 1071: Does `A` Indicate `any` or `for every`?

draft-ietf-jsonpath-base.md
```diff
@@ -1063,27 +1063,27 @@ descendant-selector = ".." ( dot-member-name      /  ; ..<name>
...
 
 * the `..<name>` form (and the `..[<index>]` form where `<index>` is a `quoted-member-name`) selects those descendants that are member values of an object with the given member name.
 * the `..[<index>]` form, where `<index>` is an `element-index`, selects those descendants that are array elements with the given index.
 * the `..[<slice-index>]` form selects those descendants that are array elements selected by the given slice.
 * the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
 * the `..[*]` and `..*` forms select all the descendants.
 
-The resultant nodelist is ordered as if:
+An _array-sequenced preorder_ of the descendants of a node is a sequence of all the descendants in which:
```
sthagen, line 1079: Does  `An` Indicate `any` or `for every`?

draft-ietf-jsonpath-base.md
glyn: I changed "The" to "A" because there are multiple descendant selectors. It carries the connotation of "any".

draft-ietf-jsonpath-base.md
cabo: Not a good suggestion?

glyn: Whoops! Yes, it was a good suggestion, thanks. Since it's small, I'll piggy-back it on another PR.

glyn: See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/214/commits/c7c245a95302c5e792d34e0ba4c6c9de10e431b2

Issue #210, opened 2022-06-25
"JSON document"


cabo:
We agreed to use the RFC 8259 terms, which would be "JSON text".

cabo:
(And some occurrences actually talk about the JSON value.)

PR #223, opened 2022-07-14
Use "JSON text" term

glyn:
... instead of "JSON document" or "JSON value".

"JSON value" is retained in some contexts where it refers to specific
values i JSON rather than a whole document.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/210

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -581,7 +581,7 @@ addressed by the root selector `$`.
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
 
     {"k": "v"}
 
@@ -627,7 +627,7 @@ to the member name from any JSON object in its input nodelist. It selects no nod
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
 
     {"j": {"k": 3}}
 
@@ -664,7 +664,7 @@ string, or true/false/null) selects no node.
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
 
     {
       "o": {"j": 1, "k": 2},
@@ -800,7 +800,7 @@ not exist; this simply means that no element is selected.
 <!-- EDITING NOTE: There are non-breaking spaces here between j and j -->
 <!-- i.e., j j and not j j -->
 
-JSON document:
+JSON value:
 
     {
       "o": {"j j": {"k.k": 3}},
@@ -844,7 +844,7 @@ The `index-wild-selector` behaves identically to the `dot-wild-selector`.
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
 
     {
       "o": {"j": 1, "k": 2},
@@ -1027,7 +1027,7 @@ range of exact values, see {{synsem-overview}}.
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
 
     ["a", "b", "c", "d", "e", "f", "g"]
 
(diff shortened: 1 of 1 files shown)


cabo:
JSON text is a sequence of characters, an unparsed piece of JSON for interchange, with newlines and backslashes and all that.
JSONPath **never** looks at JSON texts.
JSON value is the RFC 8259 term for what JSONPath looks at.

cabo:
What should be done here is replacing all occurrences of "JSON document" with "JSON value".

glyn:
@cabo Fixed with a force push. Sorry for misinterpreting the issue.

cabo:
Terminology that would be nice to have could be "the JSON value that represents an entire JSON text".  People are often using "JSON document" for that, but that is ambiguous between the serialization (JSON text) and the data (JSON value).
If the text would really benefit from such a term, maybe we should coin it.

draft-ietf-jsonpath-base.md
```diff
@@ -1398,17 +1398,17 @@ Queries:
...
-`$['book'][3]` and `$['book'][5]`. For a given JSON document, there is a one to one correspondence between the document's
+`$['book'][3]` and `$['book'][5]`. For a given JSON value, there is a one to one correspondence between the value's
 nodes and the Normalized Paths that identify these nodes.
 
 A JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
 
 Since bracket notation is more general than dot notation, it is used to construct Normalized Paths.
 Single quotes are used to delimit string member names. This reduces the number of characters that
-need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON documents.
+need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON values.
```
cabo, line 1408: I think for once we actually mean JSON texts here...

draft-ietf-jsonpath-base.md
```diff
@@ -581,7 +581,7 @@ addressed by the root selector `$`.
 #### Examples
 {: unnumbered}
 
-JSON document:
+JSON value:
```
cabo, line 584: I'm wondering if we should introduce examples (which we do show as JSON texts, after all) simply with "JSON: "

glyn: Make sense. Omit needless words.

draft-ietf-jsonpath-base.md
glyn: Fixed.

draft-ietf-jsonpath-base.md
```diff
@@ -1370,12 +1370,12 @@ Queries:
 
 ## Semantics of `null`
 
-Note that JSON `null` is treated the same as any other JSON value: it is not taken to mean "undefined" or "missing".
+Note that JSON `null` is treated the same as any other JSON: it is not taken to mean "undefined" or "missing".
```
cabo, line 1373: I think this should be "as any other JSON value" here.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: Whoops. Thanks.

gregsdennis:
This was opened and closed in less than 8 hours.  I was asleep that entire time and had no chance to review.

Please be sure that PRs stay open for at least a day so they can be reviewed by everyone.

glyn:
Thanks for the reminder @gregsdennis. Sorry I forgot.

Issue #212, opened 2022-06-26
Behavior for semantically invalid expressions


gregsdennis:
What is the expected behavior for expressions which are not semantically valid?

For example, `<` is only defined for numbers.  So what if someone does

```jsonpath
$[?(@.foo < 'bar')]
```

Is this a parse error?  Does it just evaluate to `false` for all potential `@.foo` values?

This kind of thing will be important to define if/when we decide to support more complex expressions (e.g. with mathematic operators) or inline JSON values such as objects or arrays.

cburgmer:
Here you can see the current behaviour of the various implementations for a similar case: https://cburgmer.github.io/json-path-comparison/#filter_expression_with_greater_than

glyn:
The spec says:

> The well-formedness and the validity of JSONPath queries are independent of the JSON value the query is applied to; no further errors can be raised during application of the query to a value.

and the grammar in the spec includes `$[?(@.foo < 'bar')]` as syntactically valid. (We could tighten up the grammar to exclude cases where non-numeric literals appear in ordered comparisons, but I don't think there's a great benefit in doing so.)

The spec also says:

> Comparisons using one of the operators <, <=, >, and >= are between numeric values only. Using these operators to compare other types of values produces a "false" comparison result.

So, `@.foo < 'bar'` evaluates to "false".

graza:
I don't understand numeric only semantic constraint.  Is there a reason behind it?  

JSON has very few value types and lexical ordering of strings is well defined in Unicode.  Standards like ISO8601, commonly used for representing date-time values in JSON, give consideration to lexical ordering.  So a useful filter for me could be to return JSON objects that have a date greater than (or equal) to a date taken from somewhere else in the same instance.

Having the comparison evaluate to false could also be confusing.  For `{ "a" : 6, "b" : "c" }`, what should `?!(@ >= 5)` yield for `$["b"]`?  Should it be true?  If so such an expression would return `[ "c" ]` which is a non-numeric result even though a numeric comparison was performed.  Or should breaking the semantic rule about numerics cause the entire predicate to yield false, regardless?

gregsdennis:
> Or should breaking the semantic rule about numerics cause the entire predicate to yield false, regardless?

[...]

PR #224, opened 2022-07-15
Follow on to PR 214

glyn:
* Editorial improvements
* Correct != comparison involving at least one empty nodelist.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1235,10 +1235,9 @@ the comparison is between numeric values which satisfy the comparison.
 of the same values using the operator `==`, `<=`, `>=`, respectively.
 
 * data types are not implicitly converted in comparisons.
-So, for example, `13 == '13'` yields false.
 
-Note that `==` comparisons between structured values, even if the values are equal, yield false.
-Also `!=` comparisons between structured values, even if the values are equal, yield true.
+Note that `==` comparisons between a structured value and any value, including the same structured value, yield false.
+Also `!=` comparisons between a structured value and any value, including the same structured value, yield true.
 <!-- issue: comparison with structured value -->
 
 ##### Regular Expressions
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1238,7 +1238,7 @@ the comparison is between numeric values which satisfy the comparison.
 of the same values using the operator `==`, `<=`, `>=`, respectively.
 
 Note that `==` comparisons between a structured value and any value, including the same structured value, yield false.
-Consequently `!=` comparisons between a structured value and any value, including the same structured value, yield true.
+Consequently, `!=` comparisons between a structured value and any value, including the same structured value, yield true.
 <!-- issue: comparison with structured value -->
 
 ###### Examples
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1207,8 +1207,11 @@ This existence test — as an exception to the general rule — also works with 
 To make use of the JSON value of a node selected by a path, explicit comparisons are necessary.
 
 When a path resulting in an empty nodelist appears on either side of a comparison, the comparison yields
-true if and only if the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
-resulting in an empty nodelist.
+true if and only if:
+
+* the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
+resulting in an empty nodelist, or
+* the comparison operator is `!=` and the other side of the comparison is not also a path resulting in an empty nodelist.
 
 When no path resulting in an empty nodelist appears on either side of a comparison, any path which appears on either
 side of the comparison and results in a nodelist consisting of a single node is replaced by the value of the node
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1208,12 +1208,9 @@ filter selector iterates over the array or object.
 A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
 This existence test — as an exception to the general rule — also works with nodes with structured values.
 
-A non-existence test (`!` followed by a singular path by itself, e.g. `!@.foo`) yields true if and only if the corresponding existence test with the same path yields true.
-In other words, a non-existence test yields true if the path does not select a node and yields false if the path selects a node.
-
 To test the value of a node selected by a path, an explicit comparison is necessary.
 For example, to test whether the node selected by the path `@.foo` has the value `null`, use `@.foo == null` (see {{null-semantics}})
-rather than the non-existence test `!@.foo` (which yields false if `@.foo` selects a node, regardless of the node's value).
+rather than the negated existence test `!@.foo` (which yields false if `@.foo` selects a node, regardless of the node's value).
 
 ##### Comparisons
 {: unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1204,7 +1204,9 @@ filter selector iterates over the array or object.
 
 A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
 This existence test — as an exception to the general rule — also works with nodes with structured values.
-To make use of the JSON value of a node selected by a path, explicit comparisons are necessary.
+To test the value of a node selected by a path, an explicit comparison is necessary.
+For example, to test whether the node selected by the path `@.foo` has the value `null`, use `@.foo == null` (see {{null-semantics}})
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1203,11 +1203,15 @@ A relative path, beginning with `@`, refers to the current array element or memb
 filter selector iterates over the array or object.
 
 A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
-To be more specific about the actual value of a node selected by a path, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with nodes with structured values.
+This existence test — as an exception to the general rule — also works with nodes with structured values.
+To make use of the JSON value of a node selected by a path, explicit comparisons are necessary.
```
timbray, line 1207: I find this really hard to understand. Might an example help?

timbray (changes requested):

draft-ietf-jsonpath-base.md
glyn: Good idea. I added an example in af06ecf5e3c32f28156d2a39fd9d1ecfd354e5fa and also talk about "the value of a node" rather than suddenly talking about "the JSON value of a node" which seems a little distracting. See what you think.

gregsdennis (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1203,11 +1203,17 @@ A relative path, beginning with `@`, refers to the current array element or memb
...
+rather than the non-existence test `!@.foo` (which yields false if `@.foo` selects a node, regardless of the node's value).
 
 When a path resulting in an empty nodelist appears on either side of a comparison, the comparison yields
-true if and only if the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
-resulting in an empty nodelist.
+true if and only if:
+
+* the comparison operator is `==`, `>=` or `<=` and the other side of the comparison is also a path
+resulting in an empty nodelist, or
+* the comparison operator is `!=` and the other side of the comparison is not also a path resulting in an empty nodelist.
```
cabo, line 1216: I'd rather have `!=` `<` `>` be defined as the opposite of `==` `>=` `<=`.

glyn: Yes, I have a lot of sympathy for that. Before I make the change, let's see if there is a consensus for https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/212#issuecomment-1186563259 as this may be relevant here.

glyn: Consider a path `p` which results in an empty nodelist. Then according to the spec, `p <= 1` and `p >= 1` yield false.

If we define `>`, `<` as the negation of the operators `<=`, `>=`, respectively, then `p > 1` and `p < 1` yield true. This breaks the normal partial order laws (`p < 1` implies `1 > p`, but then we have `1 > p` and `p > 1`, which breaks the asymmetry law that if `a < b` then not `b < a`).

Instead, I'm thinking of keeping any path `p` which results in an empty nodelist **outside** the partial order. Thus `p <= 1`, `p >= 1`, `p > 1`, and `p < 1` would all yield false (as would `p < p` and `p > p`, as expected).

draft-ietf-jsonpath-base.md
```diff
@@ -1202,36 +1202,59 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
 filter selector iterates over the array or object.
 
+##### Existence Tests
+{: unnumbered}
+
 A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
-To be more specific about the actual value of a node selected by a path, explicit comparisons are necessary. This existence test — as an exception to the general rule — also works with nodes with structured values.
+This existence test — as an exception to the general rule — also works with nodes with structured values.
+
+A non-existence test (`!` followed by a singular path by itself, e.g. `!@.foo`) yields true if and only if the corresponding existence test with the same path yields true.
```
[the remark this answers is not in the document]
glyn: Yeah, fair enough. Deleted.

cabo:
I'm not sure that `1 < [2, 3]` is so obviously false, so maybe mentioning that a structured value on one side spoils the comparison may be needed.

draft-ietf-jsonpath-base.md
```diff
@@ -1202,36 +1202,89 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+
+* data types are not implicitly converted in comparisons.
+
+Note that `==` comparisons between a structured value and any value, including the same structured value, yield false.
+Also `!=` comparisons between a structured value and any value, including the same structured value, yield true.
 <!-- issue: comparison with structured value -->
 
-Data types are not implicitly converted in comparisons.
-So `13 == '13'` yields false.
+###### Examples
```
glyn, line 1243: We may wish to move these examples to the Filter selector examples section to keep all the examples in one place.

draft-ietf-jsonpath-base.md
```diff
@@ -1202,36 +1202,89 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+is between equal primitive values.
+
+* comparisons using one of the operators `<=` or `>=` yield true if and only if
 the comparison is between numeric values which satisfy the comparison.
 
-Note that comparisons between structured values, even if the values are equal, yield false.
+* any comparison of two values using one of the operators `!=`, `>`, `<` is defined as the negation of the comparison
+of the same values using the operator `==`, `<=`, `>=`, respectively.
+
+* data types are not implicitly converted in comparisons.
```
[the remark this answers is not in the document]
timbray: Oops, I meant the last bullet point "data types are not implicitly converted in comparisons" - but my suggestion stands: there's no real need to bring the notion of "conversion" in, just assert that all type-incompatible comparisons, and all comparisons involving structured values, return false.

draft-ietf-jsonpath-base.md
```diff
@@ -1202,36 +1202,87 @@ The `filter-selector` works with arrays and objects exclusively. Its result is a
...
+
+* comparisons using one of the operators `<=` or `>=` yield true if and only if
 the comparison is between numeric values which satisfy the comparison.
 
-Note that comparisons between structured values, even if the values are equal, yield false.
+* any comparison of two values using one of the operators `!=`, `>`, `<` is defined as the negation of the comparison
+of the same values using the operator `==`, `<=`, `>=`, respectively.
+
+Note that `==` comparisons between a structured value and any value, including the same structured value, yield false.
+Also `!=` comparisons between a structured value and any value, including the same structured value, yield true.
```
cabo, line 1238: Maybe s/also/As a consequence/?

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: > This is getting nicer. But I'm not crazy about this title, because what people care about isn't "conversion", it's the result. How about text something along the lines of "All comparisons between values which are not of the same JSON type, or where either or both values are structured, return false."

There is a simple solution: delete the offending bullet. The other bullets imply that comparison of values with distinct JSON types yield false. I've preserved an example of this and changed the comment to talk about type mismatch. Please see ef71e3ef709496e12d49237f7e6e1c3017dee9bf.

> 
> Perhaps the phrase "JSON type" should be footnoted with a reference to https://datatracker.ietf.org/doc/html/rfc8259#section-3

That section doesn't say much about JSON types, but the introduction does:
> JSON can represent four primitive types (strings, numbers, booleans,
   and null) and two structured types (objects and arrays).

Our spec uses the term "type" (meaning JSON type) in many places. For the avoidance of doubt, I added the term in b0c9f2dd2e79da499606e6f62a403f554818a8e4.

draft-ietf-jsonpath-base.md
glyn: I like it.

draft-ietf-jsonpath-base.md
```diff
@@ -1238,7 +1238,7 @@ the comparison is between numeric values which satisfy the comparison.
 of the same values using the operator `==`, `<=`, `>=`, respectively.
 
 Note that `==` comparisons between a structured value and any value, including the same structured value, yield false.
-Also `!=` comparisons between a structured value and any value, including the same structured value, yield true.
+Consequently `!=` comparisons between a structured value and any value, including the same structured value, yield true.
```
cabo, line 1241: [SENT_START_CONJUNCTIVE_LINKING_ADVERB_COMMA] A comma may be missing after the conjunctive/linking adverb ‘Consequently’. -> (Consequently,)

(Thanks for picking this up)

draft-ietf-jsonpath-base.md
cabo: I believe this is now addressed; dismissing this so we can progress.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: > "All comparisons between values which are not of the same JSON type, or where either or both values are structured, return false."

That would not work very well, as you no longer could compare with null (or any other value used as a stand-in value).

PR #225, opened 2022-07-17
We no longer expand "JSON"

cabo:
* because it is no longer needed
* because the expansion is misleading

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -124,7 +124,7 @@ venue:
 --- abstract
 
 JSONPath defines a string syntax for selecting and extracting values
-within a JavaScript Object Notation (JSON, RFC 8259) value.
+within a JSON (RFC 8259) value.
 
 --- middle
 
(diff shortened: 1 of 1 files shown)


glyn (approved):
LGTM, thanks.

gregsdennis:
Why is the expansion misleading?

glyn:
> Why is the expansion misleading?

I would have thought it was because the notation is not just for _objects_.

cabo:
Using the expanded version of the acronym gives the impression that this specification is in some way entangled with JavaScript.  (The fact that Glyn reads "JavaScript Object" as a reference to the string-keyed maps JavaScript started out with is another example for the confusion.)

gregsdennis:
I feel like anyone who knows about JSON had likely come across our asked about its relationship with its namesake.  I don't really think this is something that is needed.

I mean, do it if it makes you feel better, but it's not really necessary.

cabo:
In the RFC series, the expansion is no longer needed, which is a good thing, because the expansion is misleading.
This PR adapts the JSONPath specification to this fact.
I'm not sure what "this" and "it" are in your message.
This PR is not about making anyone of us feel better, it is about improving the document to be less misleading.

glyn:
I think we are all agreed that merging is fine.

Issue #229, opened 2022-07-26
Behaviour for invalid JSON


glyn:
@timbray [wrote](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/212#issuecomment-1195623511):
> Hmm, I don't see anything in the draft spec about what happens if you try to apply a JSONPath and what you're trying to apply it to is broken, i.e. not valid JSON. Maybe we're in a world where callers have to live with failed applications anyhow?

glyn:
It depends. For example, the APIs of some JSONPath implementations take a rich type which represents a valid JSON value. In that case, handling invalid JSON happens before the API is called and is not a concern of the API. But other implementations could have an API which takes a string, which may or may not represent valid JSON.

So, to be clear, we probably need some words in the spec which says the spec does not prescribe how invalid JSON is handled. This is better than silence.

Or we could require that invalid JSON cannot result in a non-empty nodelist from a JSONPath query.

danielaparker:
@timbray
 
> I don't see anything in the draft spec about what happens if you try to apply a JSONPath and what you're trying to apply it to is broken

But "brokenness" would apply to JSON texts, wouldn't it? As @cabo noted in 
(https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/223#issuecomment-1184361289), "JSON text is a sequence of characters, an unparsed piece of JSON for interchange, with newlines and backslashes and all that.
JSONPath never looks at JSON texts. JSON value is the RFC 8259 term for what JSONPath looks at."

timbray:
Agree with last two points. The spec describes the application of a JSONPath expression to parsed JSON, so JSON texts are out-of-scope. As @glyn says, we should make that explicit.

gregsdennis:
JSON Schema simply [defines](https://json-schema.org/draft/2020-12/json-schema-core.html#section-4.2.1) that it operates on the JSON data model.

> JSON Schema interprets documents according to a data model.

This section covers a range of issues that we've found in our history, including handling whitespace differences, duplicate keys, etc.

I think this is a great approach.

PR #231, opened 2022-07-27
Clarify scope regarding invalid JSON

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/229

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -382,7 +382,7 @@ constant.
 
 ## Overview {#synsem-overview}
 
-A JSONPath query is a string which selects zero or more nodes of a piece of JSON.
+A JSONPath query is a string which selects zero or more nodes of a JSON value.
 
 A query MUST be encoded using UTF-8.
 The grammar for queries given in this document assumes that its UTF-8 form is first decoded into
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -421,6 +421,10 @@ type errors when pondering well-formedness and validity, while
 resource depletion and related errors are comparable to 500 type
 errors.)
 
+The JSON value the JSONPath query is applied to is, by definition, valid JSON.
```
cabo, line 424: I'm not sure "valid JSON" is defined.  "a valid JSON value" clarifies this and also exposes the redundancy of this statement.

```suggestion
The JSON value the JSONPath query is applied to is, by definition, a valid JSON value.
```

cabo (approved):
Thanks!

draft-ietf-jsonpath-base.md
```diff
@@ -421,6 +421,10 @@ type errors when pondering well-formedness and validity, while
 resource depletion and related errors are comparable to 500 type
 errors.)
 
+The JSON value the JSONPath query is applied to is, by definition, a valid JSON value.
+The parsing of a JSON text into a JSON value and what happens if a JSON
+text does not represent valid JSON are not defined by this specification.
```
timbray, line 426: perhaps “not addressed” or “not discussed”?

glyn: I chose "not defined" because it implies invalid JSON produces undefined behaviour as far as our spec is concerned.

It wouldn't be true to say "not discussed" although we could say "not discussed further". Similarly for "not addressed". The additional paragraph discusses and addresses the case of invalid JSON, albeit very briefly.

I toyed with describing invalid JSON as being "out of scope" of the specification, but "not defined" seemed clearer.

gregsdennis: A agree with @glyn that "not defined" is more appropriate here.

Issue #232, opened 2022-08-06
Semantics of the list form of descendant selector


glyn:
The [semantics](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/802861e49848e96ff5b0cc087006330757595f04/draft-ietf-jsonpath-base.md#semantics-8) (as of 2022-08-06) for the descendant selector involving a list:

* Do not state which descendants are selected.
* Order selected items contrary to current implementations.
* Lead to contradictions over the ordering of duplicate nodes.

## Which descendants are selected

It is currently implicit that the list selector should select descendants based on the [list semantics](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/802861e49848e96ff5b0cc087006330757595f04/draft-ietf-jsonpath-base.md#semantics-7). This should be made explicit.

## Ordering vs implementation consensus

An [example from the JSONPath comparison project](https://cburgmer.github.io/json-path-comparison/results/union_with_keys_after_recursive_descent.html) applies the path `$..['c', 'd']` to the value:
```
[
 {"c":"cc1","d":"dd1","e":"ee1"},
 {"c": "cc2", "child": {"d": "dd2"}},
 {"c": "cc3"},
 {"d": "dd4"},
 {"child": {"c": "cc5"}}
]
```

The spec states:
> The resultant nodelist of a `descendant-selector` applied to a node must be a sub-sequence of an array-sequenced preorder of the descendants of the node.

Because of the ordering of the array, in all array-sequenced preorders of the above value, `"cc1"`, `"dd1"`, and `"ee1"` appear before `"cc2"` and `"dd2"` which appear before `"cc3"` which appears before `"dd4"` which appears before `"cc5"`. So the order of the result according to the spec is:
```
["cc1", "dd1", "cc2", "dd2", "cc3", "dd4", "cc5"]
```

However, the current consensus according to the comparison project is:
```
["cc1", "cc2", "cc3", "cc5", "dd1", "dd2", "dd4"]
```
where it seems the value is traversed once per item of the list (`['c', 'd']`).

## Ordering of duplicate nodes

Consider the path `$..[0, 1, 0]` applied to the value:
```
[[1, 2]]
```

[...]

PR #233, opened 2022-08-08
Define descendant list semantics

glyn:
Also tweak the wording of the list semantics section.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/232

Reviewers may find the [rendered version](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/232-descendant-list-semantics/draft-ietf-jsonpath-base.html#name-semantics-10) useful.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1334,7 +1334,7 @@ list-entry     =  ( quoted-member-name /
 #### Semantics
 {: unnumbered}
 
-A list selector selects the nodes that are selected by at least one of
+A list selector selects the nodes that are selected by
 the selector entries in the list and yields the concatenation of the
 lists (in the order of the selector entries) of nodes selected by the
 selector entries.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1393,7 +1393,7 @@ Note that `..` on its own is not a valid selector.
 
 A descendant selector selects zero or more descendants of a node.
 
-The descendants are first enumerated in a nodelist known as a _descendant nodelist_, such that:
+A nodelist enumerating the descendants is known as a _descendant nodelist_ when:
 
 * nodes of any array appear in array order,
 * nodes appear immediately before all their descendants.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1391,22 +1391,31 @@ Note that `..` on its own is not a valid selector.
 #### Semantics
 {: unnumbered}
 
-A `descendant-selector` selects certain descendants of a node:
+A descendant selector selects zero or more descendants of a node.
 
-* the `..<name>` form (and the `..[<index>]` form where `<index>` is a `quoted-member-name`) selects those descendants that are member values of an object with the given member name.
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1381,22 +1381,30 @@ Note that `..` on its own is not a valid selector.
...
-A `descendant-selector` selects certain descendants of a node:
+A descendant selector selects zero or more descendants of a node.
 
-* the `..<name>` form (and the `..[<index>]` form where `<index>` is a `quoted-member-name`) selects those descendants that are member values of an object with the given member name.
-* the `..[<index>]` form, where `<index>` is an `element-index`, selects those descendants that are array elements with the given index.
-* the `..[<slice-index>]` form selects those descendants that are array elements selected by the given slice.
-* the `..[<filter>]` form selects those descendants that are array elements or object values selected by the given filter.
-* the `..[*]` and `..*` forms select all the descendants.
-
-An _array-sequenced preorder_ of the descendants of a node is a sequence of all the descendants in which:
```
glyn, line 1392: This PR removes the need for the term "array-sequenced preorder" by talking instead about a nodelist of the descendants with the same properties. I think this is easier to read and gives the same ordering properties.

draft-ietf-jsonpath-base.md
```diff
@@ -1326,7 +1326,7 @@ list-entry     =  ( quoted-member-name /
 #### Semantics
 {: unnumbered}
 
-A list selector selects the nodes that are selected by at least one of
+A list selector selects the nodes that are selected by
```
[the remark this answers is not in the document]
glyn: That wasn't my intention. The "at least one of" wording was vestigial and related to when we were considering removing duplicates. Now that duplicate nodes and values are retained, the wording is no longer needed.

Or am I missing something?

glyn:
@timbray Just a gentle reminder to re-review as I'd like to get this PR merged. Thanks.

draft-ietf-jsonpath-base.md
```diff
@@ -1391,22 +1391,31 @@ Note that `..` on its own is not a valid selector.
...
-
-An _array-sequenced preorder_ of the descendants of a node is a sequence of all the descendants in which:
+A nodelist enumerating the descendants is known as a _descendant nodelist_ when:
 
 * nodes of any array appear in array order,
 * nodes appear immediately before all their descendants.
 
-This definition does not stipulate the order in which the children of an object appear, since JSON objects are unordered.
+This definition does not stipulate the order in which the children of an object appear, since
+JSON objects are unordered.
```
timbray, line 1402: Slightly more correct I think: The order of the members in a JSON object is not information-bearing.

timbray (approved):

draft-ietf-jsonpath-base.md
glyn: The sentence "The order of the members in a JSON object is not information-bearing." is a bit tricky because it refers to "The order of the members in a JSON object" as if that was somehow well defined. Also, I'm not sure how many people would understand the sentence because the term "information-bearing" is a little unusual. I think I'll stick with the informal language for now.

PR #241, opened 2022-08-16
Clarify undefined object comparisons

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -626,7 +633,10 @@ characters — MUST NOT be used with the `dot-selector`.
 {: unnumbered}
 
 The `dot-selector` selects the node of the member value corresponding
-to the member name from any JSON object in its input nodelist. It selects no nodes from any other JSON value.
+to the member name from any JSON object in its input nodelist.
+The behaviour is unpredictable if the JSON object has duplicate names.
```
gregsdennis, line 637: Is the specific callout here (and the others places below) necessary?  It seems the general statement above is sufficient.

glyn: I agree that the general statement above should be sufficient, but I think it is worth having a "reminder" here because someone might assume that if the dot selector's name is unique in an object with duplicate names, then the node identified by that name should be selected.

gregsdennis: If you think it's valuable, sure, let's leave it.

gregsdennis (approved):

goessner (approved):
Totally agree ... good work ... thanks.

draft-ietf-jsonpath-base.md
goessner: I think it's valuable ... "Constant dripping wears away the stone"

cabo: There are no JSON objects with duplicate names, so I don't like this statement at all.
I'll create a PR that rectifies this.

goessner: If we could banish that "duplicate names"-beast from JSONPath all together, that would be definitly a good thing.

But regarding parsers RFC8259 states:

> Many implementations report the last name/value pair
 only. Other implementations report an error or fail to parse the
 object, and some implementations report all of the name/value pairs,
 including duplicates.

If we decide to ignore the last few "some implementations", I would be ok with that.

glyn: I agree an ideal abstract JSON object does not have duplicate names. But I would say JSONPath needs to acknowledge the existence of JSON _implementations_ which retain duplicate names of objects and make it clear that the behaviour is undefined with such implementations.

So perhaps we could use a phrase like "a JSON object implementation with duplicate names" to distinguish between the ideal behaviour and the possibly imperfect reality?

cabo: See #250 for my proposal.

PR #243, opened 2022-08-17
Delete confusing language about Normalized Paths

glyn:
The one to one correspondence sentence gives us most of the desired
properties. The fact that for distinct Normalized Paths there will always
be at least one JSON value which can distinguish the paths is interesting
but not crucial. If it's confusing to some people, then it's probably
a distraction.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1456,10 +1456,6 @@ Since bracket notation is more general than dot notation, it is used to construc
 Single quotes are used to delimit string member names. This reduces the number of characters that
 need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON texts.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1456,10 +1456,6 @@ Since bracket notation is more general than dot notation, it is used to construc
 Single quotes are used to delimit string member names. This reduces the number of characters that
 need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON texts.
 
-The syntax of Normalized Paths is restricted so that there is one and only one way of identifying a given node.
```
cabo, line 1459: This is mostly a restatement of the first paragraph, so I'm fine with removing this paragraph.

cabo (approved):

goessner (commented):
yes ... this is a good minimalistic solution ... thanks

Issue #244, opened 2022-08-17
Ordered comparisons (<, >) of arrays


glyn:
At today's interim meeting (2022-08-17), we agreed to attempt to specify how `<` and `>` would apply to arrays.

This needs to include arrays with objects as elements and arrays with corresponding elements (at the same index) of different types.

glyn:
Closing now PR 248 has been review and closed.

PR #248, opened 2022-08-18
Ordered comparisons (<, >) of arrays

glyn:
Ordered comparisons of arrays are closely modelled on ordered
comparisons of strings, but if either array has a descendent of an
unordered type, the comparison yields false.

Added a note about slimming down the comparison examples table.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/244

Reviewers may find this [rendered version](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/244-array-ordering/draft-ietf-jsonpath-base.html#name-comparisons) useful.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1200,17 +1200,22 @@ is between:
 is not between equal values of the same type.
 
 * a comparison using one of the operators `<`, `<=`, `>`, or `>=` yields true if and only if
-the comparison is between values of the same type which are both numbers or both strings and which satisfy the comparison:
+the comparison is between values of the same type which are both numbers, both strings, or both arrays and which satisfy the comparison:
 
     * numbers expected to interoperate as per {{Section 2.2 of -i-json (I-JSON)}} MUST compare using the normal mathematical ordering;
       numbers not expected to interoperate as per I-JSON MAY compare using an implementation specific ordering
     * the empty string compares less than any non-empty string
     * a non-empty string compares less than another non-empty string if and only if the first string starts with a
       lower Unicode scalar value than the second string or if both strings start with the same Unicode scalar value and
-      the remainder of the first string compares less than the remainder of the second string.
+      the remainder of the first string compares less than the remainder of the second string
+    * the empty array compares less than any non-empty array whose descendants do not include an object, a boolean, or `null`
+    * a non-empty array compares less than another non-empty array if and only if the first element of the first array is
+      less than the first element of the second array or if both arrays start with the same number, string, or
+      array (whose descendants do not include an object, a boolean, or `null`) and
+      the remainder of the first array compares less than the remainder of the second array.
 
 Note that comparisons using any of the operators `<`, `<=`, `>`, or `>=` yield false if either value being
-compared is an object, array, boolean, or `null`.
+compared is an object, a boolean, `null`, or an array whose descendants include an object, a boolean, or `null`.
 
 ###### Examples
 {: unnumbered}
@@ -1220,7 +1225,11 @@ JSON:
 
     {
       "obj": {"x": "y"},
-      "arr": [2, 3]
+      "arr": [2, 3],
+      "empty": [],
+      "b": [true],
+      "c": [2],
+      "d": [2, 1]
     }
 
 | Comparison | Result | Comment |
@@ -1242,17 +1251,21 @@ JSON:
 | `$.arr != $.arr` | false | Array comparison |
 | `$.obj == 17` | false | Type mismatch |
 | `$.obj != 17` | true | Type mismatch |
-| `$.obj <= $.arr` | false | Objects and arrays are not ordered |
-| `$.obj < $.arr` | false | Objects and arrays are not ordered |
+| `$.obj <= $.arr` | false | Type mismatch |
+| `$.obj < $.arr` | false | Type mismatch |
 | `$.obj <= $.obj` | false | Objects are not ordered |
-| `$.arr <= $.arr` | false | Arrays are not ordered |
-| `1 <= $.arr` | false | Arrays are not ordered |
-| `1 >= $.arr` | false | Arrays are not ordered |
-| `1 > $.arr` | false | Arrays are not ordered |
-| `1 < $.arr` | false | Arrays are not ordered |
+| `1 <= $.arr` | false | Type mismatch |
+| `1 >= $.arr` | false | Type mismatch |
+| `1 > $.arr` | false | Type mismatch |
+| `1 < $.arr` | false | Type mismatch |
 | `true <= true` | false | Booleans are not ordered |
 | `true > true` | false | Booleans are not ordered |
+| `$.c < $.arr`| true | Array comparison |
+| `$.d < $.arr` | true | Array comparison |
+| `$.arr <= $.arr` | true | Array comparison |
+| `$.empty < $.b`| false | Booleans are not ordered |
 {: title="Comparison examples" }
+<!-- May be worth deleting some examples from the above table -->
 
 ##### Regular Expressions
 {: unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1227,7 +1227,7 @@ JSON:
       "obj": {"x": "y"},
       "arr": [2, 3],
       "empty": [],
-      "b": [true],
+      "b": [2, 3, 1, [7, true]],
       "c": [2],
       "d": [2, 1]
     }
@@ -1263,9 +1263,12 @@ JSON:
 | `$.c < $.arr`| true | Array comparison |
 | `$.d < $.arr` | true | Array comparison |
 | `$.arr <= $.arr` | true | Array comparison |
-| `$.empty < $.b`| false | Booleans are not ordered |
+| `$.c < $.b`| false | Booleans are not ordered |
 {: title="Comparison examples" }
 <!-- May be worth deleting some examples from the above table -->
+
+Note: The `$.c < $.b` example shows that the descendants of each array need to be
+taken into account in ordered comparisons.
 
 ##### Regular Expressions
 {: unnumbered}


draft-ietf-jsonpath-base.md
```diff
@@ -1188,17 +1188,22 @@ is between:
...
 
     * numbers expected to interoperate as per {{Section 2.2 of -i-json (I-JSON)}} MUST compare using the normal mathematical ordering;
       numbers not expected to interoperate as per I-JSON MAY compare using an implementation specific ordering
     * the empty string compares less than any non-empty string
     * a non-empty string compares less than another non-empty string if and only if the first string starts with a
       lower Unicode scalar value than the second string or if both strings start with the same Unicode scalar value and
-      the remainder of the first string compares less than the remainder of the second string.
+      the remainder of the first string compares less than the remainder of the second string
+    * the empty array compares less than any non-empty array whose descendants do not include an object, a boolean, or `null`
+    * a non-empty array compares less than another non-empty array if and only if the first element of the first array is
```
cabo, line 1200:
```suggestion
    * a non-empty array compares less than another non-empty array if and only if the first element of the first array compares
```

glyn:
**Reviewers please note:** I intend to make an improvement if/when this PR lands: see https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/249. Since the improvement may make the text longer, I'd like to work with the current text while we decide whether we want to support ordered comparisons of strings at all. If this PR is merged, I'd then like to fix https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/249 in a similar way for strings and arrays.

draft-ietf-jsonpath-base.md
```diff
@@ -1188,17 +1188,22 @@ is between:
...
     * the empty string compares less than any non-empty string
     * a non-empty string compares less than another non-empty string if and only if the first string starts with a
       lower Unicode scalar value than the second string or if both strings start with the same Unicode scalar value and
-      the remainder of the first string compares less than the remainder of the second string.
+      the remainder of the first string compares less than the remainder of the second string
+    * the empty array compares less than any non-empty array whose descendants do not include an object, a boolean, or `null`
+    * a non-empty array compares less than another non-empty array if and only if the first element of the first array compares
+      less than the first element of the second array or if both arrays start with the same number, string, or
+      array (whose descendants do not include an object, a boolean, or `null`) and
+      the remainder of the first array compares less than the remainder of the second array.
```
gregsdennis, line 1203: Can this be reworded so that it's not a recursive (albeit elegant) description?

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
glyn: It could be reworded to avoid the recursive definition, but I don't think the resultant text would be as clean, especially given the "non-local effect" of descendants which are objects, booleans, or `null`.

What is your concern? That the definition is hard to understand? That we might tempt implementers to use recursion? Or something else?

gregsdennis: My concern is that recursion is best read in code or math rather than English.  It takes a few brain cycles to understand what's going on: first that it's a recursive definition, then second to parse that definition.

I also think it could influence implementors.

goessner: I already stated in interim meeting, that I do have objections against `</>` comparisons of different arrays.

## Arrays containing Numbers only

With arrays (seen as vectors mathematically) mostly a distinct meaning of members is implied:
* `[x,y,z]`  ... Spatial vector.
* `[r,g,b]` ... RGB color.

There is no  linear order in the plane, in the space or in any n-dimensional structure. So with the rgb color example, we won't say, that `blue` is smaller than `green` or `green` is smaller than red, i.e. `[0,0,255] < [0,255,0]`.

Of course, we could apply lexicographical ordering here, but many people would be confused by this.  In fact there is a lexicographical [ordering of vector spaces](https://en.wikipedia.org/wiki/Ordered_vector_space) defined. But from my experience in math and geometry it is of rare practical use. 

Only looking at complex numbers, which can be seen as planar vectors `[a,b]` [ordering them](https://web.maths.unsw.edu.au/~angell/articles/complexorder.pdf) is proofed as [not possible](https://math.stackexchange.com/questions/487997/total-ordering-on-complex-numbers).

## Arrays containing Strings only

The same applies to an array of strings. there must be a reason not to create a single string here.

## Arrays containing no Numbers, no Strings exclusively

I think we all agree on the fact that the relation `<` or `>` always should yield false.

glyn: > I already stated in interim meeting, that I do have objections against `</>` comparisons of different arrays.
> ## Arrays containing Numbers only
> 
> With arrays (seen as vectors mathematically) mostly a distinct meaning of members is implied:
> 
>     * `[x,y,z]`  ... Spatial vector.
> 
>     * `[r,g,b]` ... RGB color.
> 
> 
> There is no linear order in the plane, in the space or in any n-dimensional structure. So with the rgb color example, we won't say, that `blue` is smaller than `green` or `green` is smaller than red, i.e. `[0,0,255] < [0,255,0]`.
> 
> Of course, we could apply lexicographical ordering here, but many people would be confused by this. In fact there is a lexicographical [ordering of vector spaces](https://en.wikipedia.org/wiki/Ordered_vector_space) defined. But from my experience in math and geometry it is of rare practical use.

You may be right, but why do you say many people be confused by this? Since there isn't a single obvious way of ordering arrays, they shouldn't be surprised by our choice of a lexicographic ordering.


> 
> Only looking at complex numbers, which can be seen as planar vectors `[a,b]` [ordering them](https://web.maths.unsw.edu.au/~angell/articles/complexorder.pdf) is proofed as [not possible](https://math.stackexchange.com/questions/487997/total-ordering-on-complex-numbers).

It is [possible](https://shaileshk.medium.com/an-order-on-complex-numbers-6357c8b4aec2) to define a total order on complex numbers if you regard them as merely equivalent to planar vectors. The above proof that this cannot be done uses the "field" properties of complex numbers, such as `i*i = -1`.

glyn:
The reason for the restriction that array comparisons yield false if either array has a descendant which is an object, a boolean, or `null` is that without this restriction, transitivity is broken. For example, without this restriction `[0, true] < 1` would yield true as would `1 < [2, 1]`. If `<` was to remain transitive, then `[0, true] <  [2, 1]` would yield false, but it would actually yield false because the comparison fails on the second element.

(I am thinking of adding a note to the text about this to clarify the rationale.)

draft-ietf-jsonpath-base.md
glyn: > My concern is that recursion is best read in code or math rather than English. It takes a few brain cycles to understand what's going on: first that it's a recursive definition, then second to parse that definition.
> 
> I also think it could influence implementors.

Fair enough, thanks. I'm going to defer improving the text for now in case we throw out the whole PR....

goessner:
Ok ... let's assume, we want to adopt lexicographic order `[a,b] < [c,d]  if  a<c or a=c and b<d` for arrays, analogous to strings, then few questions arise:

* `'ab' < 'abc'` yields `true`, but `['a','b'] < ['a','b','c']` will yield `false` ... 
* how to deal with nested arrays and possible edge cases ?
* why not use [product order](https://en.wikipedia.org/wiki/Product_order)  `[a,b] < [c,d]  if  a<c and b<d` instead ?

The benefit for users is questionable at best (I would like to get some use cases documented) ... while the effort for implementers (and spec writers :-) is significant.

So I propose ...
1. to allow general relations `==,!=,<,>,<=,>=` for numbers and strings.
2. comparison of all other types support `==, !=` and yield `false` otherwise (mixed types!).

... as long as no strong demand from users and/or implementers is heard.

glyn:
> Ok ... let's assume, we want to adopt lexicographic order `[a,b] < [c,d] if a<c or a=c and b<d` for arrays, analogous to strings, 

Note that the above is not the same spec that this PR provides. This PR is more like this:
```
for all values a, c and arrays b, d where a, b, c, and d and their descendants
consist only of numbers, strings, and arrays:
    [] < [a] ^ b
    [a] ^ b < [c] ^ d if a < c or (a == c and b < d)
```

(Perhaps this strengthen's @gregsdennis's point that the recursive definition is hard to read.)

> then few questions arise:
> 
>     * `'ab' < 'abc'` yields `true`, but `['a','b'] < ['a','b','c']` will yield `false` ...

Not so. `['a','b'] < ['a','b','c']` yields true in this PR.

> 
>     * how to deal with nested arrays and possible edge cases ?

Nested arrays are straightforward provided they and their descendants consist only of numbers, strings, and arrays.

I think the main edge cases are where one or other value being compared is not, or has a decendent which is not, a number, string, or array. In all such cases `<` yields false.

> 
>     * why not use [product order](https://en.wikipedia.org/wiki/Product_order)  `[a,b] < [c,d]  if  a<c and b<d` instead ?

Because that seems of (even) less practical use than lexicographic order.

> 
> 
> The benefit for users is questionable at best (I would like to get some use cases documented) ... while the effort for implementers (and spec writers :-) is significant.

I agree about the lack of benefit for users. The main use case we've had so far is multi-string date ordering, although it has been pointed out that RFC 3339 strings provide a better approach for dates. I'm going to resist the temptation to try to dream up more use cases.

> 
> So I propose ...
> 
>     1. to allow general relations `==,!=,<,>,<=,>=` for numbers and strings.
> 
>     2. comparison of all other types support `==, !=` and yield `false` otherwise (mixed types!).
> 
> 
> ... as long as no strong demand from users and/or implementers is heard.

IIRC you are proposing the status quo. To implement your proposal we would simply need to close this PR unmerged. I'm happy to do that if there is a rough consensus in favour.

glyn:
@cabo, @gregsdennis, @timbray, and others: would any of you care to defend ordered comparisons (<, >) of arrays? If not, we can safely close it unmerged.

timbray:
Glyn, you have made a valiant effort here, and thanks for that, but yes, let's please go back to just != and == for structured values.  In the future if someone asks why JSONpath does not include ordering, we now have a good answer as to why.

cabo:
> @cabo, @gregsdennis, @timbray, and others: would any of you care to defend ordered comparisons (<, >) of arrays? If not, we can safely close it unmerged.

I'm sorry, I haven't found anything that is problematic about ordering arrays yet.
Obviously, it is easy to come up with unworkable ways of doing that.
The trick is to start from a very small set of axioms.

As in (add axioms relating < and ==, same as in strings)

* [] == []
* [] < [x]
* [x, ...y] <=> [x, ...z]  ≝  [...y] <=> [...z]
* [x, ...y] <=> [z, ...w]  ≝  x <=> z if x ≠ z
 
That should do it.

cabo:
What is the `^` that occurs in some of the comments above?

timbray:
> As in (add axioms relating < and ==, same as in strings)
> 
> * [] == []
> * [] < [x]
> * [x, ...y] <=> [x, ...z]  ≝  [...y] <=> [...z]
> * [x, ...y] <=> [z, ...w]  ≝  x <=> z if x ≠ z

I don't understand that.  Could it be expressed in human-readable language?

cabo:
> I don't understand that. Could it be expressed in human-readable language?

Yes.  But pure English is a bad tool for discussing this.  We can translate to English once we understand the structure that we are trying to achieve.

timbray:
> > I don't understand that. Could it be expressed in human-readable language?
> 
> Yes. But pure English is a bad tool for discussing this. We can translate to English once we understand the structure that we are trying to achieve.

OK, but I still don't understand. I'm not being rhetorical here, I have a math degree (granted, a miserable B.Sc., decades old) and I don't understand your notation.

cabo:
OK.  The notation essentially tries to answer whether a <=> b, where a and b a JSON values.  <=> is one of the comparison operators we want to define -- usually, they all work the same, so saying something about <=> just says its true for all of them.
...x and ...y pick up the rest of an array; I could have used [x | y] or some similar notation that split an array into first and rest.

The point is to define comparison by looking at the first elements; if there is no first element on at least one side, the first two rules operate, if there is, the second two rules do.
(Oh, and axioms such as mutual exclusion of a < b, a == b, a > b, and a < b ≝ b > a.)

draft-ietf-jsonpath-base.md
cabo: > Can this be reworded so that it's not a recursive (albeit elegant) description?

I don't think this is fundamentally possible for a format such as JSON that is essentially recursively defined.
(We can try to hide it, but boolean-expr already is recursive.)

glyn:
> What is the `^` that occurs in some of the comments above?

Oops, sorry, I meant to define that. It's array concatenation.

glyn:
@cabo `<=>` in maths usually means "if and only if", which makes the axioms above quite hard to read. That's why I wrote down axioms just for `<` and left the rest not very far from the imagination.

But, anyway, the point here is not that ordering of arrays is particularly hard to specify, but that lexicographic ordering is somewhat arbitrary and not very useful in practice.

cabo:
> @cabo `<=>` in maths usually means "if and only if", which makes the axioms above quite hard to read. That's why I wrote down axioms just for `<` and left the rest not very far from the imagination.

OK, so maybe use some generic operator, such as ⊜.
(<=> is used on some dynamic languages as the generic comparison operator, that's why I used it.)

> But, anyway, the point here is not that ordering of arrays is particularly hard to specify, but that lexicographic ordering is somewhat arbitrary and not very useful in practice.

Well, it is analogous to strings, so I don't know that is particularly arbitrary.
Whether it is useful depends on what you use arrays for; arrays of course can be used for a lot of applications only some of which will benefit from lexicographic ordering; but then that is still more than not having ordering at all.

glyn:
> > But, anyway, the point here is not that ordering of arrays is particularly hard to specify, but that lexicographic ordering is somewhat arbitrary and not very useful in practice.
> 
> Well, it is analogous to strings, so I don't know that is particularly arbitrary. Whether it is useful depends on what you use arrays for; arrays of course can be used for a lot of applications only some of which will benefit from lexicographic ordering; but then that is still more than not having ordering at all.

I guess allowing strings and arrays to be members of the arrays we are trying to order actually limits the options quite a bit. For example, comparing numeric arrays by their length when considered to be a vector wouldn't really make sense for arrays of arrays. So perhaps lexicographic ordering isn't _that_ arbitrary.

In terms of usefulness, I found @timbray's comment at the interim meeting persuasive: "if I saw lexicographic array comparison in a code review, I'd be concerned about it". I think the number of reasonable applications of lexicographic array ordering is likely very small.

Another concern I haven't mentioned yet is the "non-local effect" of objects, booleans, and `null`. This would require complete scanning of array arguments of ordered comparisons to make sure none of their descendants was of the wrong type. (Without this "non-local effect", we lose transitivity as I [mentioned above](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/248#issuecomment-1221325225).) Clearly this could be cached, but it does marginally complicate implementations.

Checking existing implementations, I've only found one ([Goessner](https://jsonpath.herokuapp.com/), and he's against this feature!) which behaves similarly to this PR. Similarly, JMESPath does not support ordered comparisons of arrays. So this PR would seem to be diverging from our charter given that we don't have a rough consensus that the approach is technically best.

Closing.

goessner:
> As in (add axioms relating < and ==, same as in strings)
> 
>     * [] == []
> 
>     * [] < [x, ...y]
> 
>     * [x, ...y] <=> [x, ...z]  ≝  [...y] <=> [...z]
> 
>     * [x, ...y] <=> [z, ...w]  ≝  x <=> z if x ≠ z
> 

This is a very clean and complete definition, thanks.

With arrays we always have an implicit meaning regarding their elements. So for example

* `[1,2,3]` ... spatial coordinates
* `[0,128,255]` ... RGB values
* `[2022,08,22]` ... date

only the third example qualifies to be ordered lexicographic. In the same way, as it might be coverted to a string and compared to another date. Noone would apply linear ordering to the first two "objects". So the JSON author defines by implicit meaning of her array elements, if ordering of arrays of that same type makes sense.

We as spec authors don't know about meaning of elements of an arbitrary array. We only offer lexicographic ordering in general for those, who might use it (hopefully those that don't, won't do that accidentially).

But what, if we now make the meaning explicit?

* `{"x":1,"y":2,"z":3}`  ... spatial coordinates
* `{"r":0,"g":128,"b":255}` ... RGB values
* `{"y":2022,"m":08,"d":22}` ... date

How could we argue, not offering lexicographic ordering for objects also? The same axioms apply ...

* `{} == {}`
* `{} < {x, ...y}`
* `{x, ...y} <=> {x, ...z}  ≝  {...y} <=> {...z}`
* `{x, ...y] <=> {z, ...w}  ≝  x <=> z if x ≠ z`

(Excuse for not using `""` here)

"Objects themself are not ordered" is no valid argument, as we are able to judge (member-wise) equality of objects.

I cannot see a strong demand from implementers and/or users yet. If that would come surprisingly later we could add that ordering then. Maybe we could alternatively offer special extension points for that.

glyn:
> I cannot see a strong demand from implementers and/or users yet. If that would come surprisingly later we could add that ordering then.

I think that would be a backwards incompatible change, so it would be difficult to introduce. It would also impact interoperation.

> Maybe we could alternatively offer special extension points for that.

If these ordering features could be handled by extension point, that would be great!

goessner:
hmm ... my concept of ordering of different objects won't work ... exactly because of the argument: "Objects themself are not ordered".

glyn:
A partial ordering of objects could be defined in terms of the subset relationship when objects are considered to be sets of members. But, again, this is fairly arbitrary from an application POV.

goessner:
> Checking existing implementations, I've only found one ([Goessner](https://jsonpath.herokuapp.com/), and he's against this feature!) 

 :-)

In fact testing in JavaScript (node.js) surprisingly shows somehow lexicographic ordering of arrays. But this is not (sufficiently) documented at least at [MSN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Less_than).

goessner:
> I think that would be a backwards incompatible change, so it would be difficult to introduce. It would also impact interoperation.

I only was thinking: "later" this year ...

cabo:
So let me rewrite this:

> As in (add axioms relating < and ==, same as in strings)
> 
> * [] == []
> * [] < [x, ...y]
> * [x | y] ⊜ [x | z]  ≝  y ⊜ z
> * [x | y] ⊜ [z | w]  ≝  x ⊜ z if x ≠ z

[ A | B ] splits an array into first (A) and rest (B), so [1, 2, 3] becomes A = 1 and B = [2, 3]

This really is best thought of using the spaceship operator (<=>) that returns -1, 0, or 1 for a comparison (less, equal, greater); we'd need to augment this with fail (absent).

a == b ≝ (a <=> b) == 0
a != b ≝ (a <=> b) != 0
a <= b ≝ (a <=> b) <= 0
a >= b ≝ (a <=> b) >= 0
a < b ≝ (a <=> b) < 0
a > b ≝ (a <=> b) > 0

(Sorry about **not** reading `<=>` as ⇔ -- just as we don't read `<=` as ⇐ in the usual programming languages, which misreading actually caused some languages to write `<=` as `=<`.)

You can derive tautologies from the axioms like

[a] ⊜ [b] ≣ a ⊜ b

without which the principle of least surprise is heavily violated.

(This also speaks against actively searching out for non-comparables before applying these axioms, but this is a separate consideration.)

glyn:
It was that kind of example which motivated the approach in this PR: to force `<` and `>` to yield false if there is an object, boolean, or `null` among the descendants of either operand. This also helped l to preserve transitivity (e.g. `a < b` and `b < c` implies `a < c`).

Issue #252, opened 2022-08-25
Include input value in descendant selector


glyn:
The current draft applies descendant selectors to the (strict) descendants of the input value. However, as mentioned in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/201#issuecomment-1225065352, it would better reflect current implementations if descendant selectors applied to the input value as well as its descendants.

The comparison project has a [test](https://cburgmer.github.io/json-path-comparison/results/dot_notation_after_recursive_descent.html) for this which shows a consensus among existing implementations.

glyn:
See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/201#issuecomment-1228825045

glyn:
@cabo I think it would still be legitimate to call the descendant selectors by that name since the values they select are all descendants of the input value. (The one exception would be if we were to allow `..` on its own, in which case the input value would be in the resultant nodelist.)

glyn:
At the 2022-9-27 interim meeting there was a consensus for implementing this issue.

PR #253, opened 2022-08-27
Descendant selectors act on input value and its descendants

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/252

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1400,7 +1400,7 @@ Note that `..` on its own is not a valid selector.
 
 A descendant selector selects zero or more descendants of a node.
 
-A nodelist enumerating the descendants is known as a _descendant nodelist_ when:
+A nodelist enumerating the node and its descendants is known as a _reflexive descendant nodelist_ when:
 
 * nodes of any array appear in array order,
 * nodes appear immediately before all their descendants.
@@ -1411,7 +1411,7 @@ JSON objects are unordered.
 
 The resultant nodelist of a descendant selector is the result of applying a selector
 (or no selector), depending on the variant of the descendant selector, to a
-descendant nodelist, as shown below:
+reflexive descendant nodelist, as shown below:
 
 | Variant | Selector to apply | Comment |
 | :---: | :---: | ------- |
@@ -1443,6 +1443,7 @@ Queries:
 | `$..[0]` | `5` <br> `{"j": 4}` | `$['a'][0]` <br> `$['a'][2][0]` | Array values       |
 | `$..[0]` | `{"j": 4}` <br> `5` | `$['a'][2][0]` <br> `$['a'][0]` | Alternative result |
 | `$..[*]` <br> `$..*` | `{"j": 1, "k" : 2}` <br> `[5, 3, [{"j": 4}]]` <br> `1` <br> `2` <br> `5` <br> `3` <br> `[{"j": 4}]` <br> `{"j": 4}` <br> `4` | `$['o']` <br> `$['a']` <br> `$['o']['j']` <br> `$['o']['k']` <br> `$['a'][0]` <br> `$['a'][1]` <br> `$['a'][2]` <br> `$['a'][2][0]` <br> `$['a'][2][0]['j']` | All values    |
+| `$..o`   | `{"j": 1, "k": 2}` | `$['o']` | Reflexive descendant nodelist includes input value |
 {: title="Descendant selector examples"}
 
 Note: The ordering of the results for the `$..[*]` and `$..*` examples above is not guaranteed, except that:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1400,7 +1400,9 @@ Note that `..` on its own is not a valid selector.
 
 A descendant selector selects zero or more descendants of a node.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1400,7 +1400,7 @@ Note that `..` on its own is not a valid selector.
 
 A descendant selector selects zero or more descendants of a node.
 
-A nodelist enumerating the descendants is known as a _descendant nodelist_ when:
+A nodelist enumerating the node and its descendants is known as a _reflexive descendant nodelist_ when:
```
gregsdennis, line 1403: Recursive, instead of reflexive?

glyn: No - I deliberately borrowed the mathematical term "reflexive" because the nodelist includes the node itself.

The definition of _descendant_ is already recursive:

> Descendants (of a node): The children of the node, together with the children of its children, and so forth recursively. More formally, the descendants relation between nodes is the transitive closure of the children relation.

What we have here is the node and its descendants. This is the reflexive closure of the descendants relation. Or, putting it another way, it's the reflexive transitive closure of the children relation. Of course, there's the extra condition about the various possible orderings of the nodelist.

Is there a better phrase?

gregsdennis: I still don't think "reflexive" is quite right.  It seems like a stretched attempt to come up with a single phrase.

Maybe we just update the "descendants" definition to include the node.  Is there any time we use it to mean strictly children?

glyn: The binary relation I had in mind is "is a descendant of" which is a relation on JSON values and is the transitive closure of the "is a child of" relation. I considered redefining "descendants" to include the value itself, but that messes up the current descendants selectors because they only select strict descendants of the input value (NB. we don't currently support `..` on its own).

I have pushed a commit which avoids introducing a special term. See if that helps.

glyn:
CI failed with some kind of certificate error which reproduces if I try to generate the rendered version locally. I'm hoping this will sort itself out in due course as it's not clear that there is anything on the CI server or my desktop to blame.

draft-ietf-jsonpath-base.md
goessner: > Maybe we just update the "descendants" definition to include the node. Is there any time we use it to mean strictly children?

Yes, we decided in the past that "descendants" are children and children of children, not the node itself.

danielaparker: > No - I deliberately borrowed the mathematical term "reflexive" because the nodelist includes the node itself.
> 
> The definition of _descendant_ is already recursive:
> 
> > Descendants (of a node): The children of the node, together with the children of its children, and so forth recursively. More formally, the descendants relation between nodes is the transitive closure of the children relation.
> 
> What we have here is the node and its descendants. This is the reflexive closure of the descendants relation. Or, putting it another way, it's the reflexive transitive closure of the children relation. Of course, there's the extra condition about the various possible orderings of the nodelist.
> 
> Is there a better phrase?

XPath 3.1 uses [descendant-or-self](https://www.w3.org/TR/xpath-31/#unabbrev).

JSONAta uses [a few extra words](https://docs.jsonata.org/path-operators#-descendants) to explain

If the draft is targeting an audience of users, "reflexive" may not be the best choice.

glyn: @danielaparker Please see the updated PR which avoids "reflexive".

glyn:
> CI failed with some kind of certificate error which reproduces if I try to generate the rendered version locally. I'm hoping this will sort itself out in due course as it's not clear that there is anything on the CI server or my desktop to blame.
> 
> Raised [cabo/kramdown-rfc#179](https://github.com/cabo/kramdown-rfc/issues/179) to track the problem.

This healed itself up. :-)

draft-ietf-jsonpath-base.md
```diff
@@ -1400,25 +1400,24 @@ Note that `..` on its own is not a valid selector.
...
-The resultant nodelist of a descendant selector is the result of applying a selector
-(or no selector), depending on the variant of the descendant selector, to a
-descendant nodelist, as shown below:
+The selector to apply is shown below:
 
 | Variant | Selector to apply | Comment |
 | :---: | :---: | ------- |
 | `..<name>` | `.<name>`| [Dot selector](#dot-selector)  |
-| `..*` | _none_ | All descendants |
+| `..*` | `.*` | All descendants |
```
gregsdennis, line 1418: This is an interesting change and brings up and interesting point.

`.*` doesn't select the current node, only immediate children.  Now I'm questioning whether `..*` should select the current node.  I'll need to think about this some.

glyn: The comparison project shows a consensus for `..*` _not_ including the current node. See [this test](https://cburgmer.github.io/json-path-comparison/results/dot_notation_with_wildcard_after_recursive_descent.html).

gregsdennis: I think only including children is the right way.  I can't think of an example where it'd be necessary to include the current node, and it has similarity with `.*` this way.

glyn:
Closing since superseded by https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/266 which is effectively the same change applied to a base which has moved on.

Issue #255, opened 2022-08-30
Allow wildcard in list selector


glyn:
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/201#issuecomment-1231469606 suggested that, in current draft terminology, a list selector should include wildcards. Even though this isn't particularly useful, it is in the interest of both ease of implementation and to keep the specification more concise.

gregsdennis:
I'm open to this, but can we hold this until #201 is resolved, please? I'm really trying to ensure there are no functional changes with it, and merging functional changes while I'm writing a PR are hard to track.  I fear something will be missed.

glyn:
I think we need to agree the process in a bit more detail.

I thought we were identifying technical changes in #210, raising those as separate issues/PRs, and merging them _before_ embarking on a purely editorial PR to cover the rest of #201.

Otherwise, a PR covering #201 and avoiding technical changes won't get the benefit of the changes we are identifying.

Precisely who creates the editorial PR is to be decided. I'd be happy to work on that once we've settled the outstanding questions. If I work on it, I will try to construct it as a series of incremental commits to make it easy to review.

With that in mind, @gregsdennis's branch can be considered a prototype which we can use to help us agree the direction of #201.

Is that process acceptable?

Assuming it is, there is then the question of whether #255 (and this PR) is a technical change we want to adopt. @cabo proposed it and I agreed it was an improvement. I'd be interested in what others have to say.

gregsdennis:
> I thought we were identifying technical changes in #210, raising those as separate issues.

Yes.

> /PRs, and merging them before embarking on a purely editorial PR to cover the rest of #201.

No.  If like us to stay focused on the editorial change for now.  We can keep track of functional changes in issues like this one, but trying to track functional changes that are happening while attempting to make purely editorial change will inevitably lead to a mistake.  This can be avoided by focusing on one thing at a time.

> Precisely who creates the editorial PR is to be decided.

This is my proposal, and I'd like to do it to make sure it's in line with what I have in mind.  Additionally, I've already done the legwork to identify a target.  All that remains is massaging the commits.

> Assuming it is, there is then the question of whether #255 (and this PR) is a technical change we want to adopt. @cabo proposed it and I agreed it was an improvement. I'd be interested in what others have to say.

I have no disagreement that this is something that we want to change.  I just prefer not to change it until the editorial is complete.

glyn:
[...]

Issue #201, opened 2022-06-22
Syntax breakdown model for selectors needs to be reworked


gregsdennis:
I think we've cornered ourselves with the way we've defined the selectors, specifically around the array-format selectors.

Implementing a parser for JSON Path, it feels more correct to say that the bracket syntax houses (potentially) multiple selectors.

In [Section 3.5](https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base#section-3.5), The text gives examples of thes selectors by including the brackets:

```
   *  Index selector [<index>], where <index> is either a (possibly negative, see Section "Semantics") array index or an object member name.
   *  Index wildcard selector [*].
   *  Array slice selector [<start>:<end>:<step>], where the optional values <start>, <end>, and <step> are integer literals.
   *  Descendants selector ...
   *  List selector [<sel1>,<sel2>,...,<selN>], holding a comma separated list of index and slice selectors.
   *  Filter selector [?(<expr>)]
```

If instead we describe each select without the brackets, and describe the brackets as merely a housing for one or more selectors, then we can actually remove the "list" selector as it becomes defined implicitly.  I would change the above to

```
   *  Name selector '<name>', where <name> is a string key of an object.
   *  Index selector <index>, where <index> is either a (possibly negative, see Section "Semantics") array index.
   *  Index wildcard selector *.
   *  Array slice selector <start>:<end>:<step>, where the optional values <start>, <end>, and <step> are integer literals.
   *  Descendants selector ...
   *  Filter selector ?(<expr>)
```

along with some text stating that each of these must be contained within brackets and may be combined by comma-delimiting them within the brackets.

***NOTE**  In the above, I've removed List Selector and added Name Selector (split from Index).*

From an implementor's point of view, this makes a lot more sense.

cabo:
On 2022-06-22, at 10:04, Greg Dennis ***@***.***> wrote:
> 
> From an implementor's point of view, this makes a lot more sense.

Yes. The current description tries to provide a useful exposition and is not optimized for 1:1 translation into an implementation of the parser.  

This may be a bit unsatisfactory, but the syntax is what it is.

In any case, we need to make sure that we only call actual selectors selectors, not any syntactic component of selectors.

Grüße, Carsten

cabo:
Hi Greg,

> On 2022-06-22, at 11:29, Greg Dennis ***@***.***> wrote:
> 
> 
> > the syntax is what it is - @cabo
[...]

PR #256, opened 2022-08-30
Allow wildcard in list selector

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/255

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1325,7 +1325,7 @@ renamed into "list selector".
 {: unnumbered}
 
 The list selector is syntactically related to the
-`dot-selector`, `index-selector`, `slice-selector`, and the `filter-selector`.
+`dot-selector`, `index-selector`, `index-wild-selector`, `slice-selector`, and the `filter-selector`.
 It contains two or more entries, separated by commas.
 
 ~~~~ abnf
@@ -1333,6 +1333,7 @@ list-selector  = "[" S list-entry 1*(S "," S list-entry) S "]"
 
 list-entry     =  ( quoted-member-name /
                     element-index      /
+                    wildcard /
                     slice-index /
                     filter
                   )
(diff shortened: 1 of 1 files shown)


gregsdennis:
I'm open to this, but can we hold this until #201 is resolved, please? I'm really trying to ensure there are no functional changes with it, and merging functional changes while I'm writing a PR are hard to track.  I fear something will be missed.

glyn:
> I'm open to this, but can we hold this until #201 is resolved, please? I'm really trying to ensure there are no functional changes with it, and merging functional changes while I'm writing a PR are hard to track. I fear something will be missed.

Ok, I think we need to agree a strategy. I thought that we would pick out any functional changes arising in #201 so that all that remains in a PR for #201 would be editorial changes.

If we agree this strategy, then we need to decide whether #255 (implemented in this PR) is a functional change arising in #201 that we want to adopt. @cabo suggested it. I agree with it. What do others think?

gregsdennis:
> I thought that we would pick out any functional changes arising in #201 so that all that remains in a PR for #201 would be editorial changes.

Yes, this is the strategy.

> we need to decide whether #255 (implemented in this PR) is a functional change...

Yes, this is a functional change.  Currently, the document disallows `*` inside what is currently the "list selector."  The change is to allow this.

> ... arising in #201 that we want to adopt.

Perhaps arising from, but not included in #201.  #201 is intended to only contain editorial, non-functional changes.  This change was suggested by #201, but it was a mistake.  I believe it has been reverted from my draft edits so that current functionality is maintained.

glyn:
> This change was suggested by #201, but it was a mistake.

Ok, let's defer merging this for now. We can discuss it separately from #201. Also, the diff is small, so we can easily apply it after #201 if we decide to. I'll turn it into a draft PR for now to avoid it being merged prematurely.

(BTW github seems to be playing up. My comment https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/256#issuecomment-1232524305 took a while to appear and I thought it had gone AWOL, so I posted a similar comment with a more expanded description of the process which I presume will become visible in a few hours. Ho hum...)

glyn:
This was implemented in PR #258.

Issue #249, opened 2022-08-18
Improve specification of string ordered comparisons


glyn:
The current spec defines string comparison using one of the operators `<`, `<=`, `>`, or `>=` in terms of when one string is _less than_ another. That's fine for `<`, but:

* for `<=` it effectively assumes [issue 245](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/245) is fixed (i.e. that `a <= b` for strings `a` and `b` if `a < b` or `a == b`)
* for `>` and `>=` it leaves quite a lot to the reader's imagination.

Once we've settled on [issue 245](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/245), the above should be addressed.

~We should also adopt a similar approach for ordered comparisons (<, >) of arrays if [issue 244](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/244) is adopted.~ (It wasn't.)

(I labelled this issue as `editorial` because I don't believe it will change the intended semantics in either case.)

glyn:
I propose to let https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/251 land before addressing this issue.

PR #257, opened 2022-08-30
Define > in terms of <

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/249

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1117,14 +1117,14 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
 ##### Comparisons
 {: unnumbered}
 
-The comparison operators `==`, `<`, and `>` are defined first and then these are used to define `!=`, `<=`, and `>=`.
+The comparison operators `==` and `<` are defined first and then these are used to define `!=`, `<=`, `>`, and `>=`.
 
 When a path resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
 is between two paths each of which result in an empty nodelist.
 
-* a comparison using either of the operators `<` or `>` yields false.
+* a comparison using the operator `<` yields false.
 
 When any path on either side of a comparison results in a nodelist consisting of a single node, each such path is
 replaced by the value of its node and then:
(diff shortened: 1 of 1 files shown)


timbray (commented):
I personally prefer the previous version, with standalone definitions, but think this is a matter of editorial discretion.

gregsdennis (commented):
I agree with Tim. It's fine and correct to define `<=` in terms of other operators, but I feel `<` and `>` should both be equally defined.

Defining one in terms of the other means you have to make an arbitrary choice of which is directly defined and which is indirectly defined.

glyn:
Thanks for your comments @timbray and @gregsdennis. I feel I should explain. The current draft defines the orderings `<` and `>` using the same text. For strings, the text talks about "less than" but not "greater than". So the current text implicitly assumes that for strings `a` and `b`, `a > b` if and only if `b < a`. Non-mathematicians would tend to say this is patently obvious. But this PR avoids that assumption by only defining `<` using the same text and then defining `>` in terms of `<`.

However, I take the point about the standalone definitions and the arbitrariness of defining one in terms of the other.

So my question is: is the current draft text sufficiently precise? If not, I'll rework this PR to make it more precise. (When I started out doing that, it began to get quite wordy, so I changed tack.)

goessner:
Despite the fact, that Glyn's approach is mathematically clean, I would favour the symmetric approach by defining `==`, `>`,`<`.

I am quite sure, that nobody will cry "redundancy" then.

cabo (approved):
This is exactly the right way to do this.

cabo:
> Defining one in terms of the other means you have to make an arbitrary choice of which is directly defined and which is indirectly defined.

Yes.  Is that a problem?  `<` comes before `>`...

glyn:
Actually, one the symmetry point, we were comfortable defining `!=` in terms of `==` and that was asymmetric.

gregsdennis:
> `<` comes before `>`

In what system of mathematics are operations ordered?  This is completely arbitrary.

I'm not convinced this needs to change.  I prefer defining both `<` and `>`.

And for symmetry, I'd change strings to match.

glyn:
Applying editorial discretion here. Merging.

PR #258, opened 2022-09-05
Update to syntax breakdown

gregsdennis:
Resolves #201 
Resolves #255 

Reorganizes and simplifies the syntax declarations.

Initial discussions yielded [this document](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/compare/main...gregsdennis:draft-ietf-jsonpath-base:different-syntax-breakdown?expand=1).  This served as a target.  I've added the target to the repo as a reference for myself as I reworked it with nicer commits.  It will be removed just before merge as a last step of this PR.

I'm sure that there are some build issues.  I probably have links that go nowhere.  I _think_ I got all of the ABNF reworked correctly.

Also, I've used `//PICKER//` and `//PICKERS//` as placeholders for us to find a name for these things that appear inside the `[]`.  Let's do that now.

PR #259, opened 2022-09-15
Unify selector syntax

glyn:
This draft PR attempts to extract the syntactic advantages from PR https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/258 while preserving the semantic structure. It's only a start, but it gives an idea of a purely syntactic approach to issue #201. The semantic structure and document structure are unchanged.

To do:

- [ ] Shorthand syntax as in PR https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/258

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/201

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -565,8 +565,8 @@ A JSONPath query consists of a sequence of selectors. Valid selectors are
   * Index wildcard selector `[*]`
   * Array slice selector `[<start>:<end>:<step>]`, where the optional
     values `<start>`, `<end>`, and `<step>` are integer literals
-  * Child selector `[<matcher1>,...,<matcherN>]`, holding a comma
-    separated list of index, slice, and filter matchers
+  * Child selector `[<entry1>,...,<entryN>]`, holding a comma
+    separated list of index, slice, and filter entries
   * Filter selector `[?(<expr>)]`
   * Current item selector `@` (used in expressions)
   * Descendants selectors starting with a double dot `..`
@@ -709,9 +709,11 @@ Queries:
 
 An index selector `[<index>]` selects at most one object member value or at most one array element value.
 
+An index selector is one variant of a child selector ({{child-selector}}).
+
 ~~~~ abnf
-index-selector      = "[" S index-matcher S "]"
-index-matcher       = quoted-member-name / element-index
+index-selector      = "[" S name-or-index S "]"
+name-or-index       = quoted-member-name / element-index
 ~~~~
 
 Applying the `index-selector` to an object value in its input nodelist, a
@@ -893,9 +895,10 @@ The array slice selector has the form `[<start>:<end>:<step>]`.
 It selects elements starting at index `<start>`, ending at — but
 not including — `<end>`, while incrementing by `step`.
 
+An array slice selector is one variant of a child selector ({{child-selector}}).
+
 ~~~~ abnf
-slice-selector = "[" S slice-matcher S "]"
-slice-matcher = slice-index
+slice-selector = "[" S slice-index S "]"
 
 slice-index    =  [start S] ":" S [end S] [":" [S step ]]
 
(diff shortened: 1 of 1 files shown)


gregsdennis:
> while minimising the semantic disruption

There should be no semantic disruptionin #258 as is a non-functional change.

draft-ietf-jsonpath-base.md
```diff
@@ -710,10 +707,11 @@ Queries:
 #### Syntax {#syntax-index}
 {: unnumbered}
 
-An index selector `[<index>]` addresses at most one object member value or at most one array element value.
+An index selector `[<index>]` selects at most one object member value or at most one array element value.
 
 ~~~~ abnf
-index-selector      = "[" S (quoted-member-name / element-index) S "]"
+index-selector      = "[" S index-matcher S "]"
```
gregsdennis, line 713: The point of this approach is that we don't need `___-selector`s anymore.  There's no reason to define it since 1-length lists aren't used.

glyn: I agree _syntactically_. Note that `index-selector` is deliberately unreferenced in the ABNF. It's meant to be a specific case of a child selector.

The reason for defining the index, slice, and filter selectors is to break the semantics of the child selector into manageable chunks which can then be combined in the semantics section of the child selector while avoiding defining the semantics of matchers except in the context of a selector.

glyn:
> > while minimising the semantic disruption
> 
> There should be no semantic disruption in #258 as is a non-functional change.

I agree that #258 aims to preserve the current JSONPath semantics. The disruption is, at least from my perspective, in the way the semantics is defined in #258 with the splits between child selector and various *pickers.

draft-ietf-jsonpath-base.md
gregsdennis: The problem is that you're still calling it an "index selector" when there is no such thing.  It's merely a child selector that happens to have a single index *picker.

Separately, we should keep using the placeholder until we have _all_ agreed on something else.

goessner:
We agreed, that Greg should elaborate his approach of strict distinction between *pickers and selectors, which he did with #258.

Now we need to discuss the result, which might be the main topic of the next interim meeting. Some aspects:

* I doubt, that we can say: "One approach is more correct than the other". They are only (fundamentally) different.
* We urgently need to term that *pickers beast. I already made a proposal here.
* I would prefer to say '*pickers match nodes' over '*pickers match values'.
* I am not so happy with calling *pickers 'matchers'.
* Glyn made two proposal to define, how *pickers and selectors relate and interoperate. They go into the right direction and to get them straight and make one out of them is quite important.

-- Stefan

draft-ietf-jsonpath-base.md
glyn: > The problem is that you're still calling it an "index selector" when there is no such thing. It's merely a child selector that happens to have a single index *picker.

I have clarified that "index selector" is simply a variant of a "child selector". "index selector" is defined so it can be given semantics and then the semantics of the child selector can be constructed from the semantics of its syntactic variants. That's the gist of this approach: to gain the syntactic unification of PR #258 while preserving the current semantic structure.

> 
> Separately, we should keep using the placeholder until we have _all_ agreed on something else.

Fair enough. Actually I think, in this PR, there is no need for a special term so I've avoided it.

cabo: I don't think we should create special-case selectors.  Instead, *pickers should have defined semantics that can interact with the defined semantics of the two selectors.
(Shortcuts are just a different way to write the selectors and don't have their own semantics.)

glyn:
The purpose of this draft PR was to test a possible consensus around a purely syntactic unification of child selectors. It seems that most people (except me) would like to continue with the approach of PR #258 and define *picker semantics. I'll therefore close this PR and look to PR #258 to gain (rough - possibly excluding me) consensus.

PR #262, opened 2022-10-05
Update to syntax breakdown - PR for comments only

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -246,7 +246,11 @@ Unicode Scalar Value:
   E000 to 10FFFF. JSON values of type string are sequences of Unicode scalar values.
 
 Singular Path:
-: A JSONPath expression built from selectors which each select at most one node.
+: A JSONPath expression built from appenders which each select at most one node.
```
glyn, line 249: This sentence demonstrates a weakness in the "appender"/"selector" terminology. In plain English, it is more natural to talk about selectors selecting and appenders appending.

draft-ietf-jsonpath-base.md
```diff
@@ -246,7 +246,11 @@ Unicode Scalar Value:
...
+
+Selector:
+: A single item within a bracketed (`[]`) child appender that matches the values which
+  are to be selected.
```
glyn, line 253: * This definition positions a selector as something which _matches_, leaving _selecting_ to the child appender. That division is not consistent with the rest of the document.

* Selectors also appear with descendant appenders.

* "the values" is a bit vague. I think we need to say that a selector selects zero or more children of a given value.

draft-ietf-jsonpath-base.md
```diff
@@ -289,23 +293,25 @@ JSONPath expressions are applied to a JSON value, the *argument*.
...
-or the more general *bracket notation*
+but may also use the shorthand *dot notation*
 
 ~~~~
-$['store']['book'][0]['title']
+$.store.book[0].title
 ~~~~
 
 to build paths that are input to a JSONPath implementation.
+Generally, the dot notation is preferred for most applications.
```
glyn, line 309: I'm not sure why we are expressing such a preference. It would be helpful to explain why this is the case.

draft-ietf-jsonpath-base.md
```diff
@@ -289,23 +293,25 @@ JSONPath expressions are applied to a JSON value, the *argument*.
...
+Mixing these syntaxes within a single path is also useful.
 
 JSONPath allows the wildcard symbol `*` to select any member of an
 object or any element of an array ({{wildcard}}).
-The descendant operators (which start with `..`) select some or all of the descendants ({{descendant-selectors}}) of a node.
+The descendant appender (which starts with `..`) recursively selects some or all of the descendants ({{descendant-appender}}) of a node.
```
glyn, line 314: It is sufficient to talk about selecting descendants: "recursively" adds no value.

draft-ietf-jsonpath-base.md
```diff
@@ -320,19 +326,21 @@ $.store.book[?(@.price < 10)].title
...
+| `..[<selectors>]`   | [descendant child appender](#descendant-appender): recursive version of the child appender                             |
+| `*`                 | [wildcard selector](#name-selector): selects all children of an array or object                                         |
+| `'name'`            | [name selector](#name-selector): selects a named child of an object                                                     |
+| `3`                 | [index selector](#index-selector): selects an indexed child of an array (from 0)                                        |
+| `0:100:5`           | [array slice selector](#slice): start:end:step for arrays                                                               |
+| `?<expr>`           | [filter selector](#filter-selector): selects particular children using a boolean expression                             |
+| `@`                 | [current node appender](#filter-selector) (valid only within filter selectors)                                          |
```
glyn, line 339: `@` is a *starter. For consistency with `$`, it would be better to describe it as a "current node identifier".

draft-ietf-jsonpath-base.md
```diff
@@ -320,19 +326,21 @@ $.store.book[?(@.price < 10)].title
...
-| `?...`      | [filter selector](#filter-selector)                                                                                                 |
-| `()`        | expression: within [filter selectors](#filter-selector)                                                                            |
+| JSONPath            | Description                                                                                                             |
+|---------------------|-------------------------------------------------------------------------------------------------------------------------|
+| `$`                 | [root node identifier](#root-identifier)                                                                                |
+| `[<selectors>]`     | [child appender](#child-appender) selects zero or more children of JSON objects and arrays; contains one or more selectors, separated by commas        |
+| `..[<selectors>]`   | [descendant child appender](#descendant-appender): recursive version of the child appender                             |
```
glyn, line 333: "descendant appender" is sufficient - no  need to mention "child".

A description more consistent with that of the child appender would be: "selects zero or more descendants of JSON objects and arrays; contains one or more selectors, separated by commas"

draft-ietf-jsonpath-base.md
```diff
@@ -320,19 +326,21 @@ $.store.book[?(@.price < 10)].title
...
+| `*`                 | [wildcard selector](#name-selector): selects all children of an array or object                                         |
```
glyn, line 334: Does it select or match?

draft-ietf-jsonpath-base.md
```diff
@@ -448,37 +456,30 @@ behavior for JSONPath queries in these situations.
...
-{{<filter-selector}}, and {{<descendant-selectors}} describe behavior that
+Specifically, the "Semantics" subsections of Sections
+{{<name-selector}}, {{<wildcard}},
+{{<filter-selector}}, and {{<descendant-appender}} describe behavior that
 turns unpredictable when the JSON value for one of the objects
 under consideration was constructed out of JSON text that exhibits
 multiple members for a single object that share the same member name
 ("duplicate names", see {{Section 4 of -json}}).
-Also, matching ({{selectors}}) and comparing
+Also, matching ({{appenders}}) and comparing
```
glyn, line 466: Isn't matching what selectors do, rather than something appenders do?

draft-ietf-jsonpath-base.md
```diff
@@ -448,37 +456,30 @@ behavior for JSONPath queries in these situations.
...
 ({{comparisons}} in Section {{<filter-selector}}) of strings assumes these
 strings are sequences of Unicode scalar values, turning unpredictable
 if they aren't ({{Section 8.2 of -json}}).
 
 ## Syntax
 
-Syntactically, a JSONPath query consists of a root selector (`$`), which
+Syntactically, a JSONPath query consists of a root starter (`$`), which
```
glyn, line 473: This is called a "root identifier" elsewhere.

draft-ietf-jsonpath-base.md
```diff
@@ -448,37 +456,30 @@ behavior for JSONPath queries in these situations.
...
-json-path = root-selector *(S (dot-selector        /
-                               dot-wild-selector   /
-                               index-selector      /
-                               index-wild-selector /
-                               slice-selector      /
-                               filter-selector     /
-                               list-selector       /
-                               descendant-selector))
+json-path = root-identifier *(S (child-appender               /
+                                 descendant-child-appender))
```
glyn, line 479: `descendant-appender` is sufficient IMO.

draft-ietf-jsonpath-base.md
```diff
@@ -489,35 +490,35 @@ implementation.
 The semantics are that a valid query is executed against a value,
 the *argument*, and produces a list of zero or more nodes of the value.
 
-The query is a sequence of zero or more *selectors*, each of
-which is applied to the result of the previous selector and provides
-input to the next selector.
+The query is a sequence of zero or more *appenders*, each of
```
glyn, line 493: Queries start with a root identifier rather than an appender.

draft-ietf-jsonpath-base.md
```diff
@@ -489,35 +490,35 @@ implementation.
...
+which is applied to the result of the previous appender and provides
+input to the next appender.
 These results and inputs take the form of a *nodelist*, i.e., a
 sequence of zero or more nodes.
 
-The nodelist presented to the first selector contains a single node,
+The nodelist presented to the first appender contains a single node,
```
glyn, line 499: This needs to talk about the root identifier rather than the first appender.

draft-ietf-jsonpath-base.md
```diff
@@ -489,35 +490,35 @@ implementation.
...
+The nodelist resulting from the last appender is presented as the
 result of the query; depending on the specific API, it might be
 presented as an array of the JSON values at the nodes, an array of
 Normalized Paths referencing the nodes, or both — or some other
 representation as desired by the implementation.
 Note that the API must be capable of presenting an empty nodelist as
 the result of the query.
 
-A selector performs its function on each of the nodes in its input
+A appender performs its function on each of the nodes in its input
```
glyn, line 509: An appender...

draft-ietf-jsonpath-base.md
```diff
@@ -489,35 +490,35 @@ implementation.
...
 nodelist, during such a function execution, such a node is referred to
 as the "current node".  Each of these function executions produces a
 nodelist, which are then concatenated to produce
-the result of the selector. A node may be selected more than once and
+the result of the appender. A node may be selected more than once and
 appear that number of times in the nodelist. Duplicate nodes are not removed.
 
-The processing within a selector may execute nested queries,
+The processing within a appender may execute nested queries,
```
glyn, line 516: an appender

draft-ietf-jsonpath-base.md
```diff
@@ -527,7 +528,7 @@ Consider this example. With the argument `{"a":[{"b":0},{"b":1},{"c":2}]}`, the
 query `$.a[*].b` selects the following list of nodes: `0`, `1`
 (denoted here by their value).
 
-The query consists of `$` followed by three selectors: `.a`, `[*]`, and `.b`.
+The query consists of `$` followed by three appenders: `.a`, `[*]`, and `.b`.
 
 Firstly, `$` selects the root node which is the argument.
```
glyn, line 533: Talking of `$` as if it was a selector doesn't seem consistent with "root identifier" language.

draft-ietf-jsonpath-base.md
```diff
@@ -549,58 +550,34 @@ The result is a list containing `0`, `1`.
 This is the concatenation of three lists, two of length one containing
 `0`, `1`, respectively, and one of length zero.
 
-As a consequence of this approach, if any of the selectors selects no nodes,
+<!--
+Greg: this could be reworded since appenders don't select.  Maybe they do
+vicariously through selectors.
```
glyn, line 555: Agreed. The PR needs to use consistent terms throughout.

draft-ietf-jsonpath-base.md
```diff
@@ -612,118 +589,36 @@ Queries:
...
-
-~~~~ abnf
-dot-wild-selector    = "." wildcard       ;  dot followed by asterisk
-wildcard             = "*"
-~~~~
+## Selectors
 
-#### Semantics
-{: unnumbered}
+Selectors are defined solely within the context of the [child appender](#child-appender).
```
glyn, line 596: * I find the notion of selectors being _defined_ in the context of the child appender confusing. They need to be defined here and then referenced by the child appender.

* Selectors are used in descendant appenders too.

draft-ietf-jsonpath-base.md
```diff
@@ -612,118 +589,36 @@ Queries:
...
+A selector outside of a child appender is invalid.
```
glyn, line 597: This statement seems unnecessary. Why would anyone dream of constructing a query with a selector outside of a child (or descendant) appender? The ABNF does not allow it.

draft-ietf-jsonpath-base.md
```diff
@@ -612,118 +589,36 @@ Queries:
...
 
-A `dot-wild-selector` acts as a wildcard by selecting the nodes of
-all member values of an object in its input nodelist as well as all
-element nodes of an array in its input nodelist.
+A selector identifies, or matches, child elements to be selected for the
+resulting nodelist.
```
glyn, line 600: This is confusing. We should reserve identification language for `$` (and perhaps `@`). I think the rest of the PR suggests that selectors match rather than select and it's the child/descendant appender that actually does the selecting.

draft-ietf-jsonpath-base.md
```diff
@@ -806,18 +687,17 @@ in the table below:
 | \\uXXXX            | U+XXXX              | unicode character           |
 {: title="Escape Sequence Replacements" cols="c c"}
 
-The `index-selector` applied with a `quoted-member-name` to an object
-selects the node of the corresponding member value from it, if and only if that object has a member with that name.
+The name selector applied to an object
+matches the node of the corresponding member value from it, if and only if that object has a member with that name.
 Nothing is selected from a value that is not a object.
```
glyn, line 692: "Nothing is matched..."?

draft-ietf-jsonpath-base.md
```diff
@@ -839,34 +718,30 @@ Queries:
...
 {: unnumbered}
 
-An `index-wild-selector`
-selects the nodes of all member values of an object as well as of all elements of an
-array.
+A `wildcard` selects the nodes of all children of an object or array.
 
-Applying the `index-wild-selector` to a primitive JSON value (that is,
+Applying the `wildcard` to a primitive JSON value (that is,
 a number, a string, `true`, `false`, or `null`) selects no node.
```
glyn, line 743: "matches no node"

draft-ietf-jsonpath-base.md
```diff
@@ -839,34 +718,30 @@ Queries:
...
+wildcard             = "*"
 ~~~~
 
 #### Semantics
 {: unnumbered}
 
-An `index-wild-selector`
-selects the nodes of all member values of an object as well as of all elements of an
-array.
+A `wildcard` selects the nodes of all children of an object or array.
```
glyn, line 740: "matches"

draft-ietf-jsonpath-base.md
```diff
@@ -881,24 +756,78 @@ Queries:
...
+element. JSONPath allows it to be negative (see {{index-semantics}}).
+
+Notes:
+1. An `index-selector` is an integer (in base 10, as in JSON numbers).
+2. As in JSON numbers, the syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.
+
+#### Semantics {#index-semantics}
+{: unnumbered}
+
+The `index-selector` applied to an array selects an array element using a zero-based index.
```
glyn, line 789: "matches"

draft-ietf-jsonpath-base.md
```diff
@@ -881,24 +756,78 @@ Queries:
...
+<!-- EDITING NOTE: There are non-breaking spaces here between j and j -->
+<!-- i.e., j j and not j j -->
+
+JSON:
+
+    {
+      "o": {"j j": {"k.k": 3}},
+      "a": ["a","b"],
+      "'": {"@": 2}
+    }
```
glyn, line 810: The JSON can be stripped back to `["a","b"]` and the examples simplified to `$[1]` and `$[-2]`.

draft-ietf-jsonpath-base.md
```diff
@@ -929,24 +858,24 @@ This section is non-normative.
 
 Array indexing is a way of selecting a particular element of an array using
 a 0-based index.
-For example, the expression `[0]` selects the first element of a non-empty array.
+For example, the expression `0` selects the first element of a non-empty array.
```
glyn, line 861: "matches" (same comment elsewhere in this section)

draft-ietf-jsonpath-base.md
```diff
@@ -1168,7 +1094,7 @@ The following table lists filter expression operators in order of precedence fro
 #### Semantics
 {: unnumbered}
 
-The `filter-selector` works with arrays and objects exclusively. Its result is a list of *zero*, *one*, *multiple* or *all* of their array elements or member values, respectively.
+The filter selector works with arrays and objects exclusively. Its result is a list of *zero*, *one*, *multiple* or *all* of their array elements or member values, respectively.
 Applied to other value types, it will select nothing.
```
glyn, line 1098: "match"

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
-However, the term "union" has the connotation of a set operation that involves
-merging input sets while avoiding duplicates, so the concept was
-renamed into "list selector".
+  * Child appender `[<selectors>]`, where `<selectors>` is one or more of
+    several selector types, which match the nodes to select,
+    separated by commas
+  * Descendant child appender `..[<selectors>]`
+
+The descendant child appender is a child appender that has been made to
+recursively select values within nested objects and arrays by prepending it with `..`.
```
glyn, line 1245: It can also select values which are children of the input value, not just those which appear within objects and arrays nested in the input value.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
 
 #### Syntax
 {: unnumbered}
 
-The list selector is syntactically related to the
-`dot-selector`, `index-selector`, `slice-selector`, and the `filter-selector`.
-It contains two or more entries, separated by commas.
+The child appender has the form `[<selectors>]` where `<selectors>` is a comma-delimited
+collection of one or more selectors.
+Each selector is defined below.
```
glyn, line 1255: This sentence is confusing given that selectors were defined above.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
+A `name-selector-shorthand` string is converted to a member name by removing
+the initial dot.
+
+#### Semantics
 {: unnumbered}
 
-JSON:
+A child appender operates on objects and arrays only.
+It contains a comma-delimited collection of selectors to indicate which
+object members and array elements to selects.
```
glyn, line 1306: "select"

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
 #### Syntax
 {: unnumbered}
 
-The descendant selectors start with a double dot `..` and can be
-followed by an object member name (similar to the `dot-selector`),
-a wildcard (similar to the `dot-wild-selector`),
-an `index-selector`, `index-wild-selector`, `filter-selector`, or `list-selector` acting on objects or arrays,
-or a `slice-selector` acting on arrays.
+The descendant appender starts with a double dot `..`
+followed by either a child appender (`descendant-child-appender`).
```
glyn, line 1325: strike "either"

glyn, line 1325:
```suggestion
and is followed by a child appender (`descendant-child-appender`).
```

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
-Note that `..` on its own is not a valid selector.
+The shorthand notations exist for the occasion when a
+descendant child appender is used with either
+
+* a single name selector where the name can be used in its shorthand notation
+* a single wildcard selector
+
+The shorthand is not valid for child appenders containing more than one selector
+other selector types, or name selectors that cannot themselves be represented
+in shorthand.
```
glyn, line 1344: The above sentence is hard to parse.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
+
+Note that `..` on its own is not valid.
```
glyn, line 1346: Better to say it isn't a valid appender.

draft-ietf-jsonpath-base.md
cabo: appender is placeholder terminology.

But to the point: Just because an oven bakes, we don't have to call it a baker.

glyn: > appender is placeholder terminology.

Hmmm. It looks concrete to me. Perhaps we should bring back some of the placeholders until we can think of better terms.

> 
> But to the point: Just because an oven bakes, we don't have to call it a baker.

Sure, but if we choose a term like baker, it's hard to avoid all baking connotations. In the current draft, we have selectors which select and that terminology is easy to understand and use. We need something similarly readable for these new terms.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1231,124 @@ Queries:
...
+This shorthand entirely replaces the bracketed syntax.
+For example, `['child']` and `.child` are equivalent as appenders,
+as are `[*]` and `.*`.
 
+A `name-selector-shorthand` starts with a dot `.` followed by an object member's name.
 
-#### Examples
+~~~~ abnf
+wild-selector-shorthand     = "." wildcard
+name-selector-shorthand    = "." dot-member-name
```
glyn, line 1280: The above productions are orphans. They need to be referenced in the rest of the ABNF.

glyn:
The comments in this PR were addressed in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/263.

draft-ietf-jsonpath-base.md
goessner: This definition could point us to a way out of the weakness in the "appender"/"selector" terminology, Glyn was mentioning.

If we re-rename "selector" to what it has been before and call "selector items" (or "items" in short) those items inside the brackets, above definition reads:

```
Item:
: A single item within a bracketed (`[]`) child selector that matches the values which
  are to be selected.
```
Use of `selector`/ `selector-item` would also relax the need of strict differentiating "selecting" / "matching" terminology (which is useful though).

draft-ietf-jsonpath-base.md
goessner: Yes, better remove "recursively".

PR #263, opened 2022-10-07
Update to syntax breakdown plus editorial improvements

glyn:
This is a revised version of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/258 which fixes the editorial issues identified in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/262.

Reviewers may find the [rendered version](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/201-syntax-improvement/draft-ietf-jsonpath-base.html) helpful.

TODO:

- [x] Agree a concrete term for __APPENDER__ and replace the placeholder throughout.
- [x] Fix the warnings/errors from `make`.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/201
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/255.


gregsdennis (dismissed):
I recognize that this extends #258, but that content still needs to be reviewed by the others on it's own.

glyn:
> I recognize that this extends #258, but that content still needs to be reviewed by the others on it's own.

Why does #258 need reviewing on its own? Wouldn't it be more efficient for them to review the current PR in which some editorial rough edges have been polished away?

draft-ietf-jsonpath-base.md
```diff
@@ -289,50 +298,52 @@ JSONPath expressions are applied to a JSON value, the *argument*.
...
 
 ~~~~
-$['store']['book'][0]['title']
+$.store.book[0].title
 ~~~~
 
 to build paths that are input to a JSONPath implementation.
+A single path may use a combination of bracket and dot notations.
+
+A wildcard `*` ({{wildcard}}) may be used to select all children of an
```
cabo, line 315: wildcards are also available for all descendants.

This should be fixed in the context of making wildcards selectors.

draft-ietf-jsonpath-base.md
```diff
@@ -289,50 +298,52 @@ JSONPath expressions are applied to a JSON value, the *argument*.
...
 
-Filter expressions are supported via the syntax `?(<boolean expr>)` as in
+Filter expressions `?<boolean expr>` select certain children of an object or array as in
 
 ~~~~
-$.store.book[?(@.price < 10)].title
+$.store.book[?@.price < 10].title
 ~~~~
 
 {{tbl-overview}} provides a quick overview of the JSONPath syntax elements.
```
cabo, line 329: I believe we should be explicit very early about some of the segments being shorthands of the canonical ones.
This allows us to focus on the canonical ones later.

draft-ietf-jsonpath-base.md
```diff
@@ -289,50 +298,52 @@ JSONPath expressions are applied to a JSON value, the *argument*.
...
-| `.*` <br> `[*]` | all child member values and array elements: [dot wildcard selector](#wildcard), [index wildcard selector](#index-wildcard-selector) |
-| `[3]`       | [index (subscript) selector](#index-selector): index current node as an array (from 0)                                              |
-| `[..,..]`   | [list selector](#list-selector): allow combining selector styles                                                                    |
-| `[0:100:5]` | [array slice selector](#slice): start:end:step                                                                                      |
-| `?...`      | [filter selector](#filter-selector)                                                                                                 |
-| `()`        | expression: within [filter selectors](#filter-selector)                                                                            |
+| JSONPath            | Description                                                                                                             |
```
cabo, line 331: This should be sorted into query starts ('root identifiers'), query steps ('segments', sorted into the canonical ones and their shorthands), and selectors.
We need to make sure these are not mixed up in the minds of the readers.

draft-ietf-jsonpath-base.md
```diff
@@ -489,35 +493,30 @@ implementation.
...
+the result of the __APPENDER__. A node may be selected more than once and
+appears that number of times in the nodelist. Duplicate nodes are not removed.
 
-The processing within a selector may execute nested queries,
-which conform to the semantics defined here.
-Typically, the argument to that query will be the current node of the
-selector or a set of nodes subordinate to that current node.
-
-A syntactically valid selector MUST NOT produce errors.
+A syntactically valid __APPENDER__ MUST NOT produce errors.
```
cabo:
```suggestion
A syntactically valid __APPENDER__ MUST NOT produce errors when executing the query.
```

... is a bit more explicit.

draft-ietf-jsonpath-base.md
```diff
@@ -549,58 +547,30 @@ The result is a list containing `0`, `1`.
...
-prior to comparison.
-
-### Root Selector
-
-#### Syntax
+### Syntax
 {: unnumbered}
 
-Every valid JSONPath query MUST begin with the root selector `$`.
+Every valid JSONPath query MUST begin with the root identifier `$`.
```
cabo, line 560: What does the word "valid" mean here?
Do we also describe "invalid JSONPath queries"?

This sentence has the disadvantage that it makes @ queries into something different, which then require another word for $ queries and @ queries collectively.

draft-ietf-jsonpath-base.md
```diff
@@ -839,33 +717,29 @@ Queries:
...
 #### Semantics
 {: unnumbered}
 
-An `index-wild-selector`
-selects the nodes of all member values of an object as well as of all elements of an
-array.
-
-Applying the `index-wild-selector` to a primitive JSON value (that is,
-a number, a string, `true`, `false`, or `null`) selects no node.
+A `wildcard` selector selects the nodes of all children of an object or array.
```
cabo, line 738: Did we define "children of an object"?
(These are the values, not the members.)

glyn:
@gregsdennis On the basis of your statement that you don't care about the language so long as the architecture of PR #258 is preserved (which the current PR does preserve), please would you approve this PR?

draft-ietf-jsonpath-base.md
```diff
@@ -879,26 +753,80 @@ JSON:
...
+An index selector `<index>` matches at most one array element value.
+
+~~~~ abnf
+index-selector  = int                             ; decimal integer
+
+int             = ["-"] ( "0" / (DIGIT1 *DIGIT) ) ; -  optional
+DIGIT1          = %x31-39                         ; 1-9 non-zero digit
+~~~~
+
+Applying the numerical `index-selector` is required to select the corresponding
```
cabo, line 779:
```suggestion
Applying the numerical `index-selector` selects the corresponding
```

draft-ietf-jsonpath-base.md
```diff
@@ -879,26 +753,80 @@ JSON:
...
+
+Notes:
+1. An `index-selector` is an integer (in base 10, as in JSON numbers).
+2. As in JSON numbers, the syntax does not allow octal-like integers with leading zeros such as `01` or `-01`.
 
-### Array Slice Selector {#slice}
+#### Semantics {#index-semantics}
+{: unnumbered}
+
+The `index-selector` applied to an array selects an array element using a zero-based index.
```
cabo, line 789: Should we say what happens when it is not applied to an array?
(Nothing is selected.)

draft-ietf-jsonpath-base.md
```diff
@@ -929,24 +857,24 @@ This section is non-normative.
 
 Array indexing is a way of selecting a particular element of an array using
```
cabo, line 857: Can we say this is the same as with an index selector?

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1230,99 @@ Queries:
...
 
-#### Examples
-{: unnumbered}
+A `dot-member-name-shorthand` of the form `.<member-name>` is shorthand for `['<member-name>']`.
 
-JSON:
+Member names containing characters other than allowed by
+`dot-member-name-shorthand` — such as space (U+0020), minus (U+002D), dot (U+002E)
+or escaped characters which appear in the [name selector semantics section](#name-semantics) —
+MUST NOT be used with the `dot-member-name-shorthand`.
```
cabo, line 1276: I don't understand this sentence.  How can the names be used if you can't even notate them?
This should simply say that the set of member names that can be used is limited by the syntax of dot-member-name-shorthands.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1230,99 @@ Queries:
...
+(In such cases, the `name-selector` syntax MUST be used instead.)
```
cabo, line 1277: This is a weird MUST, as it is a statement of fact ("if the key at hand does not fit the door, the key that does fit MUST be used"???).

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1230,99 @@ Queries:
...
+
+Note that `..` on its own is not a valid __APPENDER__.
 
 #### Semantics
 {: unnumbered}
 
-A descendant selector selects zero or more descendants of a node.
+<!-- The following does not address https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/252 -->
+
+A descendant __APPENDER__ selects zero or more descendants of the input value.
```
cabo: I don't think a segment "selects".  It applies its selector to each of the self-or-descendant axis nodes (we need more words).  The sequence of application is partially deterministic.  Outputs of the selectors are concatenated in the sequence of application.

draft-ietf-jsonpath-base.md
glyn: I believe that context is the current PR. The wildcard selector is defined in section 3.5.2.

I'll have to think about how best to address this because a selector acts on a single node and selects zero or more of its children. A selector _per se_ never selects other descendants. But at this stage in the document, we aren't ready to talk about segments and selectors, so we may need to fudge the wording here.

draft-ietf-jsonpath-base.md
glyn: Adopted in ef09a71faad2e8263b38a61c8360feea1564e343.

draft-ietf-jsonpath-base.md
glyn: Yes we did. The terminology section says this:

> Children (of a node):
>
>    If the node is an array, each of its elements, or if the node is an object, each of its member values (but not its member names). If the node is neither an array nor an object, it has no children.

draft-ietf-jsonpath-base.md
glyn: See the rewording in e6ef8e8460b2c5a6f629811f8d8728f790bf5d86.

draft-ietf-jsonpath-base.md
glyn: I made a start on this in de4e00a42764c9739c2f18ce1559d80c8607f063, but we probably need to do more.

draft-ietf-jsonpath-base.md
glyn: d4c8c39da0f1dc4c69fd1f9d5a6cb297a2e6c3e2 puts the identifiers first, then segments, then some selectors, then the shorthands. (Moving the shorthands up wouldn't seem to be very readable since they need to build on selectors.)

draft-ietf-jsonpath-base.md
glyn: e96a561ef79ac2457c3d44f9bdb0fca77112ca2b deletes the word "valid".

I need to think more about $ and @ terminology.

draft-ietf-jsonpath-base.md
glyn: Er, that paragraph ends with:

> Nothing is selected from a value that is not an array.

draft-ietf-jsonpath-base.md
glyn: I think is simpler fix is to delete the index selector stuff from this section. See 977965ad795158e802bfbf82c0f1a62c7b0b9a2f.

draft-ietf-jsonpath-base.md
glyn: I simplified this in 397f2befc51d47caf6403dccfc6c720468081ad1.

draft-ietf-jsonpath-base.md
glyn: 397f2befc51d47caf6403dccfc6c720468081ad1 deleted this.

draft-ietf-jsonpath-base.md
glyn: I avoided saying "selects" in d5286be84535a41c0aa8e789b1c2d58c5cec5d86.

draft-ietf-jsonpath-base.md
```diff
@@ -1309,96 +1235,94 @@ Queries:
...
-{{synsem-overview}}.
+DIGIT                     =  %x30-39              ; 0-9
+ALPHA                     =  %x41-5A / %x61-7A    ; A-Z / a-z
+~~~~
 
+The `dot-wildcard-shorthand` is shorthand for `[*]`.
 
-#### Examples
+A `dot-member-name-shorthand` of the form `.<member-name>` is shorthand for `['<member-name>']`, but
+can only be used with members names containing certain characters.
```
cabo, line 1278:
```suggestion
can only be used with member names that are composed of certain characters.
```

draft-ietf-jsonpath-base.md
cabo: Thanks.  I note that there is still some repetition in the semantics section, but we can fix that later.

draft-ietf-jsonpath-base.md
cabo: Sounds good :-)

draft-ietf-jsonpath-base.md
cabo: Thanks for looking that up for me 😊

draft-ietf-jsonpath-base.md
cabo: I'll keep this conversation open, but we can do this in the next round.

draft-ietf-jsonpath-base.md
cabo: Right.  Might be useful to have some section headers (we can do this in the next round).

draft-ietf-jsonpath-base.md
cabo: I'll keep this conversation open, but we can do that in the next round.

draft-ietf-jsonpath-base.md
cabo: OK, that works, but could be improved later.  Keeping this open.

cabo (approved):
I'm approving this with 6 open conversations, as these can be addressed in a further PR and this one is now ready to be merged,

glyn:
@cabo I note that merging is currently blocked by @gregsdennis's comment about his desire for PR #258 to be reviewed separately. Since this separate review isn't going to happen, I'm planning on merging anyway. So that I am not accused of acting unilaterally, please could you give me a thumbs up if you agree with this course of action.

draft-ietf-jsonpath-base.md
glyn: @cabo See 48de8fcda69084c2ea850805af8d5041492b1943 for an early statement along the lines you suggested.

draft-ietf-jsonpath-base.md
glyn: @cabo See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/268 for a stab at improving the structure of the overview.

Issue #260, opened 2022-09-20
Insufficient non-determinism


glyn:
Currently, the JSONPath spec is insufficiently non-deterministic.

## Descendant nodelist ordering

According to the current draft, in a descendant nodelist, nodes appear immediately before all their descendants. In the example given in the spec, the selector `$..[*]` applied to the JSON value:
```
{
  "o": {"j": 1, "k": 2},
  "a": [5, 3, [{"j": 4}]]
}
```
_cannot_ produce the result shown in the table:
```
{"j": 1, "k" : 2}
[5, 3, [{"j": 4}]]
1
2
5
3
[{"j": 4}]
{"j": 4}
4
```
since `{"j": 1, "k" : 2}` does not appear immediately before `1` and `2`
(there is an intervening `[5, 3, [{"j": 4}]]`).

Solution: delete the word "immediately".

## List selector ordering

The iteration order of objects is not guaranteed to be the same from one iteration to the next. (For instance, in Go, a JSON object is usually implemented as a map and, according to the [Go language spec](https://go.dev/ref/spec), "The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next".)

So a list with more than one entry applied to an object can result in a nodelist in which there are sublists in distinct orders. This is the case if each list entry in turn is used to select a nodelist from the object.

For example, the JSONPath `$[?1==1, ?1==1]` applied to the value:
```
{ "a" : 1, "b" : 2}
```
could produce the nodelist:
```
`1`
`2`
`2`
`1`
```

## Descendant selector ordering

The descendant selectors can produce similar effects, both in the ordering of the descendant nodelists and in the list form with examples similar to that above. For example, the JSONPath `$[0, 0]..[*]` applied to the value:
```
[ {"a": 1, "b": 2}
]
```
can produce the nodelist:
```
`1`
`2`
`2`
`1`
```

## Non-determinism of repeated runs

Repeated applications of a JSONPath to a value can result in nodelists which are ordered differently. For example, the JSONPath `$[*]` applied to the value:
```
{"a": 1, "b": 2}
```
could produce the nodelist:
```
`1`
`2`
```
on one occasion and the nodelist:
```
`2`
`1`
```
on a subsequent occasion.

## Wildcard and filter selector ordering

Currently, there is no text to define the order of the nodelist resulting from a dot wildcard, index wildcard, or filter selector. The order should preserve array order, but be non-deterministic for objects.

[...]

PR #266, opened 2022-10-16
Issue fixes and editorial changes

glyn:
This PR is a series of changes which are submitted together since they would tread on each other's toes if submitted separately.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/252
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/260

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -565,6 +565,9 @@ consisting of nodes at depth in the input value of N or greater.
 * A query with N segments, where N >= 0, all of which are [child segments](#child-segment),
 produces a nodelist consisting of nodes precisely at depth N in the input value.
 
+If a query may produce a nodelist with more than one possible ordering, a particular implementation
+may also produce distinct orderings in distinct runs of the query.
+
 ## Root Identifier
 
 ### Syntax
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1352,9 +1352,9 @@ A descendant selector of the form `..[<selectors>]` visits each node of the inpu
 its descendants in such an order that:
 
 * nodes of any array are visited in array order, and
-* nodes are visited before all their descendants,
+* nodes are visited before all their descendants.
 
-applies the child segment `[<selectors>]` to each node, and concatenates the resultant nodelists together
+It applies the child segment `[<selectors>]` to each node and concatenates the resultant nodelists together
 in the order in which the nodes were visited.
 
 This definition does not stipulate the order in which the children of an object are visited, since
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1321,20 +1321,16 @@ Note that `..` on its own is not a valid segment.
 #### Semantics
 {: unnumbered}
 
-<!-- The following does not address https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/252 -->
-
 A descendant segment produces zero or more descendants of the input value.
 
-A nodelist enumerating the descendants is known as a _descendant nodelist_ when:
+The resultant nodelist of a descendant selector of the form `..[<selectors>]` is the result of applying the
+the child segment `[<selectors>]` to a nodelist enumerating the node and its descendants such that:
 
 * nodes of any array appear in array order,
 * nodes appear immediately before all their descendants.
 
 This definition does not stipulate the order in which the children of an object appear, since
 JSON objects are unordered.
-
-The resultant nodelist of a descendant segment of the form `..[<selectors>]` is the result of applying
-the child segment `[<selectors>]` to a descendant nodelist.
 
 #### Examples
 {: unnumbered}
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1355,6 +1351,7 @@ Queries:
 | `$..[0]` | `5` <br> `{"j": 4}` | `$['a'][0]` <br> `$['a'][2][0]` | Array values       |
 | `$..[0]` | `{"j": 4}` <br> `5` | `$['a'][2][0]` <br> `$['a'][0]` | Alternative result |
 | `$..[*]` <br> `$..*` | `{"j": 1, "k" : 2}` <br> `[5, 3, [{"j": 4}]]` <br> `1` <br> `2` <br> `5` <br> `3` <br> `[{"j": 4}]` <br> `{"j": 4}` <br> `4` | `$['o']` <br> `$['a']` <br> `$['o']['j']` <br> `$['o']['k']` <br> `$['a'][0]` <br> `$['a'][1]` <br> `$['a'][2]` <br> `$['a'][2][0]` <br> `$['a'][2][0]['j']` | All values    |
+| `$..o`   | `{"j": 1, "k": 2}` | `$['o']` | Input value is included in the enumeration |
```
gregsdennis, line 1354: How does this example differ from the `..j` examples?

glyn: > How does this example differ from the `..j` examples?

The `$..j` example is unaffected by whether or not the input value is part of the enumeration. On the other hand, this example depends on that decision. If the input value was not part of the enumeration, then the resultant nodelist would be empty. So this example demonstrates that the input value is part of the enumeration.

gregsdennis: I don't see how that's the case.  The `..j` examples pull the values at those locations. The `..o` example does the same thing.

glyn: I wonder if the wording of the semantics section needs improving as it seems that you aren't reading it the same way as me.

There are two steps involved:

1. enumerating a nodelist of descendants _plus the input value_, and
2. applying the child segment corresponding to the descendant segment to that nodelist.

With that in mind, does my [previous comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/266#discussion_r996497750) now explain the need for the `$..o` example?

We could add a `$..*` example, but please note this would return just the descendants of the input value since step 2 would apply the child segment `$[*]` which returns all the children of each of its input values (and not the value itself).

glyn: BTW please note that table 2 already includes a `$..*` example with the result:

> all member values and array elements contained in input value

gregsdennis: > With that in mind, does my previous comment now explain the need for the $..o example?

I'm sorry, I still don't understand why this example illustrates the change.  What would the result be without the change (or, what's the alternative interpretation of `..o`)?

gregsdennis: To be clear, I agree that the unit node should be included. I just don't understand how this example drives that point home.

glyn: > I'm sorry, I still don't understand why this example illustrates the change. What would the result be without the change (or, what's the alternative interpretation of `..o`)?

Without the change, the enumerated nodelist would be something like this:
* `{"j": 1, "k": 2}`
* `1`
* `2`
* `[5, 3, [{"j": 4}]]`
* `5`
* `3`
* `[{"j": 4}]`
* `4`

Applying the corresponding child segment `['o']` to this list produces the empty nodelist.

With the change, the enumerated nodelist would include: `{
  "o": {"j": 1, "k": 2},
  "a": [5, 3, [{"j": 4}]]
}` and so applying the corresponding child segment `['o']` produces the nodelist:
* `{"j": 1, "k": 2}`

gregsdennis: Hmmm... I don't see the `..` operating separately from the `o`.  I see it all as a single step because it's defined that way in the syntax.

`..o` says, "navigate recursively and return nodes under `o` properties."  That immediately gives you 

```json
{ "j": 1, "k": 2 }
```

It doesn't say, "navigate recursively, give me all the nodes, _then from that_ give me the `o` properties."

gregsdennis: I now see that you're changing the text to mean the two-step operation, but again, I think that conflicts with the syntax.

I really think the simpler approach is to process the selection _while_ enumerating the (recursive) descendants.

(I know that we're not pushing algorithms, but _something_ needs to be in the spec.)

cabo: I agree with Greg here: Just as a child segment offers `self` to the selectors, a descendants segment offers `self` and `descendant` to the selectors.  Having the `$..o` example to illustrate this is a good idea.

(This and a few other examples need to be fixed if we stick with `"j j"` and `"k.k"`.)

glyn: Please see 17835689a97eae026254e7cc28197817cbfbcf21 in which I've expressed the semantics the way you suggest.

I believe the semantics are unchanged.
 
> (Other examples need to be fixed if we stick with `"j j"` and `"k.k"`.)

Not sure what this is referring to because the examples in this section mention neither `"j j"` nor `"k.k"`. However, I've pushed 20bb114990b615da55a7be662e5bb37e218182b7 which deletes the erroneous child selector examples.

cabo: > However, I've pushed 20bb114 which deletes the erroneous child selector examples.

Thanks.

draft-ietf-jsonpath-base.md
```diff
@@ -1321,20 +1346,24 @@ Note that `..` on its own is not a valid segment.
...
+A descendant selector of the form `..[<selectors>]` visits each node of the input value and
+its descendants in such an order that:
 
-* nodes of any array appear in array order,
-* nodes appear immediately before all their descendants.
+* nodes of any array are visited in array order, and
+* nodes are visited before all their descendants,
 
-This definition does not stipulate the order in which the children of an object appear, since
+applies the child segment `[<selectors>]` to each node, and concatenates the resultant nodelists together
```
gregsdennis, line 1357: Was this meant to start a new sentence? It seems to start in the middle of one.

gregsdennis (approved):
This sounds a lot better.  Just a minor grammatical issue.

draft-ietf-jsonpath-base.md
glyn: No, but a sentence spanning the bullets was too sprawling, so I cut it in two. Thanks.

PR #268, opened 2022-10-17
Improve overview structure

glyn:
This is a stab at addressing one @cabo's review comments in PR #263.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -297,10 +297,22 @@ stable JSON query language.
 ## Overview of JSONPath Expressions {#overview}
 
 JSONPath expressions are applied to a JSON value, the *argument*.
-Within the JSONPath expression, the abstract name `$` is used to refer
-to the *root node* of the argument, i.e., to the argument as a whole.
 
-JSONPath expressions use the *bracket notation*, for example:
+A JSONPath expression consists of an identifier followed by a series
+of zero or more segments each of which contains one or more selectors.
+
+### Identifiers
+
+The _root node identifier_ `$` refers to the *root node* of the argument,
+i.e., to the argument as a whole.
+Every JSONPath expression begins with the root node identifier.
+
+The _current node identifier_ `@` refers to the *current node* in the context
+of a filter expression (described later).
+
+### Segments
+
+Segments can use the *bracket notation*, for example:
 
 ~~~~
 $['store']['book'][0]['title']
@@ -312,10 +324,14 @@ or the more compact *dot notation*, for example:
 $.store.book[0].title
 ~~~~
 
-to build paths that are input to a JSONPath implementation.
-A single path may use a combination of bracket and dot notations.
+A JSONPath expression may use a combination of bracket and dot notations.
 
 Dot notation is a shorthand way of writing certain bracket notations.
+
+This document treats the bracket notations as canonical and defines various shorthands in terms
+of bracket notation.
+
+### Selectors
 
 A wildcard `*` ({{wildcard}}) in the expression `[*]` selects all children of an
 object or an array and in the expression `..[*]` selects all descendants of an object or an array.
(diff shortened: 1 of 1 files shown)


glyn:
@cabo If you don't think this PR is going in the right direction, please close it and we'll circle back to the matter later.

draft-ietf-jsonpath-base.md
```diff
@@ -296,10 +296,22 @@ stable JSON query language.
...
-to the *root node* of the argument, i.e., to the argument as a whole.
 
-JSONPath expressions use the *bracket notation*, for example:
+A JSONPath expression consists of an identifier followed by a series
+of zero or more segments each of which contains one or more selectors.
+
+### Identifiers
+
+The _root node identifier_ `$` refers to the *root node* of the argument,
+i.e., to the argument as a whole.
```
timbray, line 306: no comma after i.e. I think?

draft-ietf-jsonpath-base.md
```diff
@@ -296,10 +296,22 @@ stable JSON query language.
...
+Every JSONPath expression begins with the root node identifier.
+
+The _current node identifier_ `@` refers to the *current node* in the context
+of a filter expression (described later).
```
timbray, line 310: I think this really means "in the context of the evaluation of a filter expression"?

draft-ietf-jsonpath-base.md
```diff
@@ -311,11 +323,15 @@ or the more compact *dot notation*, for example:
 $.store.book[0].title
 ~~~~
 
-to build paths that are input to a JSONPath implementation.
-A single path may use a combination of bracket and dot notations.
+A JSONPath expression may use a combination of bracket and dot notations.
 
 Dot notation is merely a shorthand way of writing certain bracket notations.
 
+This document treats the bracket notations as canonical and defines various shorthands in terms
```
timbray, line 330: Do you mean "as canonical but uses the bracket shorthand in examples and descriptions"?

timbray (dismissed):

draft-ietf-jsonpath-base.md
cabo: CMOS (the basis for RFC 7322) was in the "always comma after i.e. and e.g." camp last time I looked.
I have only the kindle version of CMOS 17, and that is so broken that I don't find the place where it says so.
I do find the occasional RFC that gets this wrong, overwhelmingly below RFC 5100.

draft-ietf-jsonpath-base.md
cabo: Yes, this could be expanded.  The idea is that we define JSONPath in terms of bracketed steps, but as syntactic sugar we also define a few shorthands.

draft-ietf-jsonpath-base.md
cabo: Found it:

```
i.e.; e.g. The first is the abbreviation for id est ("that is") ; the second is the abbreviation for exempli gratia ("for example") . The English equivalents are preferable in formal prose, though sometimes the compactness of these twocharacter abbreviations makes them de sirable. Always put a comma after ei
ther one. See also 6.20, 6.51.
```

(totally broken formatting, but very clear rule).

timbray: bah snarl grumble. Having said that I'd rather be standardized than idiomatic.

I believe that written English should be consistent with the spoken form, and when I speak, I say “refers to the root node of the argument *pause* that is to the  argument as a whole" without a pause after "that is". But OK.

cabo: Amazingly, I pause after the "that is" as well.

Of course, there is the general discussion of Fowler's "punctuation pitch" and the difference in US usage (pretty unanimously requiring the comma) and British usage (more varied).

But standardization is good, and the RFC editor has opined for CMOS, so there we are.

draft-ietf-jsonpath-base.md
goessner: (Somehow off-topic here) I would like to avoid collision of traditional term "current node" in this context with its use in 3.3 Semantics
```
A segment performs its function on each of the nodes in its input
   nodelist, during such a function execution, such a node is referred
   to as the "current node",
```
There I propose to exchange  "current node" by "context node", which is used by XPath in a similar context.

glyn: > (Somehow off-topic here) I would like to avoid collision of traditional term "current node" in this context with its use in 3.3 Semantics
> 
> ```
> A segment performs its function on each of the nodes in its input
>    nodelist, during such a function execution, such a node is referred
>    to as the "current node",
> ```
> 
> There I propose to exchange "current node" by "context node", which is used by XPath in a similar context.

@goessner Well spotted. Please can we address this separately to this PR? I suggest starting with an issue to hash out a consensus on the terminology.

glyn: > (Somehow off-topic here) I would like to avoid collision of traditional term "current node" in this context with its use in 3.3 Semantics
> 
> ```
> A segment performs its function on each of the nodes in its input
>    nodelist, during such a function execution, such a node is referred
>    to as the "current node",
> ```
> 
> There I propose to exchange "current node" by "context node", which is used by XPath in a similar context.

@goessner Well spotted. Please can we address this separately to this PR? I suggest starting with an issue to hash out a consensus on the terminology.

> (Somehow off-topic here) I would like to avoid collision of traditional term "current node" in this context with its use in 3.3 Semantics
> 
> ```
> A segment performs its function on each of the nodes in its input
>    nodelist, during such a function execution, such a node is referred
>    to as the "current node",
> ```
> 
> There I propose to exchange "current node" by "context node", which is used by XPath in a similar context.

This discussed was moved to https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/285.

glyn: > I think this really means "in the context of the evaluation of a filter expression"?

Adopted.

draft-ietf-jsonpath-base.md
glyn: I added a sentence along the lines @timbray suggested.

cabo (approved):

goessner (approved):

Issue #265, opened 2022-10-15
Ordering of descendant segment with multiple selectors


glyn:
The current draft does not agree with the [implementation consensus](https://cburgmer.github.io/json-path-comparison/results/union_with_keys_after_recursive_descent.html) when a descendant segment has multiple selectors. The current draft says that a corresponding child segment containing the whole list of selectors should be applied to a descendant nodelist. The implementation consensus is that each selector is be applied to a descendant nodelist and the results concatenated together.

Let's take an example involving arrays (to avoid non-determinism confusing matters). The JSONPath `$..[0, 1]` applied to the value `[1, 2, [3, 4]]` according to the current draft gives:
```
1
2
3
4
```
but according to the implementation consensus gives:
```
1
3
2
4
```

One of the more influential implementations (Jayway) agrees with the current draft.

I think it is best to go with the current draft behaviour as this can be implemented in a single traversal of the descendants, but wanted to confirm this here.

glyn:
Another advantage of the current draft behaviour is that it is more deterministic than the implementation consensus. For example, the JSONPath `$..['a', 'b']` applied to the value `{"x" : {"a" : 1, "b" : 2}, "y" : {"a" : 3, "b" : 4}}` according to the current draft gives:
```
1
2
3
4
```
or
```
3
4
1
2
```
but according to the implementation consensus (with the non-determinism of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/260) gives:
```
1
3
2
4
``` 
or
```
1
3
4
2
```
or
```
3
1
2
4
```
or
```
3
1
4
2
```

cabo:
I am very much in favor of cleaning this up.

(If the odd behaviors are very widespread, we could combine the cleaned-up semantics with an interoperability warning that applications that rely on the exact sequence will be less interoperable than ones that don't.  That is, of course, true of most of what the standard is providing...)

gregsdennis:
I agree that the current draft is better.

gregsdennis:
Reading the spec again,

> A descendant segment produces zero or more descendants of the input value.
> 
> A nodelist enumerating the descendants is known as a descendant nodelist when:
> 
> - nodes of any array appear in array order,
> - nodes appear immediately before all their descendants.
> 
> This definition does not stipulate the order in which the children of an object appear, since JSON objects are unordered.
> 
[...]

PR #269, opened 2022-10-18
Ordering of descendant segment with multiple selectors

glyn:
Give an example to emphasise the spec behaviour. At the time of writing, this disagrees with the implementation consensus here:

https://cburgmer.github.io/json-path-comparison/results/union_with_keys_after_recursive_descent.html

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/265

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1377,7 +1377,6 @@ Note that `D1` is the input value.
 
 For each `i` such that `1 <= i <= n`, the nodelist `Ri` is defined to be a result of applying
 the child segment `[<selectors>]` to the node `Di`.
-Note that `Ri` is the concatenation of the nodelists from each of the selectors.
 
 The result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1376,7 +1376,7 @@ Suppose the nodes, in the order visited, are `D1`, ..., `Dn` (where `n >= 1`).
 Note that `D1` is the input value.
 
 For each `i` such that `1 <= i <= n`, the nodelist `Ri` is defined to be a result of applying
-the child segment `[<selectors>]` (with all its selectors) to the node `Di`.
+the child segment `[<selectors>]` to the node `Di`.
 Note that `Ri` is the concatenation of the nodelists from each of the selectors.
 
 The result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1363,20 +1363,24 @@ Note that `..` on its own is not a valid segment.
 
 A descendant segment produces zero or more descendants of the input value.
 
-A descendant selector of the form `..[<selectors>]` visits each node of the input value and
-its descendants in such an order that:
+A descendant selector visits the input value and each of
+its descendants such that:
 
 * nodes of any array are visited in array order, and
-* nodes are visited before all their descendants.
+* nodes are visited before their descendants.
 
-It applies the child segment `[<selectors>]` to each node and concatenates the resultant nodelists together
-in the order in which the nodes were visited.
-
-This definition does not stipulate the order in which the children of an object are visited, since
+The order in which the children of an object are visited is not stipulated, since
 JSON objects are unordered.
 
-Where a selector can produce a nodelist in more than one possible order, the selector may produce nodelists in distinct
-orders each time it appears in the descendant segment.
+Suppose the nodes, in the order visited, are `D1`, ..., `Dn` (where `n >= 1`).
+Note that `D1` is the input value.
+
+For each `i` such that `1 <= i <= n`, the nodelist `Ri` is defined to be a result of applying
+the child segment `[<selectors>]` (with all its selectors) to the node `Di`.
+Note that `Ri` is the concatenation of the nodelists from each of the selectors.
+
+The result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
+Note that only a single traversal of the input value and its descendants is required.
 
 So a descendant segment drills down one or more levels into the structure of the input value.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1380,7 +1380,6 @@ the child segment `[<selectors>]` (with all its selectors) to the node `Di`.
 Note that `Ri` is the concatenation of the nodelists from each of the selectors.
 
 The result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
-Note that only a single traversal of the input value and its descendants is required.
 
 So a descendant segment drills down one or more levels into the structure of the input value.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1387,7 +1387,7 @@ JSON:
 
     {
       "o": {"j": 1, "k": 2},
-      "a": [5, 3, [{"j": 4}]]
+      "a": [5, 3, [{"j": 4}, {"k": 6}]]
     }
 
 Queries:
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1381,23 +1381,29 @@ Queries:
 | `$..j`   | `4` <br> `1` | `$['a'][2][0]['j']` <br> `$['o']['j']` | Alternative result |
 | `$..[0]` | `5` <br> `{"j": 4}` | `$['a'][0]` <br> `$['a'][2][0]` | Array values       |
 | `$..[0]` | `{"j": 4}` <br> `5` | `$['a'][2][0]` <br> `$['a'][0]` | Alternative result |
-| `$..[*]` <br> `$..*` | `{"j": 1, "k" : 2}` <br> `[5, 3, [{"j": 4}]]` <br> `1` <br> `2` <br> `5` <br> `3` <br> `[{"j": 4}]` <br> `{"j": 4}` <br> `4` | `$['o']` <br> `$['a']` <br> `$['o']['j']` <br> `$['o']['k']` <br> `$['a'][0]` <br> `$['a'][1]` <br> `$['a'][2]` <br> `$['a'][2][0]` <br> `$['a'][2][0]['j']` | All values    |
+| `$..[*]` <br> `$..*` | `{"j": 1, "k" : 2}` <br> `[5, 3, [{"j": 4}, {"k": 6}]]` <br> `1` <br> `2` <br> `5` <br> `3` <br> `[{"j": 4}, {"k": 6}]` <br> `{"j": 4}` <br> `4` <br> `{"k": 6}` <br> `6` | `$['o']` <br> `$['a']` <br> `$['o']['j']` <br> `$['o']['k']` <br> `$['a'][0]` <br> `$['a'][1]` <br> `$['a'][2]` <br> `$['a'][2][0]` <br> `$['a'][2][0]['j']` <br> `$['a'][2][1]` <br> `$['a'][2][1]['k']` | All values    |
```
glyn, line 1384: Perhaps this example is becoming a bit unwieldy and we should have more than one JSON document, each with their own table of queries. But, otherwise, I quite like keeping to a single JSON document and table per group of examples.

draft-ietf-jsonpath-base.md
```diff
@@ -1381,23 +1381,29 @@ Queries:
...
 | `$..o`   | `{"j": 1, "k": 2}` | `$['o']` | Input value is visited |
 | `$.o..[*, *]` | `1` <br> `2` <br> `2` <br> `1` | `$['o']['j']` <br> `$['o']['k']` <br> `$['o']['k']` <br> `$['o']['j']` | Non-deterministic ordering |
+| `$.a..[0, 1]`| `5` <br> `3` <br> `{"j": 4}` <br> `{"k": 6}` | `$['a'][0]` <br> `$['a'][1]` <br> `$['a'][0][2][0]` <br> `$['a'][0][2][1]` | Multiple segments |
```
gregsdennis, line 1387: Is it necessary to have the `.a` in here?  I thought we were trying to illustrate the multi-selector behavior.  It seems having the additional segment muddies the example.

gregsdennis:
Please see [my comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/265#issuecomment-1282967442) on the issue.

I think we need more semantic text to define what we're looking for.

glyn:
> Please see [my comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/265#issuecomment-1282967442) on the issue.
> 
> I think we need more semantic text to define what we're looking for.

I've reworked the text in c7f54040f56756c2dd10e05cebbaa0d1a8fb31c1 to make the semantics much more obvious.

draft-ietf-jsonpath-base.md
```diff
@@ -1346,20 +1349,24 @@ Note that `..` on its own is not a valid segment.
...
+The order in which the children of an object are visited is not stipulated, since
 JSON objects are unordered.
 
-Where a selector can produce a nodelist in more than one possible order, the selector may produce nodelists in distinct
-orders each time it appears in the descendant segment.
+Suppose the nodes, in the order visited, are `D1`, ..., `Dn` (where `n >= 1`).
+Note that `D1` is the input value.
+
+For each `i` such that `1 <= i <= n`, the nodelist `Ri` is defined to be a result of applying
+the child segment `[<selectors>]` (with all its selectors) to the node `Di`.
```
gregsdennis, line 1365:
```suggestion
For each descendant node `Di` such that `1 <= i <= n`, the resulting nodelist `Ri` is defined to be a result of applying
the all of the selectors in the child segment `[<selectors>]` to the node `Di`.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1346,20 +1349,24 @@ Note that `..` on its own is not a valid segment.
...
+Note that `Ri` is the concatenation of the nodelists from each of the selectors.
+
+The result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
+Note that only a single traversal of the input value and its descendants is required.
```
gregsdennis, line 1369:
```suggestion
Suppose the nodes, in the order visited, are `D1`, ..., `Dn` where `n >= 1`.
(Note that `D1` is the input value.)

For each node `Di` such that `1 <= i <= n`, the resulting nodelist `Ri` is defined to be a result of applying
the all of the selectors in the child segment `[<selectors>]` to the node `Di`.

The final result of the descendant selector is the concatenation of `R1`, ..., `Rn` (in that order).
```

I fear that

> Note that only a single traversal of the input value and its descendants is required.

may be getting too close to specifying implementation logic.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
glyn: Ok, I'll delete the single traversal sentence.

I didn't adopt the proposal because it talks about applying all the selectors in the child segment, but we need child segment semantics (specifically, concatenating the selector nodelists) instead of just selector semantics.

gregsdennis: My suggestion also provides a more precise wording and adjusts the parenthetical phrases.

> it talks about applying all the selectors in the child segment, but we need child segment semantics (specifically, concatenating the selector nodelists) instead of just selector semantics

Then I suggest we remove mention of the selectors completely.

```
Suppose the nodes, in the order visited, are `D1`, ..., `Dn` where `n >= 1`.
(Note that `D1` is the input value.)

For each node `Di` such that `1 <= i <= n`, the intermediate nodelist `Ri` is the result of applying
the child segment `[<selectors>]` to the node `Di`.

The result of the descendant segment is the concatenation of the intermediate nodelists `R1`, ..., `Rn` (in that order).
```

We also need to be sure that in the "child segment" section, we define how the selector results are concatenated, which it seems we do:

> The resulting nodelist of a child segment is the concatenation of the nodelists from each of its selectors in the order that the selectors appear in the list.

glyn: > Then I suggest we remove mention of the selectors completely.

I can live with that. See 3516bee163b81457e64be5dabe2d417893d38bbb.

gregsdennis: Why are you not taking the rest of the suggestion?

glyn: Ah, sorry, I should also have deleted "Note that `Ri` is the concatenation of the nodelists from each of the selectors.". Fixed in 0d25c700fb7e02ec7db97bf3acc3016085384fe5.

As for the rest of the suggestion, I prefer the current wording in the PR:
* I prefer to define `Ri` via universal quantification over `i`, rather than over `Di`.
* I don't see any advantage in introducing the extra term "intermediate nodelist".

Are you comfortable with approving now?

glyn: Thanks for explaining your opinions. I beg to differ. Is this a blocker for you or can you now approve the PR?

gregsdennis (approved):

draft-ietf-jsonpath-base.md
glyn: Yes, but it's only an artefact of having a single JSON document and group of examples per section of examples. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/269/files#r998358723. Let's see how this reads later.

Issue #195, opened 2022-06-14
Do we really require a singular path for `exist-expr`


cabo:
Usually, existence could be asked in the form "is there any...".
We restrict exist-expr to singular paths, though.
Do we really want to make this restriction?

gregsdennis:
I'm trying to think how this would be done practically.  All the examples for existence testing are in the form `$[?(@.foo)]` (as you mention).

What sorts of plural paths are you suggesting could also test for existence?  The only one I can think of would be `$[?(@.*)]` which I suppose could return any non-empty objects.

cabo:
```
[
  [ {"normal": 3711, "fallback": 815},
    {"normal": 4711},
    {"normal": 5711, "fallback": 615}
  ],
  ...
]
```

`$[?(@.*.fallback)]` -- which top level elements have sub-elements with a fallback member?

danielaparker:
It's not obvious to me what you're expecting here.

Consider
```json
[
  [ {"normal": 3711, "fallback": 815},
    {"normal": 4711, "fallback": 10},
    {"normal": 5711, "fallback": 615}
  ],
  [ {"normal": 3711, "fallback": 815},
    {"normal": 4711},
    {"normal": 5711, "fallback": 615}
  ],
  [ {"normal": 3711},
    {"normal": 4711},
    {"normal": 5711}
  ]
]
```
If we try `$[?(@.*.fallback)]` with Jayway JSONPath, an implementation that evaluates `@.foo` as a test for existence, and supports arbitrary JSONPath expressions in filter predicates, we obtain

```json
[
   [
      {
         "normal" : 3711,
         "fallback" : 815
      },
      {
         "normal" : 4711,
         "fallback" : 10
      },
      {
         "normal" : 5711,
         "fallback" : 615
      }
   ]
]
```
With Jayway semantics, all the evaluations of `@.*.fallback` have to return true for `@` to be accepted.

glyn:
@danielaparker yes, good question.

Note Jayway's logical treatment of existence of all members of an empty set.  Evaluating `$[?(@.*.fallback)]` against
```
[
[
]
]
```
yields:
```
[
   [
      
   ]
]
```

cabo:
Obviously, if the path is no longer singular, there needs to be a definition if this is a ∀ for-all (as in Jayway) or an ∃ exists (as I would have expected from the name).

gregsdennis:
I see no problem with this.

[...]

PR #271, opened 2022-10-18
Existence with non-singular paths

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/195

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1108,9 +1108,9 @@ Children of an array appear in array order in the resultant nodelist.
 ##### Existence Tests
 {: unnumbered}
 
-A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select a node.
+A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
 
-Existence tests are different to comparisons in that:
+Existence tests differ from comparisons in that:
 
 * they work with arbitrary relative or absolute paths (not just Singular Paths).
 * they work with paths that select structured values.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -482,8 +482,9 @@ stands for a nodelist that contains the root node of the argument,
 followed by a possibly empty sequence of *segments*.
 
 ~~~~ abnf
-json-path = root-identifier *(S (child-segment               /
-                                 descendant-segment))
+json-path = root-identifier segments
+segments  =  *(S (child-segment /
+                  descendant-segment))
 ~~~~
 
 The syntax and semantics of each segment are defined below.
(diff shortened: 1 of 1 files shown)


gregsdennis (commented):
I'm happy with the textual changes. I'd prefer @cabo or someone else look at the ABNF.

draft-ietf-jsonpath-base.md
```diff
@@ -1111,8 +1110,12 @@ filter selector iterates over the array or object.
 ##### Existence Tests
 {: unnumbered}
 
-A singular path by itself in a Boolean context is an existence test which yields true if the path selects a node and yields false if the path does not select a node.
-This existence test — as an exception to the general rule — also works with nodes with structured values.
+A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select a node.
```
timbray, line 1113: maybe "... does not select any nodes"?

draft-ietf-jsonpath-base.md
```diff
@@ -1111,8 +1110,12 @@ filter selector iterates over the array or object.
...
+
+Existence tests are different to comparisons in that:
```
timbray, line 1115: Interestingly, North Americans say "different from" not "different to". Possibly the RFC editor will impose North American style?

timbray (commented):
What Greg said, language LGTM but would like @cabo's OK on the grammar

draft-ietf-jsonpath-base.md
glyn: I'm equally comfortable with either phrase, so I changed to use the suggestion.

draft-ietf-jsonpath-base.md
glyn: We should aim for North American English. I changed the wording to "Existence tests differ from comparisons in that:" which I think reads well in either British or North American English.

cabo (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1011,27 +1012,19 @@ During the iteration process each array element or object member is visited and
...
-name-segment      = "[" name-selector "]" / dot-member-name-shorthand
-index-segment     = "[" index-selector "]"
+boolean-expr           = logical-or-expr
+logical-or-expr        = logical-and-expr *(S "||" S logical-and-expr)
+                                                            ; disjunction
+                                                            ; binds less tightly than conjunction
+logical-and-expr       = basic-expr *(S "&&" S basic-expr)  ; conjunction
+                                                            ; binds more tightly than disjunction
+
+basic-expr              = exist-expr /
```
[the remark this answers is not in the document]
glyn: > I would group this paren-expr, relation-ex, exist-expr.

Ok. The current order of `basic-expr` matches the order in which the alternatives are specified lower down. Feel free to regroup this later. 

> 
> The ABNF here conflicts with another PR already merged, but this particular change looks innocuous otherwise.

The delta clash was a little painful. It is now resolved and I hope I haven't lost any pieces in the process!
> 
> The cutting down to 69 columns needs to be done at some point; maybe after this merge will be a good start.

Eek! Ouch! That's very narrow. Is someone wanting to print this spec on punched cards?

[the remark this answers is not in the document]
cabo: While RFC 2223 is obsoleted by RFC 7322, we still follow its Section 3a, which says:

```
      Each line must be limited to 72 characters followed by carriage
      return and line feed.
```

With the usual three characters indentation, this leaves us 69 characters.

Issue #274, opened 2022-10-18
Table 15


timbray:
The last entry in table 15 is completely incomprehensible and the explanation in the following paragraph doesn't help.  Not sure how to work around this within the limitations of xml2rfc but maybe just leave it out, it's implicit in the 2nd-last-line example.

glyn:
@timbray I presume you are referring to [Table 15 in draft 07](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-07.html#table-15) with the caption "Normalized Path examples". (Table 15 at `HEAD` is a different one.)

I agree the example is a mess and doesn't pull its weight, so I'll submit a PR to delete it.

cabo:
I think it is useful to have an example that shows characters are not normally escaped, even if they are beyond ASCII.

cabo:
(By the time this goes by the RFC editor, we may be able to get rid of the useless U+.... stuff added by xml2rfc.)

PR #277, opened 2022-10-19
Delete messy example

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/274


timbray (approved):

gregsdennis:
Do we intend to replace this with anything? What is its purpose in being here?

timbray:
The N-1 line also illustrates the point so I think it's probably OK to leave it?

cabo:
Sorry, I didn't pay attention.
The purpose of this example was to show that escaped characters are normalized into their actual equivalents (unless there is a reason not to do so).

The messiness is a workaround for xml2rfc bugs.  I'm confident we can get these bugs fixed before this is published; otherwise we might indeed want to skip this example.

glyn:
Why doesn't the `$["\u0061"]` example already shows that an escaped character is normalised into its actual equivalent ("a")?

PR #279, opened 2022-10-24
Editorial nits and suggestions from Tim Bray

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -1298,6 +1294,7 @@ The `dot-wildcard-shorthand` is shorthand for `[*]`.
 
 A `dot-member-name-shorthand` of the form `.<member-name>` is shorthand for `['<member-name>']`, but
 can only be used with member names that are composed of certain characters.
+Thus, for example, `$.foo.bar` is shorthand for `$['foo']['bar']` (and not for `$['foo.bar']`).
```
timbray, line 1297: Would still like to see a caution for developers against using `"$" + join(segments, ".")`

timbray (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -697,17 +692,17 @@ in the table below:
...
-strings, with those strings coming from the JSONPath and from member
-names and string values in the JSON to which it is being applied.
+Note that processing the name selector requires comparing the member name string M
+with member name strings in the JSON to which the selector is being applied.
 Two strings MUST be considered equal if and only if they are identical
 sequences of Unicode scalar values. In other words, normalization operations
-MUST NOT be applied to either the string from the JSONPath or from the JSON
-prior to comparison.
+MUST NOT be applied to either the member name string M from the JSONPath or to
+the member name strings in the JSON prior to comparison.
```
cabo, line 705: "in the JSON" is (and was) not very preccise.

"in the object node"?

draft-ietf-jsonpath-base.md
```diff
@@ -697,17 +692,17 @@ in the table below:
...
+Applying the `name-selector` to an object node
+selects a member value whose name equals the member name M,
+or selects nothing if there is no such member value.
 Nothing is selected from a value that is not a object.
 
-Note that processing the name selector potentially requires matching strings against
-strings, with those strings coming from the JSONPath and from member
-names and string values in the JSON to which it is being applied.
+Note that processing the name selector requires comparing the member name string M
+with member name strings in the JSON to which the selector is being applied.
```
cabo, line 701: (see below)

draft-ietf-jsonpath-base.md
```diff
@@ -777,7 +772,8 @@ The following examples show the `wildcard` selector in use by a child segment.
 {: title="Wildcard selector examples"}
 
 The example above with the query `$.o[*, *]` shows that the wildcard selector may produce nodelists in distinct
```
cabo, line 774: Well, the example does not show the parentheses.

s/but not/note that this is not the case/

draft-ietf-jsonpath-base.md
cabo: This caution should then also apply to "$" + segments.map { "['" + _1 + "']" }.join
What would be a good place for this?
(Security considerations comes to mind.)

draft-ietf-jsonpath-base.md
```diff
@@ -1446,6 +1447,7 @@ A Normalized Path is a JSONPath with restricted syntax that identifies a node by
 the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
 `$['book'][3]` and `$['book'][5]`. For a given JSON value, there is a one to one correspondence between the value's
 nodes and the Normalized Paths that identify these nodes.
+Note that there is at precisely one Normalized Path that identifies each node.
```
cabo, line 1450: s/at //

draft-ietf-jsonpath-base.md
```diff
@@ -1457,6 +1459,8 @@ Certain characters are escaped, in one and only one way; all other characters ar
 
 Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-3]`, for example, is a Singular
 Path, but not a Normalized Path.
+The Normalized Path equivalent to `$[-3]` would have an index equal to the array length minus `3`.
+(The array length must be at least `3` if `$[-3]` identifies a node.)
```
cabo, line 1463: s/identifies/is to identify/

cabo (approved):
Sorry for the delay.  For some reasons, I couldn't make suggestions, so I fell back to s/// notation.
Most of the suggestions (i.e., not the typos) could be put into a new issue if they need to much time, so this one merged.

draft-ietf-jsonpath-base.md
timbray: I think it's just not member name strings in JSONPath expressions, it's also values in filters, right?

draft-ietf-jsonpath-base.md
timbray: Sorry, don't understand the problem with `"$" + segments.map { "['" + _1 + "']" }.join`

cabo: The same problem as with `segments.join(".")` -- the segment might contain characters that are interpreted, such as a single quote or a backslash.

draft-ietf-jsonpath-base.md
cabo: Yes, but this probably should be said where those comparisons are described.
(It already gives the algorithm, but doesn't explicitly talk about normalization.

draft-ietf-jsonpath-base.md
glyn: > s/at //

Fixed.

draft-ietf-jsonpath-base.md
glyn: > s/identifies/is to identify/

Adopted.

draft-ietf-jsonpath-base.md
glyn: Sorry, I don't follow either of these comments. Anyway, I'm going to merge this PR, so let's take up the sequel elsewhere.

PR #280, opened 2022-10-30
Function extension mechanism and several functions

cabo:


draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+The arguments are filter expressions that are to be interpreted as
+nodelists or values.
+For each argument and for the result, the function extension defines
+the "kind" of the item, i.e., whether the item is a value or a
+nodelist ("nodes").
+
+~~~ abnf
+function-name           = function-name-first *function-name-char
+function-name-first     = LCALPHA
+function-name-char      = DIGIT / name-first / "_"
```
glyn, line 1212: s /name-first/function-name-first/

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+LCALPHA                 = %x61-7A  ; "a".."z"
+
+
+function-expression     = function-name "(" [*(function-argument "," S) function-argument] ")"
```
glyn, line 1216: Do we want to allow more whitespace, e.g. just inside the parentheses?

glyn, line 1216: The following is more consistent with the rest of our ABNF:
```suggestion
function-expression     = function-name "(" [function-argument *("," S function-argument)] ")"
```

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+singular-path           =/ function-expression
```
glyn, line 1217: Aside: I like incremental alternatives and if we keep them here, we may want to use them elsewhere, e.g. in the definition of `selector`.

glyn, line 1217: If a function returns a nodelist, how can the result be used as a singular path?

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+function-argument       = comparable
+~~~
+
+Syntactically, a function-expression can occur anywhere where a
+singular-path can occur (exist-expr, comparable, regex-expr).
+A function-expression that employs a function extension that returns a
+nodelist MUST return a singular node to be used in a comparable or a
+regex-expr.
```
glyn, line 1225: If we fix #195, I think the language here can be simplified. Until then, we need to include `exist-expr` in the places where a singular node must be returned.

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+~~~ JSONPath
+$[?length(@.authors) >= 5]
+~~~
+
+Its only argument is a value (possibly taken from a singular path as
+in the example above).  The result also is a value; if the argument
+value is an array, the result is the number of elements in the array.
+[length-mismatch]
+
+[length-mismatch]: Define the case where the argument is not an array as well.
```
glyn, line 1248: It seems we have limited options if we are going to be consistent with the rest of the spec. We could either return a fixed value (e.g. `-1`) or an empty nodelist. The latter seems preferable as it's consistent with a singular path possibly being empty.

This makes me wonder whether we should make a function always return a nodelist and have the function spec say whether or not the returned nodelist is "singular" (meaning it has at most one member) instead of allowing nodelists of values to be returned.

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
 : no
 
+## Function Extensions {#iana-fnex}
+
+This specification defines a new "Function Extensions sub-registry" in
+a new "JSONPath Parameters registry", with the policy "expert review"
+({{Section 4.5 of -ianacons}}).
+
+The expert is instructed to be frugal in the allocation of very short
+function extension names, keeping them in reverse for applications that
```
glyn, line 1673:
```suggestion
function extension names, keeping them in reserve for applications that
```

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+
+function-expression     = function-name "(" [*(function-argument "," S) function-argument] ")"
+singular-path           =/ function-expression
+function-argument       = comparable
+~~~
+
+Syntactically, a function-expression can occur anywhere where a
+singular-path can occur (exist-expr, comparable, regex-expr).
+A function-expression that employs a function extension that returns a
+nodelist MUST return a singular node to be used in a comparable or a
```
glyn, line 1224: Rather than a singular node, perhaps return a singular nodelist (see below).

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+add filter expression functionality: "Function Extensions".
+
+A function extension defines a registered name (see {{iana-fnex}}) that
+can be applied to a sequence of zero or more arguments, resulting in a result.
+A function-expression stands for ("returns") a nodelist or a value.
+The arguments are filter expressions that are to be interpreted as
+nodelists or values.
+For each argument and for the result, the function extension defines
+the "kind" of the item, i.e., whether the item is a value or a
+nodelist ("nodes").
```
glyn, line 1207: Perhaps rather than the item being  a value, it could be a singular nodelist. This allows chaining the result of a function as an argument of another function with propagation of the empty nodelist. Whether we make a general rule that empty argument nodelists MUST produce empty result nodelists or whether this is a per-function optional behaviour is something we should decide. I quite like the uniformity and predictability of a general rule.

draft-ietf-jsonpath-base.md
glyn, line 1212:
```suggestion
function-name-char      = DIGIT / function-name-first / "_"
```

draft-ietf-jsonpath-base.md
cabo: space inside parentheses is evil.  Except for line breaks...

draft-ietf-jsonpath-base.md
cabo: This was an expedient device to fit this in without touching everything -- should be fixed IMHO

draft-ietf-jsonpath-base.md
cabo: The main problem is that the node that has the length isn't in the input JSON value

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
 
+## Function Extensions {#iana-fnex}
+
+This specification defines a new "Function Extensions sub-registry" in
+a new "JSONPath Parameters registry", with the policy "expert review"
+({{Section 4.5 of -ianacons}}).
+
+The expert is instructed to be frugal in the allocation of very short
+function extension names, keeping them in reverse for applications that
+are likely to enjoy wide use and can make good use of their shortness.
```
glyn, line 1674: Not sure about "can make good use of their shortness". When would a short name not be made good use of?

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
+The expert is also instructed to direct the registrant to provide a
+specification ({{Section 4.6 of -ianacons}}), but can make exceptions,
+for instance when a specification is not available at the time of
+registration but is likely forthcoming.
```
glyn, line 1678: This may be common policy for IANA, but the risk is that a spec may never appear and we end up with people "squatting" on function names so others can't register them.

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
 Provisional registration? (standards tree only):
 : no
 
+## Function Extensions {#iana-fnex}
+
+This specification defines a new "Function Extensions sub-registry" in
+a new "JSONPath Parameters registry", with the policy "expert review"
+({{Section 4.5 of -ianacons}}).
+
+The expert is instructed to be frugal in the allocation of very short
```
glyn, line 1672: Is there just one expert? Why not:
```suggestion
Experts are instructed to be frugal in the allocation of very short
```

cabo:
```suggestion
The experts is instructed to be frugal in the allocation of very short
```

I'd say.

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
+: (see {{Section 2.3 of -ianacons}})
+
+Reference:
+: a reference document that provides a description of the function
+  extension
+
+Initial entries in this sub-registry are as listed in {{pre-reg}}:
+
+| Function Name | Brief description | Input | Output | Change Controller | Reference |   |   |
+| length        | length of array   | value | value  | IESG              | {{fnex}} of RFCthis |   |   |
```
glyn, line 1712: Do we need to allow an empty nodelist as input so that we can apply `length` to a `singular-path`?

glyn (commented):
An excellent start, thanks! Comments provided.

draft-ietf-jsonpath-base.md
glyn: Of course. I was thinking "list of nodes" rather than _nodelist_.

glyn: If we can avoid the function having to create nodes, so much the better. So I guess that limits return values to nodelists or values after all.

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+nodelist MUST return a singular node to be used in a comparable or a
+regex-expr.
+
+###### Function Extensions: length {#length}
+{: unnumbered}
+
+When defining an extension point in a protocol, it is generally a good
+idea to provide at least one extension that makes use of the extension
+point, to serve as an example and to provide motivation to implement
+the extension point.
```
gregsdennis, line 1233: I don't know that this is need in the spec.  Might be good for a comment, but its target audience is really people who write specs.

draft-ietf-jsonpath-base.md
```diff
@@ -1176,6 +1181,72 @@ The semantics of regular expressions are as defined in {{-iregexp}}.
...
+of an array and make that available for further processing in the
+filter expression:
+
+~~~ JSONPath
+$[?length(@.authors) >= 5]
+~~~
+
+Its only argument is a value (possibly taken from a singular path as
+in the example above).  The result also is a value; if the argument
+value is an array, the result is the number of elements in the array.
```
gregsdennis, line 1245: Do we also want to support strings or objects?  They're fairly well-defined for this function.

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
 : no
 
+## Function Extensions {#iana-fnex}
+
+This specification defines a new "Function Extensions sub-registry" in
+a new "JSONPath Parameters registry", with the policy "expert review"
+({{Section 4.5 of -ianacons}}).
+
+The experts are instructed to be frugal in the allocation of very short
+function extension names, keeping them in reserve for applications that
```
gregsdennis, line 1673: What do you mean by "applications" here?

draft-ietf-jsonpath-base.md
gregsdennis: Perhaps the function name reservation can have an expiry.

draft-ietf-jsonpath-base.md
```diff
@@ -1592,6 +1663,54 @@ Change controller:
...
+they deem such a registration can avert potential future collisions.
+{: #de-instructions}
+
+Each entry in the registry must include:
+
+{:vspace}
+Function Name:
+: a lower case ASCII {{-ascii}} string that starts with a letter and can
+  contain letters, digits and underscore characters afterwards
+  (`[a-z][_a-z0-9]*`).
```
gregsdennis, line 1690: So we're only accepting English-character function names?  Just clarifying.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
cabo: (Maybe I should have added a "to me removed by RFC editor" comment.)

draft-ietf-jsonpath-base.md
cabo: Very good question.  What do implementations out there do?
For strings, I think we'd count the Unicode scalars.
For JSON objects, the number of name-value pairs.

draft-ietf-jsonpath-base.md
cabo: application in the sense of the way this is going to the used, usage

draft-ietf-jsonpath-base.md
cabo: Well, ASCII, yes.
This is an interesting question in a more general way, as so far JSONPath is mostly indifferent to languages.
But these function extensions are intended for an international audience, so it is probably best to do a lowest common denominator.

cabo:
Weirdly, I cannot reply to this observation: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/280#discussion_r1008936675

Expiry is somewhat unusual for IANA registries (early allocations do expire).
What is the behavior you would expect implementations to exhibit?
The function extension to stop working on the expiry day?

draft-ietf-jsonpath-base.md
goessner: Shouldn't the function definition being able to define the "return" even more concretely as a string or number? The latter would be, what is expected with our `length` function.

draft-ietf-jsonpath-base.md
goessner: We even could allow nodelists and return the number of nodes.

draft-ietf-jsonpath-base.md
goessner: I like the well known return value `-1` for the case of an invalid argument.

draft-ietf-jsonpath-base.md
cabo: That would need to be a separate extension, as length(X) is ambiguous if X can be interpreted as a node that contains an array (which gets us the size of the array) or a nodelist of size 1 (which gets us 1).

draft-ietf-jsonpath-base.md
glyn: Also, what happens when a function is passed an argument of the wrong kind? Does this produce a syntax error? Is there any coercion to the correct kind? Are there any "soft" runtime errors in which case what does the function return?

draft-ietf-jsonpath-base.md
gregsdennis: I wonder if another word would work better here so that readers don't conflate it with executable applications or other consuming code.

glyn: Perhaps:
```suggestion
function extension names, keeping them in reserve for functions that
```

draft-ietf-jsonpath-base.md
cabo: Obviously, I stole the ABNF from dot-member-name, which allows %x80-10FFFF -- this is a recipe for confusion, with RTL characters, format effectors, and all that.  I'm having a hard time making a suggestion as to what we should allow there -- `\p{L}|\p{N}` maybe, in parallel to allowing ALPHA + DIGIT?  But then we also allow "_", and there may be equivalent separators in some scripts.  So this is hard to get useful over a wide domain, while reducing the potential for confusion.  function names of course do not depend on whatever language was used in the input data item, so we can be more frugal in what we allow.

timbray:
Hey @cabo, based on this useful-looking discussion, do you plan to revise this significantly?  I'm wondering if I should give it a careful read now or wait for an updated version.

cabo:
Definitely.  Of course, I'd like to have your input on the points raised (the non-trivial ones).  We might even want to have a design-team meeting...

timbray:
> Definitely. Of course, I'd like to have your input on the points raised (the non-trivial ones). We might even want to have a design-team meeting...

Might be a good idea to mark the non-non-trivial issues above resolved, to help reviewer focus. If it's OK, I'll wait for the next revision to go end-to-end. Agreed on a design meeting.

draft-ietf-jsonpath-base.md
cabo: Indeed, this is really less about shortness.  Say, someone registers a size() that extracts the shirt size of a JSON description of a shirt.

draft-ietf-jsonpath-base.md
cabo: Finally, I can reply to this observation.
Copying from below:

Expiry is somewhat unusual for IANA registries (early allocations do expire).
What is the behavior you would expect implementations to exhibit?
The function extension to stop working on the expiry day?

cabo:
Here's my list of items where we cold use input:

- More frugal with allowing whitespace?
- return value of functions that don't get the right arguments (e.g., length(1))
- (interaction with #195 -- non-singular paths in exist-expr)
- IANA designated expert instructions (frugal with "good" names)
- expiring reservations?  (shades of certificates)
- domain of length (arrays now, add strings, objects?)?
- should there be another function for the size of nodelists (size())?

Probably mostly resolved:
- `[a-z][_a-z0-9]*` for function names (ASCII, lowercase, _ only)

draft-ietf-jsonpath-base.md
goessner: I don't think, that #195 is relevant here, since `exist-expr` always results in a Boolean, regardless of the presence of a singular-path or not.

glyn: My point was about an `exist-expr` which tests the existence of the results of a function. Unless/until #195 is merged, a function used in such a context MUST return a singular node.

draft-ietf-jsonpath-base.md
cabo: Well, the answer is yes; we should move any attempt to reduce variability in spacing to its own PR.

draft-ietf-jsonpath-base.md
cabo: Removed in f7295bd

draft-ietf-jsonpath-base.md
cabo: No further discussion -- I'm assuming we stick with `[a-z][_a-z0-9]*` (no change required)

draft-ietf-jsonpath-base.md
glyn: Actually, why not simply the following?
```suggestion
are likely to enjoy wide use.
```
You see I'm not sure the phrase "and can make good use of their shortness" is anything more than motherhood.

draft-ietf-jsonpath-base.md
glyn: @gregsdennis I think those questions are for you. (I wasn't suggesting expiry.)

draft-ietf-jsonpath-base.md
cabo: Addressed in 85e5525

draft-ietf-jsonpath-base.md
cabo: I think the DE instructions are open enough so that the DE can be quite insisting on a specification, so I don't see such a potential for disaster here.

draft-ietf-jsonpath-base.md
cabo: That should be a separate function (for a singular node list, it is not clear whether you want 1 or the length of the value in the node).

cabo: ABNF says:

```
function-argument       = comparable
```

So literals are the other case.

draft-ietf-jsonpath-base.md
cabo: That is certainly a part that we still need to address.
Functions that return a nodelist probably need to be categorized into singular ones and ones that are not.
Let's to that when we agree on the overall architecture of the extension.

draft-ietf-jsonpath-base.md
cabo: Let's revisit this with the complete proposal

draft-ietf-jsonpath-base.md
cabo: I made a different proposal specifically for length(), but we do need to come up with a more general solution.

draft-ietf-jsonpath-base.md
cabo: Let's see which mistakes can be caught at compile time, and whether we can stick with "no runtime errors".

draft-ietf-jsonpath-base.md
cabo: Separate function now in my proposal update.

draft-ietf-jsonpath-base.md
cabo: Separate function (called "count") provided now.

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+~~~ JSONPath
+$[?count(@.*.author) >= 5]
+~~~
+
+Its only argument is a nodelist.
+The result is a value, an unsigned integer, that gives the number of
+nodes in the nodelist.
+Note that there is no deduplication of the nodelist. [^dedup]
+
+[^dedup]: Well, that can be discussed.
```
glyn, line 1294: Perhaps add a remdup/dedup function?

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+in the example above).  The result also is a value, an unsigned
+integer.
+
+* If the argument value is a string, the result is the number of
+  Unicode scalar values in the string.
+* If the argument value is an array, the result is the number of
+  elements in the array.
+* If the argument value is an object, the result is the number of
+  members in the object.
+* For any other argument value, the result is one.
```
glyn, line 1269: Does that include the case of a singular path which returns an empty nodelist?

draft-ietf-jsonpath-base.md
cabo: Maybe.  Maybe that is a bit too much invention for the base standard.  I myself don't have an opinion on this.

draft-ietf-jsonpath-base.md
cabo: That is a very good question.  As defined, the argument value needs to be a value, so handling empty nodelists is not defined.  But it should be possible to do this...

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+
+
+###### Function Extensions: count {#count}
```
gregsdennis, line 1272: What's the difference between `count()` and `length()`?  Why have a distinction?

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+
+Its only argument is a nodelist.
+The result is a value, an unsigned integer, that gives the number of
+nodes in the nodelist.
+Note that there is no deduplication of the nodelist. [^dedup]
+
+[^dedup]: Well, that can be discussed.
+
+
+###### Function Extensions: match {#match}
```
gregsdennis, line 1297: Since we have `match()` and `search()` now, we should probably remove `=~`.

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+~~~
+
+Its first argument is a string that is searched for at least one
+substring that matches against the iregexp contained in the string
+that is the second argument.
+The result is `true` if such a substring exists, `false` otherwise.
+
+o o o o o o o o o o o o o o o o o o o o o o o o o o o o o o o o
 
 #### Examples
```
gregsdennis, line 1345: Need examples of `match()` and `search()` to better highlight the difference between them.  The text varies only by one saying "contains a match" vs just "matches," which I don't think is clear.  In my experience with regexes, containing a match _is_ matching.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
```diff
@@ -1195,19 +1188,159 @@ compared is an object, array, boolean, or `null`.
...
+
+Result:
+: value (unsigned integer)
+
+The "count" function extension provides a way to obtain the number of
+nodes in a nodelist and make that available for further processing in
+the filter expression:
+
+~~~ JSONPath
+$[?count(@.*.author) >= 5]
```
glyn, line 1286: This seems syntactically invalid since any function argument has to be a `comparable` which `@.*.author` isn't.

draft-ietf-jsonpath-base.md
glyn: It has mostly gone. There are just two vestigial mentions left.

cabo: Fixed in 52ca26b

draft-ietf-jsonpath-base.md
cabo: Right.  The one example that has a match() would work the same with a search()... we probably need to set up a pair of contrasting examples.  Let's do a round through the examples once the technical content has stabilized.

draft-ietf-jsonpath-base.md
cabo: Yes.  This doesn't have the results of merging #195 incorporated.

draft-ietf-jsonpath-base.md
cabo:
``` 
###### Function Extensions: length {#length}
{: unnumbered}

Arguments:
: 1. value

###### Function Extensions: count {#count}
{: unnumbered}

Arguments:
: 1. nodelist
```

One is for JSON values, one is for nodelists.

draft-ietf-jsonpath-base.md
cabo: The text is fine though.  87d95d6 adds a few words that might help grasping the difference.

draft-ietf-jsonpath-base.md
gregsdennis: This isn't apparent to me from the text.  Some additional clarification would be good. (May be provided with the examples I asked for in another comment.)

draft-ietf-jsonpath-base.md
```diff
@@ -1402,7 +1403,7 @@ Queries:
 | `$[?@[?@.b]]` | `[3, 5, 1, 2, 4, 6, {"b": "j"}, {"b": "k"}, {"b": {}}]` | `$['a']` | Nested filters |
 | `$.a[?@.b, ?@.b]` | `{"b": "j"}` <br> `{"b": "k"}` <br> `{"b": "k"}` <br> `{"b": "j"}` | `$['a'][6]` <br> `$['a'][7]` <br> `$['a'][7]` <br> `$['a'][6]` | Non-deterministic ordering |
 | `$.a[?@<2 || @.b == "k"]` | `1` <br> `{"b": "k"}` | `$['a'][2]` <br> `$['a'][7]` | Array value logical OR |
-| `$.a[?@.b =~ "[jk]"]` | `{"b": "j"}` <br> `{"b": "k"}` | `$['a'][6]` <br> `$['a'][7]` | Array value regular expression |
+| `$.a[?match(@.b, "[jk]")]` | `{"b": "j"}` <br> `{"b": "k"}` | `$['a'][6]` <br> `$['a'][7]` | Array value regular expression |
```
gregsdennis, line 1406: Can we get one with `search()` as well, please?

glyn:
@cabo: I rebased this and resolved some conflicts.

draft-ietf-jsonpath-base.md
cabo: grammar fixed in 08dd7ad
now add semantic restriction

cabo:
Update:

> Here's my list of items where we could use input:
> 
> * More frugal with allowing whitespace?

Address whitespace in a separate PR, if desired.

> * return value of functions that don't get the right arguments (e.g., length(1))

➔ Semantic issue: What to do with value of wrong kind?

> * (interaction with [Do we really require a singular path for `exist-expr` #195](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/195) -- non-singular paths in exist-expr)

Fixed in 08dd7ad

> * IANA designated expert instructions (frugal with "good" names)

Fixed in afda44a

> * expiring reservations?  (shades of certificates)

Probably not.

> * domain of length (arrays now, add strings, objects?)?

➔ Semantic issue: What to do with value of wrong kind?

> * should there be another function for the size of nodelists (size())?

Yes, that is called count() now (e3a565a)

> Probably mostly resolved:
> 
> * `[a-z][_a-z0-9]*` for function names (ASCII, lowercase, _ only)

Let's stick with that name space.

draft-ietf-jsonpath-base.md
```diff
@@ -1207,19 +1200,161 @@ compared is an object, array, boolean, or `null`.
...
+                             *(S "," S function-argument)] S ")"
+singular-path           =/ function-expression ; if value
+filter-path             =/ function-expression ; if nodelist
+function-argument       = comparable / filter-path
+~~~
+
+Syntactically, a function-expression can occur anywhere where a
+singular-path can occur (exist-expr, comparable).
+A function-expression that employs a function extension that returns a
+nodelist MUST return a singular node to be used in a comparable.
```
glyn, line 1253: Why would it not be acceptable to return an empty nodelist since that's valid for use in a comparable?

glyn: Separately: what should the behaviour be if a nodelist with multiple nodes is returned in this situation?

glyn:
> > * return value of functions that don't get the right arguments (e.g., length(1))
> 
> ➔ Semantic issue: What to do with value of wrong kind?

We may be able to make this syntactically invalid. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/320 for some ideas.

draft-ietf-jsonpath-base.md
cabo: One way to handle this would be to build a type system with values, singular paths (which are implicitly converted into values where a value is required), and filter-paths.
Functions already have to define the types of their arguments and their result, so this could be checked statically.

glyn: Yes, I like that. The more static checks we can get, the fewer runtime errors the spec will need to cope with.

gregsdennis (approved):

PR #281, opened 2022-11-01
Address James Gould's review points

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -746,7 +744,7 @@ since JSON objects are unordered.
 Children of an array appear in array order in the resultant nodelist.
 
 The `wildcard` selector selects nothing from a primitive JSON value (that is,
-a number, a string, `true`, `false`, or `null`).
+a number, a string, a boolean, or `null`).
```
gregsdennis, line 747: I'm not sure what the comment was, but JSON defines true/false as explicit values.

It may also help us distinguish between a JSON boolean and an expression result boolean, which we've already decided are incompatible.

draft-ietf-jsonpath-base.md
```diff
@@ -1047,7 +1045,7 @@ relation-expr = comp-expr /                           ; comparison test
                 regex-expr                            ; regular expression test
 ~~~~
 
-Comparisons are restricted to Singular Path values and primitive values (that is, numbers, strings, `true`, `false`,
+Comparisons are restricted to Singular Path values and primitive values (that is, numbers, strings, booleans,
```
gregsdennis, line 1048: Same here about the booleans.

Also, do we not want to include `null` for some reason?

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
glyn: The comment was a suggestion to use boolean instead of `true` and `false`. The JSON spec defines a boolean type, so this makes sense. I don't think avoiding the term "boolean" here really disambiguates the other occurrences of "boolean" since this spec should be read in the context of the JSON spec. Best if we are explicit about boolean JSON values and boolean expressions.

draft-ietf-jsonpath-base.md
glyn: `null` is mentioned on the next line.

cabo: The JSON spec mentions "booleans" [sic!] only in the introduction.  The actual spec uses `false` and `true` (e.g., Section 3).  I slightly prefer to follow this precedent, as there is nothing in JSON that makes two out of the three special values a "type" (JSON actually doesn't have types -- that word is mentioned again in the intro only, and in Section 5 where the word is used to say that JSON arrays do not use the concept).

glyn: Ok, I'll revert those two changes.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: Based on @cabo's comment below, I am persuaded we should stay with `true` and `false`.

draft-ietf-jsonpath-base.md
```diff
@@ -800,7 +798,7 @@ Notes:
 {: unnumbered}
 
 The `index-selector` applied to an array selects an array element using a zero-based index.
-For example, the selector `0` selects the first and the selector `4` the fifth element of a sufficiently long array.
+For example, the selector `0` selects the first and the selector `4` selects the fifth element of a sufficiently long array.
```
timbray, line 801: I like the original version better but meh

timbray (approved):

PR #283, opened 2022-11-01
Fixes to review comments from Michael Brooks

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -272,8 +272,8 @@ A query is applied to an argument, and the output is a nodelist.
 ## History
 
 This document picks up {{{Stefan Gössner}}}'s popular JSONPath proposal
-dated 2007-02-21 {{JSONPath-orig}} and provides a normative definition
-for it.
+dated 2007-02-21 {{JSONPath-orig}}, addresses some of its shortfalls, and
```
cabo, line 275: I'm not so hot on calling evolving this proposal "addressing some of its shortfalls".
Much of the evolution is about how the environment has moved over the last 15 years, not foreseeing which is not really a "shortfall" of the 2007 proposal...

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: The draft fixes, for example,  an issue called out by the original article: "Currently only single quotes allowed inside of JSONPath expressions". Does "shortfall" sound too critical? If so, we could say instead:
```suggestion
dated 2007-02-21 {{JSONPath-orig}}, tidies it up, and
```

cabo: how about:

```suggestion
dated 2007-02-21 {{JSONPath-orig}}, addresses some of the experience from the wide deployment of its implementations, and
```

glyn: I could live with that, but I'd prefer to "build on experience" rather than "address" it.

glyn: See bc8249185f20453c0be32b2494ebea9e2f1e105b.

goessner: JSONPath was never meant to be restricted to single quotes only. Therefore the issue. Reason was very probable the use of lean and simple regular expressions.

PR #284, opened 2022-11-02
Improve abstract/intro wording

glyn:
This is an attempt to make two sentences in the abstract and introduction less clunky while not jettisoning their precision.


cabo (approved):
Works for me, although one would need to understand that JSON values are nested to understand the abstract.  (But we can't start from the creation of the universe.)

glyn:
Thanks @cabo. Approve away!

Issue #285, opened 2022-11-03
consensus on the term "current node"


goessner:
The term "current node" is traditionally used with iterating over children of structured nodes in filter selectors.

Now it also appears in 3.3 Semantics chapter in a different context:

```
A segment performs its function on each of the nodes in its input
nodelist, during such a function execution, such a node is referred
to as the "current node".
```

To avoid confusion of readers, we might use a different term there. A good alternative would be "context node", which is also used by XPath in a similar context.

glyn:
Perhaps a simple solution would be to avoid the use of the term in 3.3 as this term doesn't appear to be used elsewhere.

glyn:
Please see the simplification in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/286.

PR #286, opened 2022-11-03
Delete unused and conflicting term

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/285

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -515,8 +515,8 @@ representation as desired by the implementation.
 Note that an empty nodelist is a valid query result.
 
 A segment performs its function on each of the nodes in its input
-nodelist. During such a function execution, each such node is referred to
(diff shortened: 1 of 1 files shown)


cabo (approved):
Note that there is still an instance of "current node identifier" in table 1.

I also note that then English usage of "function" now collides with the term we coin in the function extensions.

But the basic change here is right, so I approve.

glyn:
> Note that there is still an instance of "current node identifier" in table 1.

I left that in deliberately. We have removed the colliding term and we are now free to use current node (`@`) throughout. I will follow up with a PR which does just that.

> 
> I also note that [the] English usage of "function" now collides with the term we coin in the function extensions.

I objected to the term "function" in this context previously. I'll follow up with a PR to delete it, which will leave it free for function extensions.

> 
> But the basic change here is right, so I approve.

PR #287, opened 2022-11-03
Use current node consistently

glyn:


glyn:
There may be some more shake-out as filter semantics are scattered through the syntax section. We may need to live with this because it makes more sense to talk about semantics in the context of each particular piece of syntax. But let's use the current PR just to use the current node term consistently.

draft-ietf-jsonpath-base.md
```diff
@@ -1026,12 +1026,15 @@ exist-expr        = [logical-not-op S] singular-path  ; path existence or non-ex
...
 ~~~~ abnf
-singular-path     = rel-singular-path / abs-singular-path
-rel-singular-path = "@" *(S (name-segment / index-segment))
-abs-singular-path = root-identifier *(S (name-segment / index-segment))
-name-segment      = "[" name-selector "]" / dot-member-name-shorthand
-index-segment     = "[" index-selector "]"
+singular-path           = rel-singular-path / abs-singular-path
+rel-singular-path       = current-node-identifier *(S (name-segment / index-segment))
+current-node-identifier = "@"
+abs-singular-path       = root-identifier *(S (name-segment / index-segment))
```
cabo, line 1035: I would probably factor `*(S (name-segment / index-segment))` into a new rule `singular-path-segments`.  Not very strongly related to this PR though.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: I like it. Done.

draft-ietf-jsonpath-base.md
```diff
@@ -998,15 +998,15 @@ Queries:
...
+The filter selector has the form `?<expr>`. It iterates over structured values, i.e., arrays and objects.
 
 ~~~~ abnf
 filter-selector = "?" S boolean-expr
 ~~~~
 
-During the iteration process each array element or object member is visited and its value — accessible via symbol `@` — or one of its descendants — uniquely defined by a relative path — is tested against a boolean expression `boolean-expr`.
-
-The current item is selected if and only if the boolean expression yields true.
+During the iteration process the node of each array element or object member visited is known as the current node.
```
cabo, line 1007:
```suggestion
During the iteration process the node of each array element or object member being visited is known as the current node.
```

(this doesn't "stick", but is true for the current node.)

cabo (approved):

PR #288, opened 2022-11-03
Free up the term function

glyn:
We expect to use this term in the description of extensions.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -514,7 +514,7 @@ Normalized Paths referencing the nodes, or both — or some other
 representation as desired by the implementation.
 Note that an empty nodelist is a valid query result.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -514,10 +514,8 @@ Normalized Paths referencing the nodes, or both — or some other
 representation as desired by the implementation.
 Note that an empty nodelist is a valid query result.
 
-A segment performs its function on each of the nodes in its input
-nodelist.
-Each of these function executions produces a
-nodelist, which are then concatenated to produce
+A segment operates on each of the nodes in its input nodelist in turn
```
cabo, line 517:
```suggestion
A segment operates on each of the nodes in its input nodelist in turn,
```

cabo (approved):

PR #289, opened 2022-11-03
Motivate Normalized Paths

glyn:


goessner:
I welcome this ... thanks.

draft-ietf-jsonpath-base.md
```diff
@@ -1449,13 +1449,23 @@ Queries:
...
+Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+For example, the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`.
+
+Normalized Paths are used in this document to clarify the contents of nodelists in examples.
+
+Since output nodelists may be input to subsequent processing (e.g., removal of duplicate nodes), it is essential to provide a standard
+string representation of their nodes.
+
+For a given JSON value, there is a one to one correspondence between the value's
```
goessner, line 1465: Should be 
```
For a given JSON value, there is a one to one correspondence between the value's
node and the Normalized Path that identify this node.
```

draft-ietf-jsonpath-base.md
```diff
@@ -1449,13 +1449,23 @@ Queries:
...
 nodes and the Normalized Paths that identify these nodes.
 Note that there is precisely one Normalized Path that identifies each node.
-
-A JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+Consequently, Normalized Paths may be used to provide more predictable behavior and simplify testing.
```
goessner, line 1468: perhaps skip "more".

draft-ietf-jsonpath-base.md
timbray: one-to-one with hyphens?

draft-ietf-jsonpath-base.md
```diff
@@ -1449,13 +1449,23 @@ Queries:
...
+for example, `$['book'][3]`.
+
+A Normalized Path provides a concrete string representation of a node in a nodelist.
+Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+For example, the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`.
+
+Normalized Paths are used in this document to clarify the contents of nodelists in examples.
+
+Since output nodelists may be input to subsequent processing (e.g., removal of duplicate nodes), it is essential to provide a standard
```
timbray, line 1462: s/essential/useful/ ? or helpful?

draft-ietf-jsonpath-base.md
glyn: > Should be
> 
> ```
> For a given JSON value, there is a one to one correspondence between the value's
> node and the Normalized Path that identifies this node.
> ```

That wasn't what I was trying to say. The idea is that all the nodes in the value, including descendants, are in a one-to-one correspondence with their Normalized Paths. Please see 1538578f1cd9df6de27c9eae9d2da09f6cb5dee9 for a rewording.

draft-ietf-jsonpath-base.md
glyn: Adopted.

draft-ietf-jsonpath-base.md
glyn: Reworded using "useful" in 329940c811e260f0fd86dfd21339a6e6d0beb54e.

draft-ietf-jsonpath-base.md
goessner: Hmm ... if it is true for any single value and its associated normalized Path, it is consequently true for all values.

So the singular formulation above is sufficient IMHO.

draft-ietf-jsonpath-base.md
```diff
@@ -1474,13 +1474,23 @@ Queries:
...
-Note that there is precisely one Normalized Path that identifies each node.
+A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node,
+for example, `$['book'][3]`.
+
+A Normalized Path provides a concrete string representation of a node in a nodelist.
+Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+For example, the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
+`$['book'][3]` and `$['book'][5]`.
+
+Normalized Paths are used in this document to clarify the contents of nodelists in examples.
```
goessner, line 1485: perhaps "show" instead of "clarify" ... ?

draft-ietf-jsonpath-base.md
glyn: A one-to-one correspondence, also known as a bijection, is only really useful when there is more than one item in the domain and range of the correspondence. There is always a one-to-one correspondence between one item and another taken in isolation. So the singular formulation would essentially say nothing.

This is a little subtle because the one-to-one correspondence is a consequence of the way we have defined Normalized Paths, but I want to point it out since it's a very useful property.

To see the general point, let's step away from Normalized Paths briefly to a fictional "label" of a node and look at an example value `[0]` with an array node and a number node. Taking these nodes individually, we could say that `"x"` is the label of the array node and that `"x"` is also the label of the number node. Then:
* There is a one-to-one correspondence between the array node and the label `"x"`.
* There is a one-to-one correspondence between the number node and the label `"x"`.
* But there isn't a one-to-one correspondence between the set of nodes and their labels, since both nodes have the same label.

draft-ietf-jsonpath-base.md
glyn: > perhaps "show the resulting" instead of "clarify the contents of" ... ?

Yes. Adopted in 761f2608cc30ef5c1a3a59c9f091144b09e69ac8.

timbray: Also consider "represent"

glyn: I considered "represent", but "show" is more concrete, and is perhaps a brief respite for the reader from one-to-one correspondences and other abstract language. :-)

sthagen:
From the side line: Do we need to add description of behavior (with a `may` qualifier) to motivate the truly welcome attributes of the `Normalized Path`?

draft-ietf-jsonpath-base.md
```diff
@@ -1474,13 +1474,23 @@ Queries:
...
 
-A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node. For example,
-the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
-`$['book'][3]` and `$['book'][5]`. For a given JSON value, there is a one to one correspondence between the value's
-nodes and the Normalized Paths that identify these nodes.
-Note that there is precisely one Normalized Path that identifies each node.
+A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node,
+for example, `$['book'][3]`.
+
+A Normalized Path provides a concrete string representation of a node in a nodelist.
```
cabo, line 1480: An address is not a representation.

glyn:
> From the side line: Do we need to add description of behavior (with a `may` qualifier) to motivate the truly welcome attributes of the `Normalized Path`?

A Normalized Path is just a particular kind of JSONPath query. Its only behaviour is to identify a node within the tree of nodes in a value. Or did you have some other behaviour in mind?

draft-ietf-jsonpath-base.md
cabo: We may want to be explicit that these paths are a **convention** for expressing **node identify**.
(This is part of the specification because identifying a "conventional" convention is good, as the text below says.)

timbray:
> A Normalized Path is just a particular kind of JSONPath query

Oh, of course, but that formulation had never occurred to me, and it's useful.  If it's not in the spec, it should be.

sthagen:
> > From the side line: Do we need to add description of behavior (with a `may` qualifier) to motivate the truly welcome attributes of the `Normalized Path`?
> 
> A Normalized Path is just a particular kind of JSONPath query. Its only behaviour is to identify a node within the tree of nodes in a value. Or did you have some other behaviour in mind?

I meant the first of the two `Consequently, ...` sentences:
> Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.

Here we describe the behavior of an implementation (output as response to an unspecified request that triggers the former).

I just wonder if this section is a good place to describe a part of a protocol, when the focus is a) the "projection" itself and b) potential benefits (all from a data perspective).

cabo:
> Here we describe the behavior of an implementation (output as response to an unspecified request that triggers the former).

Well, we describe the nature of what it is to implement JSONPath.
Generally, in a protocol, you not only need to prescribe what is on the wire, but also what it means.
That is only ever in the implementation.  That doesn't mean that it describes the behavior of an implementation.

E.g., in TCP, the byte stream that is transported by the protocol only ever happens in the implementation.  Nothing on the wire reflects this byte stream.  But describing that byte stream as the underlying model of TCP is rightly not criticized as "describing the behavior of an implementation".

glyn:
> From the side line: Do we need to add description of behavior (with a `may` qualifier) to motivate the truly welcome attributes of the `Normalized Path`?

The penny has just dropped: you are questioning the need for such a description rather than asking for more of the same! I agree with @cabo's response.

draft-ietf-jsonpath-base.md
glyn: Hopefully this is now covered.

draft-ietf-jsonpath-base.md
goessner: thanks for enlightening me ...

glyn: You're welcome. However, given the complexity of the one-to-one correspondence language, I took the opportunity to simplify things by talking about a Normalized Path being an identity of a node. This implies a one-to-one correspondence, but I think it's easier for readers to understand.

draft-ietf-jsonpath-base.md
```diff
@@ -1474,23 +1474,29 @@ Queries:
...
+need escaping when Normalized Paths appear in double quote delimited strings, e.g., in JSON texts.
 
 Certain characters are escaped, in one and only one way; all other characters are unescaped.
 
-Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-3]`, for example, is a Singular
-Path, but not a Normalized Path.
-The Normalized Path equivalent to `$[-3]` would have an index equal to the array length minus `3`.
+Normalized Paths are Singular Paths, but not all Singular Paths are Normalized Paths.
+For example, `$[-3]` is a Singular Path, but is not a Normalized Path.
+The Normalized Path equivalent of `$[-3]` would have an index equal to the array length minus `3`.
```
cabo, line 1499: e·quiv·a·lent | əˈkwɪv(ə)lənt |
adjective
equal in value, amount, function, meaning, etc.: one unit is equivalent to one glass of wine.
• [predicative] (equivalent to) having the same or a similar effect as: some regulations are equivalent to censorship.

Note the preference for "equivalent to".
(And it's the same in British English.)

draft-ietf-jsonpath-base.md
```diff
@@ -1474,23 +1474,29 @@ Queries:
...
 Single quotes are used to delimit string member names. This reduces the number of characters that
-need escaping when Normalized Paths appear as strings (which are delimited with double quotes) in JSON texts.
+need escaping when Normalized Paths appear in double quote delimited strings, e.g., in JSON texts.
 
 Certain characters are escaped, in one and only one way; all other characters are unescaped.
 
-Normalized Paths are Singular Paths. Not all Singular Paths are Normalized Paths: `$[-3]`, for example, is a Singular
-Path, but not a Normalized Path.
-The Normalized Path equivalent to `$[-3]` would have an index equal to the array length minus `3`.
+Normalized Paths are Singular Paths, but not all Singular Paths are Normalized Paths.
```
cabo, line 1497: This probably should be in a note.

draft-ietf-jsonpath-base.md
```diff
@@ -1474,23 +1474,29 @@ Queries:
...
-nodes and the Normalized Paths that identify these nodes.
-Note that there is precisely one Normalized Path that identifies each node.
+A Normalized Path is a conventional representation of the identity of a node in a value.
+Specifically, a Normalized Path is a JSONPath query with restricted syntax (defined below),
+e.g., `$['book'][3]`, which when applied to the value results in a nodelist consisting
+of just the node identified by the Normalized Path.
+Note that a Normalized Path represents the identity of a node _in a specific value_.
 
-A JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
+Normalized Paths provide a predictable format that simplifies testing and post-processing
```
cabo, line 1483: This is a sentence about usage; at this point it isn't quite clear that the normalized path of a node is deterministic, i.e., for the same node in the same value, two implementations will generate the exact same normalized path.

cabo (commented):
Definitely the right direction.  Thank you!

goessner (approved):

draft-ietf-jsonpath-base.md
gregsdennis: I could also argue "equivalent for" here.  I think this is just a regional preference.  They all sound fine to me.

gregsdennis:
I think an important aspect that this is missing is that normalized paths are contexted to the JSON data that was queried; whereas singular paths exist independently, as does any other path.

For example, the path `$[-3]` (used in the text) is singular and can rightfully exist on its own.  However it can't be normalized without querying JSON data.  Once queried, the resulting normalized paths can be different based on the data.

- `[ 0, 1, 2, 3, 4 ]` yields a normalized path of `$[2]` for the value `2`
- `[ 0, 1, 2, 3, 4, 5, 6 ]` yields a normalized path of `$[4]` for the value `4`.

gregsdennis:
> > Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
> 
> Here we describe the behavior of an implementation (output as response to an unspecified request that triggers the former).

This text isn't getting into implementation behavior at all. It's specifying output and allowing for options.

sthagen:
> This text isn't getting into implementation behavior at all. It's specifying output and allowing for options.

Well, unforced ambiguous behavior then. Specifying output and allowing for (unrequested) options.

For instance when the SomeTool processor outputs a number 42 and the options are that it represents any base in which the symbols are generally valid (per consensus) to me would rather raise questions than answer them.
I would expect these complication (extra dimensions)  less inside a format but more in a "processor" role or protocol description. The latter describing expected behavior within the dynamics of some interchange.

I used the term `implementation` without caution. Sorry for that.
In my reading of the sentence we discuss about, the consumer of such a result has to look elsewhere to find out what the processor provides as result, right?

That is our normal task as tool users.

However, I think we better try to not bring in such ambiguities without noting the reason outside of our mandate to change. Maybe note, that existing processors (or however we like to call these providers of jsonpath responses) display mixed behaviors (what you call options).

Makes sense?

If not, sorry for the noise.

cabo:
> I think an important aspect that this is missing is that normalized paths are contexted to the JSON data that was queried; whereas singular paths exist independently, as does any other path.

Well, we do say:

```
Note that a Normalized Path represents the identity of a node _in a specific value_.
```

I wouldn't go so far to say that the Normalized Path is "contexted".  You need the value to compute the Normalized Path from the query, but after that, it doesn't require context to apply it (well, they are meant to be applied to that value, but they are regular JSONPath queries).

With IETF 115 raging and Glyn being offline for a couple of days, we'll probably not respond very fast the next few days; I think good points are being made here even it is not always clear how to modify the text further to fully reflect them.

draft-ietf-jsonpath-base.md
```diff
@@ -1474,13 +1474,23 @@ Queries:
...
-A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node. For example,
-the JSONPath expression `$.book[?(@.price<10)]` could select two values with Normalized Paths
-`$['book'][3]` and `$['book'][5]`. For a given JSON value, there is a one to one correspondence between the value's
-nodes and the Normalized Paths that identify these nodes.
-Note that there is precisely one Normalized Path that identifies each node.
+A Normalized Path is a JSONPath with restricted syntax that identifies a node by providing a query that results in exactly that node,
+for example, `$['book'][3]`.
+
+A Normalized Path provides a concrete string representation of a node in a nodelist.
+Consequently, a JSONPath implementation may output Normalized Paths instead of, or in addition to, the values identified by these paths.
```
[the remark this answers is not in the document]
glyn: I have removed the word "consequently".

draft-ietf-jsonpath-base.md
glyn: Adopted "equivalent to".

draft-ietf-jsonpath-base.md
glyn: Done.

draft-ietf-jsonpath-base.md
glyn: Added a sentence at the end of the first paragraph of this section to make that clear.

glyn:
> I think an important aspect that this is missing is that normalized paths are contexted to the JSON data that was queried; whereas singular paths exist independently, as does any other path.
> 
> For example, the path `$[-3]` (used in the text) is singular and can rightfully exist on its own. However it can't be normalized without querying JSON data. Once queried, the resulting normalized paths can be different based on the data.
> 
>     * `[ 0, 1, 2, 3, 4 ]` yields a normalized path of `$[2]` for the value `2`
> 
>     * `[ 0, 1, 2, 3, 4, 5, 6 ]` yields a normalized path of `$[4]` for the value `4`.

I believe the PR now makes this property of Normalized Paths clear. (I'd prefer not to labour the example.)

glyn:
> > This text isn't getting into implementation behavior at all. It's specifying output and allowing for options.
> 
> Well, unforced ambiguous behavior then. Specifying output and allowing for (unrequested) options.
> 
> For instance when the SomeTool processor outputs a number 42 and the options are that it represents any base in which the symbols are generally valid (per consensus) to me would rather raise questions than answer them. I would expect these complication (extra dimensions) less inside a format but more in a "processor" role or protocol description. The latter describing expected behavior within the dynamics of some interchange.
> 
> I used the term `implementation` without caution. Sorry for that. In my reading of the sentence we discuss about, the consumer of such a result has to look elsewhere to find out what the processor provides as result, right?
> 
> That is our normal task as tool users.
> 
> However, I think we better try to not bring in such ambiguities without noting the reason outside of our mandate to change. Maybe note, that existing processors (or however we like to call these providers of jsonpath responses) display mixed behaviors (what you call options).
> 
> Makes sense?
> 
> If not, sorry for the noise.

@sthagen I have attempted to address your concern by making the following general statement (suggested by @cabo):

> A canonical representation of a nodelist is as a JSON arrays of strings, where the strings are
Normalized Paths.

A nodelist is defined as follows (in the Terminology section):

> A list of nodes. The output of applying a query to an argument is manifested as a list of nodes. While this list can be represented in JSON, e.g. as an array, this specification does not require or assume any particular representation.

From this, it is clear that an implementation of JSONPath may output a nodelist as a JSON array of strings, where the strings are Normalized Paths. So I have deleted the following statements, which seemed to be causing confusion:

> A JSONPath implementation may output
Normalized Paths instead of, or in addition to, the values identified by these paths.
For example, the JSONPath expression `$.book[?(@.price<10)]` could select two values and
an implementation could output the Normalized Paths `$['book'][3]` and `$['book'][5]`.

(I'm not sure whether this will be acceptable to the other reviewers, so I'm going to leave the PR open until they have had a chance to review the latest version.)

danielaparker:
> > I think an important aspect that this is missing is that normalized paths are contexted to the JSON data that was queried; whereas singular paths exist independently, as does any other path.
> > For example, the path `$[-3]` (used in the text) is singular and can rightfully exist on its own. However it can't be normalized without querying JSON data. 

Normalizing _paths_? What does that mean? There is no concept of "normalizing" _paths_, it's meaningless, there is no suggestion of that in Goessner's original article, or anywhere else that I'm aware of. 

>> Once queried, the resulting normalized paths can be different based on the data.
> > ```
> > * `[ 0, 1, 2, 3, 4 ]` yields a normalized path of `$[2]` for the value `2`
> > 
> > * `[ 0, 1, 2, 3, 4, 5, 6 ]` yields a normalized path of `$[4]` for the value `4`.
> > ```
> 

Normalized paths are generally _produced_ from a specific JSON value, but they can be _applied_ to other JSON values, and in fact this is not an uncommon use case. Much like a JSONPointer, except a JSONPointer cannot be fully interpreted without reference to the data, an integer appearing in a JSONPointer could be an array index or it could be an object key, whereas a normalized JSONPath, once produced, stands on its own. Think about a JSON array with a gigabyte worth of more or less heterogenous elements. It's not uncommon to produce normalized paths for the first element, and then apply selected ones to all elements.

glyn:
@cabo Not sure if you've seen the final form of this PR, but I'm going to merge it to close off the discussion threads. We can adjust the section later if necessary.

PR #290, opened 2022-11-03
Cram ABNF into 69 columns

glyn:
This isn't necessarily the prettiest solution, but it gets a big diff out of the way while there are very few PRs in flight. We can fine tune the details later on a section-by-section basis.


cabo (approved):
A few improvements are possible, but this is good enough to get the issue out of the way.
Thanks!

Issue #291, opened 2022-11-03
Fix duplicate id in generated HTML


glyn:
Commit 87397d6fbed174e88745cfb293ebbb6d8af42f38 introduced this error:
```
(No source line available): Warning: Duplicate id="identifiers" found in generated HTML.
```

PR #292, opened 2022-11-03
Avoid duplicate id

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/291


cabo (approved):
You didn't say with which id this is conflicting, but this change is certainly OK.

PR #294, opened 2022-11-13
Editorial improvements to Terminology section

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -253,19 +254,15 @@ Unicode Scalar Value:
   E000 to 10FFFF. JSON values of type string are sequences of Unicode scalar values.
 
 Singular Path:
-: A JSONPath expression built from segments which each produce at most one node.
+: A JSONPath expression built from segments which each, regardless of the input value,
+  produces at most one node.
```
cabo, line 258: I can't quite figure out whether this should be "produce" (segments which) or "produces" (each).

draft-ietf-jsonpath-base.md
```diff
@@ -253,19 +254,15 @@ Unicode Scalar Value:
...
 
 Selector:
 : A single item within a segment that takes the input value and produces a nodelist
   consisting of child nodes of the input value.
 
 For the purposes of this specification, a value as defined by
 {{-json}} is also viewed as a tree of nodes.
-Each node, in turn, holds a value.
```
cabo, line 264: Where is this snippet going?  E.g., saying that out of the members of a map, the values are nodes and the keys are not, seems important to me.

draft-ietf-jsonpath-base.md
glyn: The intention is that "produces" binds to "each". Otherwise "which each" would be left dangling.

We could reword to "... segments, each of which, regardless of the input value, produces at most one node." if that helps?

draft-ietf-jsonpath-base.md
glyn: The following definitions are the crucial ones and make it clear which are the values of an object:

> Children (of a node): If the node is an array, each of its elements, or if the node is an object, each of its member values (but not its member names). If the node is neither an array nor an object, it has no children.

> Descendants (of a node): The children of the node, together with the children of its children, and so forth recursively. More formally, the descendants relation between nodes is the transitive closure of the children relation.

Since we don't use the term "tree" elsewhere in the spec, I'll delete the above sentence.

draft-ietf-jsonpath-base.md
glyn: I made that change as it's better English.

draft-ietf-jsonpath-base.md
cabo: In these definitions, children and descendants are values, not nodes.  Maybe we can clarify this.

Otherwise, I agree that the deleted text is indeed covered.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: I've clarified that children (and therefore descendants) are nodes rather than values.

I've also moved a statement out of the terminology section which didn't belong there.

draft-ietf-jsonpath-base.md
```diff
@@ -296,7 +290,8 @@ stable JSON query language.
 
 ## Overview of JSONPath Expressions {#overview}
 
-JSONPath expressions are applied to a JSON value, the *argument*.
+A JSONPath expression is applied to a JSON value, the *argument*, and
```
cabo, line 293: I'd like a second sentence instead of an /and/, but this is just preference,

glyn: Yes, I like that. I'll submit a PR.

draft-ietf-jsonpath-base.md
```diff
@@ -207,21 +208,24 @@ Index:
...
 Query:
-: Short name for JSONPath expression.
+: Short name for a JSONPath expression.
 
 Argument:
 : Short name for the value a JSONPath expression is applied to.
 
 Node:
 : The pair of a value along with its location within the argument.
+  The type a node's value may also be referred to as the
```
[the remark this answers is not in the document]
glyn: Well spotted. If PR #296 lands, it will delete that sentence.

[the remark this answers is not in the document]
glyn: @goessner Fixed by PR #296.

PR #295, opened 2022-11-14
Delete confusing statement

glyn:
The deleted statement might be understood to mean that member names have nodes. But it's explicitly stated that member names are not values (and therefore they do not have nodes).

An alternative solution would be to qualify the deleted statement as follows, but this is very clunky in context:

> (but not its member names, which are not values and which do not have nodes)


gregsdennis:
Maybe this clarification can be included in a comment? I suspect it was included because someone read the meet names as included, and removing this removes that clarification.

PR #296, opened 2022-11-14
Avoid using JSON type terminology

glyn:
JSON types are not used much in the JSON spec. It is debatable whether there really is a boolean type or simply the values `true` and `false`. Also, the JSONPath spec doesn't really benefit from this term.


gregsdennis (approved):
I think this reads well without breaking any semantics.

PR #298, opened 2022-11-14
Separate bracketed statements

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -188,12 +188,11 @@ Member:
...
 
 Name:
-: The name in a name/value pair constituting a member.   (A name is not itself a value.)
-  (Also known as <!-- should we make it clear that names are (Unicode) strings? -->
-  "key", "tag", or "label".)
+: The name in a name/value pair constituting a member.   (A name is not itself a value.) <!-- should we make it clear that names are (Unicode) strings? -->
   This is also used in {{-json}}, but that specification does not
   formally define it.
   It is included here for completeness.
+  (Also known as "key", "tag", or "label".)
```
gregsdennis, line 195: Can we also add "property"?

glyn: I was actually considering deleting this line as I don't think it adds much, if anything, to the definition of name and is somewhat of a distraction. (When I read it, I find myself wondering in what context "label" is used in this way.) We don't use these terms elsewhere in the spec except for the use of "key" in Table 15, which could easily be replaced.

And, as you point out, there is a temptation to add other synonyms.

Alternatively, I'd be happy to slim down the list to just "key".

gregsdennis: I'm happy with deleting it over it being incomplete.

goessner: yes ... better delete it.

glyn: Deleted!

cabo (approved):

PR #301, opened 2022-11-16
Introduction tweaks

glyn:
Move the initial examples into the introduction since the purpose of the examples is introductory.

Mark certain sections as informative (rather than normative).

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -259,6 +259,8 @@ Selector:
 
 ## History
 
+This section is informative.
+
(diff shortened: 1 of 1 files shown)


cabo (approved):
This works for me.  
That said, "This section is for information only" might be a less SDO-geeky way of saying this.

glyn:
> This works for me. That said, "This section is for information only" might be a less SDO-geeky way of saying this.

That still feels a little vacuous. How about "This section is non-normative."? Too formal?

glyn:
Merging, to avoid clashes with another PR.

PR #302, opened 2022-11-16
Avoid use of the term key

glyn:
This follows on from https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/298 which deleted "key" and other synonyms of "name".

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1460,7 +1460,7 @@ Queries:
 | `$.b[?@]` | `null` | `$['b'][0]` | Existence |
 | `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
 | `$.c[?(@.d==null)]` | | | Comparison with "missing" value |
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1460,7 +1460,7 @@ Queries:
 | `$.b[?@]` | `null` | `$['b'][0]` | Existence |
 | `$.b[?@==null]` | `null` | `$['b'][0]` | Comparison |
 | `$.c[?(@.d==null)]` | | | Comparison with "missing" value |
-| `$.null` | `1` | `$['null']` | Not JSON null at all, just a string as object key |
+| `$.null` | `1` | `$['null']` | Not JSON null at all, just a name string |
```
cabo, line 1463:
```suggestion
| `$.null` | `1` | `$['null']` | Not JSON null at all, just a member name string |
```

I'd say, to be more specific

cabo (approved):

PR #303, opened 2022-11-16
Use informative instead of non-normative?

glyn:
Non-normative may be a better option, throughout, because surely normative sections are informative too?

@cabo WDYT?


goessner (approved):

glyn:
Given we use Informative for the non-normative references, I guess we should do the same throughout. Hopefully no-one will misunderstand...

cabo (commented):
👍

PR #306, opened 2022-11-16
Adjust some wording for JSON values and nodes

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -284,7 +286,10 @@ stable JSON query language.
 
 ## JSON Values
 
-The JSON value the JSONPath query is applied to is, by definition, a valid JSON value.
+The JSON value a JSONPath query is applied to is, by definition, a
+valid JSON value. A JSON value is usually constructed by parsing
```
cabo, line 290:
```suggestion
valid JSON value. A JSON value is often constructed by parsing
```

draft-ietf-jsonpath-base.md
```diff
@@ -284,7 +286,10 @@ stable JSON query language.
...
+a JSON text.
```
glyn, line 291: @cabo I'm not sure this says quite enough. It would be _possible_ to parse a JSON text into a structure in which strings included their surrounding double quotes. The problem is that RFC 8259 doesn't define an abstract JSON value. Values in that RFC are textual representations, as the following ABNF extract makes clear (e.g. values can be objects with left curly braces):
```
value = false / null / true / object / array / number / string
...
object = begin-object [ member *( value-separator member ) ]
               end-object
...
begin-object    = ws %x7B ws  ; { left curly bracket
```
I'm not sure where the notion of a JSON value as an abstract concept came from, but it sorely needs defining.

draft-ietf-jsonpath-base.md
```diff
@@ -209,7 +209,9 @@ Argument:
 : Short name for the value a JSONPath expression is applied to.
 
 Node:
-: The pair of a value along with its location within the argument.
+: The pair of a value along with its location within the argument
+  except for member names which do not have nodes. (Members are not
```
cabo, line 213: Maybe we can separate this out into 
-- member names do not have nodes
-- members aren't values in the first place, so they also do not have nodes
Maybe we can also cross-ref "Child".

draft-ietf-jsonpath-base.md
cabo: An object doesn't have curly braces.  Its representation in a JSON text does.

> the notion of a JSON value as an abstract concept came from, but it sorely needs defining.

Absolutely.  But we can't fix RFC 8259 here.  (Well, maybe we can, by simply providing the generic data model that JSON didn't define, but that might be a distraction.)

draft-ietf-jsonpath-base.md
goessner: and ... what exactly is `its location`? 

If it's the member name or element index, we should say so. If it's not, we need to define it.

cabo: The word location is used in its English sense here.
A location can be represented using a normalized path (which is a representation of a sequence of edges in the tree that a JSON value is).

goessner: Hmm ... maybe this definition qualifies to be inserted in the Terminology section.

draft-ietf-jsonpath-base.md
cabo: > Also, RFC 8259 positions a value as nothing more than a JSON text minus any external whitespace:
> 
> ```
> JSON-text = ws value ws
> ```

Yes.  Each ABNF rule name should be called representation-for-x, representation-for-false, representation-for-object, representation-for-string.  This would get tedious quickly, so they left it off, leaving it for the reader to figure out that the ABNF defines the representation in a JSON text.

draft-ietf-jsonpath-base.md
glyn: I considered separate sentences for member names and members, but that could contradict the first sentence of the definition of node. Member names, as you point out, are values and they *might* have locations (since we don't actually define what a location is!). Hence my use of "except for".

We could _note_ that a node is associated either with the argument or one of its descendants. But we can't actually define a node that way because the definition would be circular.

goessner: This is quite complicated. What about:

```
... except for member names which are no JSON values.
```

glyn: @goessner @cabo pointed out that members names are strings which _are_ JSON values.

goessner: Hmm ... so our members as `name/value` pairs are de facto `value/value` pairs? 

I would prefer to define a member name as a `string` in the sense of JSON and being quiet about the confusing logical conclusion, that it must be a JSON value then ... but of course without being a structured value (potentially having nodes) or any other primitive type.

glyn: Ok, I defined "Location" as a term, split out the statements about member names and members not being nodes, and cross referenced descendants from the definition of node.

draft-ietf-jsonpath-base.md
glyn: This may go down like a lead balloon, but I've come clean about our assumption of an abstract JSON value without going so far as to define it.

draft-ietf-jsonpath-base.md
```diff
@@ -183,12 +183,14 @@ Value:
   composed of components such as structured values, namely JSON objects and arrays, and
   primitive data, namely numbers and text strings as well as the special
   values null, true, and false.
+  {{-json}} focuses on the textual representation of JSON values and
+  doesn't define the value abstraction assumed here.
```
cabo, line 187:
```suggestion
  doesn't fully define the value abstraction assumed here.
```

Section 1 and 3 together do define a value abstraction, only just not very well.

draft-ietf-jsonpath-base.md
```diff
@@ -183,12 +183,14 @@ Value:
...
 
 Member:
 : A name/value pair in an object.  (A member is not itself a value.)
 
 Name:
-: The name in a name/value pair constituting a member.   (A name is not itself a value.) <!-- should we make it clear that names are (Unicode) strings? -->
+: The name (a string) in a name/value pair constituting a member.
   This is also used in {{-json}}, but that specification does not
```
cabo, line 194: Worse, Section 3 defines "false", "true" and "null" as "literal names".
These cannot be used in a member name position!

draft-ietf-jsonpath-base.md
```diff
@@ -208,8 +210,18 @@ Query:
 Argument:
 : Short name for the value a JSONPath expression is applied to.
 
+Location:
+: the position of a value within the argument consisting of a possibly
```
cabo, line 214: This doesn't say what these names and indexes do.
(I would prefer "can be imagined as" over "consists of".)

draft-ietf-jsonpath-base.md
glyn: That's a shame. As you wrote earlier, let's not try and fix JSON here. But if you were suggesting we say something about this, please say so.

draft-ietf-jsonpath-base.md
glyn: I've amplified the text a bit, but stopped short of providing a formal, recursive definition.

draft-ietf-jsonpath-base.md
cabo: Proposal not good?

draft-ietf-jsonpath-base.md
```diff
@@ -208,8 +210,19 @@ Query:
...
+  as a sequence of names and indexes navigating to the value through
+  the objects and arrays in the argument, with the empty sequence
+  indicating the argument itself.
+  A location can be represented as a Normalized Path (defined below).
+
 Node:
 : The pair of a value along with its location within the argument.
+  Member names do not have locations and so do not have nodes.
+  Members are not values and so do not have nodes.
+  Note that a node is associated with either the argument or one of
```
cabo, line 224: I'm not entirely happy with "is associated", but don't have a much better idea.
Maybe "relates to"?  "comprises"?

cabo (approved):
Two potential further improvements.

draft-ietf-jsonpath-base.md
glyn: Yes, sorry, I thought I'd already adopted that suggestion.

draft-ietf-jsonpath-base.md
glyn: I reworded this to avoid the awkward relationship.

PR #307, opened 2022-11-17
Add security considerations about injection attacks

cabo:


draft-ietf-jsonpath-base.md
```diff
@@ -1646,6 +1647,23 @@ exponential, CPU usage or, for example via a naive recursive implementation of t
...
 
+## Attack Vectors on How JSONPath Queries are Formed
+
+JSONPath queries are often not static, but formed from variables that
+provide index values, member names, or values to compare with in a
+filter expression.
+These variables need to be translated into the form they take in a
+JSONPath query, e.g., by escaping string delimiters, or by only
+allowing specific constructs such as `.name` to be formed when the
+given values allow that.
```
timbray, line 1658: Don't understand this point at all.

draft-ietf-jsonpath-base.md
```diff
@@ -1646,6 +1647,23 @@ exponential, CPU usage or, for example via a naive recursive implementation of t
...
+Failure to perform these translations correctly can lead to unexpected
+failures, which can lead to Availability, Confidentiality, and
+Integrity breaches, in particular if an adversary has control over the
+values (e.g., by entering them into a Web form).
+The resulting class of attacks, *injections* (e.g., SQL injections),
+is consistently found among the top causes of application security
+vulnerabilities and requires particular attention.
```
timbray, line 1665: Would be improved by examples?

goessner: I am reluctant here. Examples might be used as starting point for potential attackers.

glyn: @goessner I understand your reluctance, but the examples are likely to be simple attacks which anyone with the relevant background (attackers, penetration testers, etc.) would think of immediately.

draft-ietf-jsonpath-base.md
```diff
@@ -1646,6 +1647,23 @@ exponential, CPU usage or, for example via a naive recursive implementation of t
...
 to mitigate these attacks.
 
+## Attack Vectors on How JSONPath Queries are Formed
+
+JSONPath queries are often not static, but formed from variables that
+provide index values, member names, or values to compare with in a
+filter expression.
+These variables need to be translated into the form they take in a
+JSONPath query, e.g., by escaping string delimiters, or by only
+allowing specific constructs such as `.name` to be formed when the
```
glyn, line 1657: > Don't understand this point at all.

How about this?

```suggestion
allowing specific constructs, such as  `dot-member-name-shorthand`, to be formed when the
```

cabo: This assumes the reader of the Security Considerations has the ABNF at hand.  I'm not sure that is a good assumption.
I think this can be expanded into an example; I'll make an update to the PR.

glyn: Fair enough.

cabo:
We wanted to add examples, but maybe we can merge this as is.  Marking as ready for review.

glyn (approved):
Let's get this in and polish it later.

PR #308, opened 2022-11-18
Polish introduction

glyn:


cabo (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -231,7 +231,7 @@ Children (of a node):
 : If the node is an array, the nodes of its elements.
   If the node is an object, the nodes of its member values.
   If the node is neither an array nor an object, it has no children.
-  Note that the member names of an object are not values and do not
+  Note that the members and member names of an object do not
   have nodes.
```
goessner, line 235: `Note that the members of an object do not have nodes.`

This is a subtlety ... in fact they (potentially) do have implicitly. Most readers will be confused.

`Note that the member names of an object do not have nodes.`

This is a confusing subtlety again ... how should they, as they are merely strings.

cabo: > `Note that the members of an object do not have nodes.`
> 
> This is a subtlety ... in fact they (potentially) do have implicitly. Most readers will be confused.

Why are you saying this?
Members (the name/value pairs) do not have nodes.  The values in there do.

> `Note that the member names of an object do not have nodes.`
> 
> This is a confusing subtlety again ... how should they, as they are merely strings.

Strings have nodes.  They do not have children.
`{"foo": "bar"}` + `$.foo` ➔ the string node with the value "bar"

goessner: > > `Note that the members of an object do not have nodes.`
> > This is a subtlety ... in fact they (potentially) do have implicitly. Most readers will be confused.
> 
> Why are you saying this? Members (the name/value pairs) do not have nodes. The values in there do.

This is what I mean by 'implicitly'. At least we need to add that explanation: "The values in there potentially do."
> 
> > `Note that the member names of an object do not have nodes.`
> > This is a confusing subtlety again ... how should they, as they are merely strings.
> 
> Strings have nodes. They do not have children. `{"foo": "bar"}` + `$.foo` ➔ the string node with the value "bar"

We are talking about the member name `"foo"`, which does not have nodes. It only is a component of one. Here an explanation is required also.

PR #309, opened 2022-11-18
Clarify why members and member names do not have nodes

glyn:
This is a draft for us to contemplate and update to see if we can reach a satisfactory clarification.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -219,10 +219,11 @@ Location:
 
 Node:
 : The pair of a value along with its location within the argument.
-  Member names do not have locations and so do not have nodes.
-  Members are not values and so do not have nodes.
   Note that a node is either the root node (defined below) or one of
   its descendants (also defined below).
+  Member names and members cannot be selected by a query, but a member
+  value can. So nodes are defined for member values, but not for
+  members or member names.
 
 Root Node:
 : The unique node whose value is the entire argument.
@@ -231,8 +232,7 @@ Children (of a node):
 : If the node is an array, the nodes of its elements.
   If the node is an object, the nodes of its member values.
   If the node is neither an array nor an object, it has no children.
-  Note that the members and member names of an object do not
-  have nodes.
+  Note that members and member names are not children of an object.
 
 Descendants (of a node):
 : The children of the node, together with the children of its children, and so forth
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -219,11 +219,6 @@ Location:
 
 Node:
 : The pair of a value along with its location within the argument.
-  Note that a node is either the root node (defined below) or one of
-  its descendants (also defined below).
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -219,10 +219,11 @@ Location:
 
 Node:
 : The pair of a value along with its location within the argument.
-  Member names do not have locations and so do not have nodes.
-  Members are not values and so do not have nodes.
   Note that a node is either the root node (defined below) or one of
   its descendants (also defined below).
+  Member names and members cannot be selected by a query, but a member
```
cabo, line 224: The only weird part here for me is that members aren't even values, so you cannot build the pair the node is supposed to be.

cabo (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -219,10 +219,11 @@ Location:
...
+  value can. So nodes are defined for member values, but not for
+  members or member names.
```
goessner, line 226: Perhaps clearer/shorter:

`Only member values can hold nodes.`

draft-ietf-jsonpath-base.md
goessner: Yes, in fact it is a current drawback of JSONPath, that we cannot select values by their member names (multiply addressed in the past).
```
{""foo":"bar"}

$[?nameOf(@) == 'foo']
```
Extension points for rescue ...

goessner: Perhaps we need to more frequently remind the distinction:

member ➔ name/value pair
node  ➔ value along with its location in the argument

draft-ietf-jsonpath-base.md
```diff
@@ -231,8 +232,7 @@ Children (of a node):
 : If the node is an array, the nodes of its elements.
   If the node is an object, the nodes of its member values.
   If the node is neither an array nor an object, it has no children.
-  Note that the members and member names of an object do not
-  have nodes.
+  Note that members and member names are not children of an object.
```
goessner, line 235: Perhaps we should sharpen the definitions instead:

```
Member:  A name/value pair in an object.  (Neither itself a value, nor a node.)
```

```
Name:  The name in a name/value pair constituting a member.  (Not itself a node.)
```

glyn: We had (something like) that previously. I prefer to avoid the forward references to "node" because some readers may carry all sorts of assumptions about the term "node" before they read the actual definition.

draft-ietf-jsonpath-base.md
glyn: > Perhaps we need to more frequently remind the distinction:
> 
> member ➔ name/value pair
> node ➔ value along with its location in the argument

I've included a reminder of what a member is in the new section introduced by 8da0011da49abb742ff22515c89560901e07b2ac.

We should check the rest of the spec to see if there are usages of "member" and "node" for which readers may need reminding of the definitions.

PR #311, opened 2022-11-21
Mention JSONPath expressions before query term

glyn:
The definition of a query as a short name for a JSONPath expression then makes more sense.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -150,7 +150,7 @@ from a JSON value.
 
 JSON {{-json}} is a popular representation
 format for structured data values.
-JSONPath defines a string syntax for selecting and extracting JSON values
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -150,7 +150,7 @@ from a JSON value.
 
 JSON {{-json}} is a popular representation
 format for structured data values.
-JSONPath defines a string syntax for selecting and extracting JSON values
+JSONPath defines string expressions for selecting and extracting JSON values
```
cabo, line 153: Hmm, this loses that this is a new syntax, specifically for JSONPath.
"expressions" is very general here, which raises more questions than it answers.
I like the old sentence much better in this place.

glyn: A fair point. Not sure how to firm up the definition of the term "query", but let's address that separately.

PR #313, opened 2022-11-22
Consistently refer to "this document"; prefer "language" to "platform"

glyn:
Generally, use "this document" rather than "this specification". Then use the term specification judiciously.

Prefer the term "language" to "platform".

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -293,16 +293,15 @@ deployment of its implementations, and provides a normative specification for it
 [XPath].
 
 JSONPath was intended as a light-weight companion to JSON
-implementations on platforms such as PHP and JavaScript, so instead of
-defining its own expression language like XPath did, JSONPath
-delegated this to the expression language of the platform.
-While the languages in which JSONPath is used do have significant
-commonalities, over time this caused non-portability of JSONPath
-expressions between the ensuing platform-specific dialects.
+implementations in languages such as PHP and JavaScript, so instead of
+defining its own expression language, like XPath did, JSONPath
+delegated to the implementation language.
+As JSONPath was implemented in more languages, JSONPath
+expressions became decreasingly portable.
 
-This document aims to remove platform dependencies and
+This document aims to remove language dependencies and
 serve as a common JSONPath specification that can be used across
-platforms.  Obviously, this means that backwards compatibility could
+languages.  Obviously, this means that backwards compatibility could
 not always be achieved; a design principle of this document is to
 go with a "consensus" between implementations even if it is rough, as
 long as that does not jeopardize the objective of obtaining a usable,
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -176,7 +176,7 @@ unchanged.
 Importantly "object" and "array" in particular do not take on a
 generic meaning, such as they would in a general programming context.
 
-Additional terms used in this specification are defined below.
+Additional terms used in this document are defined below.
 
 Value:
 : As per {{-json}}, a structure conforming to the generic data model of JSON, i.e.,
@@ -243,7 +243,7 @@ Segment:
 
 Nodelist:
 : A list of nodes.
-  While a nodelist can be represented in JSON, e.g. as an array, this specification
+  While a nodelist can be represented in JSON, e.g. as an array, this document
   does not require or assume any particular representation.
 
 Normalized Path:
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1469,7 +1469,7 @@ each time it appears in the descendant segment.
 
 The example above with the query `$.a..[0, 1]` shows that the child segment `[0, 1]` is applied to each node
 in turn (rather than the nodes being visited once per selector, which is the case for some JSONPath implementations
-that do not conform to this specification).
+that do not conform to this document).
```
glyn, line 1472: Or:
```suggestion
that do not conform to the specification in this document).
```
?

draft-ietf-jsonpath-base.md
```diff
@@ -287,23 +287,22 @@ This section is informative.
...
+defining its own expression language, like XPath did, JSONPath
+delegated to the implementation language.
+As JSONPath was implemented in more languages, JSONPath
+expressions became decreasingly portable.
+
+This document aims to remove language dependencies and
 serve as a common JSONPath specification that can be used across
-platforms.  Obviously, this means that backwards compatibility could
-not always be achieved; a design principle of this specification is to
+languages.  Obviously, this means that backwards compatibility could
```
cabo, line 304: I have no idea what languages have to do with this.
Platform was the right word.

draft-ietf-jsonpath-base.md
```diff
@@ -176,7 +176,7 @@ unchanged.
 Importantly "object" and "array" in particular do not take on a
 generic meaning, such as they would in a general programming context.
 
-Additional terms used in this specification are defined below.
+Additional terms used in this document are defined below.
```
cabo, line 179: I'm mostly neutral to this change.
(In this particular place, it is an actual improvement, but not necessarily elsewhere.)
"This document" is more confusing in a spec that talks about documents; fortunately, we don't.

draft-ietf-jsonpath-base.md
```diff
@@ -266,11 +266,11 @@ Selector:
 
 ### JSON Values as Trees of Nodes
 
-This specification models the argument as a tree of JSON values, each
+This document models the argument as a tree of JSON values, each
```
cabo, line 269: Another way to write this is "We model...".

draft-ietf-jsonpath-base.md
glyn: I'm concerned that JSONPath should interoperate between implementations in distinct languages on the same platform (as well as between platforms). Make sense?

draft-ietf-jsonpath-base.md
glyn: True. I wanted to avoid that voice because the tone can sometimes come across as a little superior.

draft-ietf-jsonpath-base.md
cabo: The implementation language is part of the platform.

(I didn't mean OS platforms.)

The user of the JSONPath implementation should not be concerned with the implementation language.
Multiple platforms can be implemented in the same language (e.g., C or C++).

glyn: To me, "platform" means either OS or hardware. [Wikipedia](https://en.wikipedia.org/wiki/Computing_platform) expands on that somewhat. So I don't think we can use the unqualified term "platform" and be sure that our readers will understand by it the same thing we do (even if we agreed among ourselves).

Most of my experience is in C, C++, Java, Go, and Rust all of which support multiple OS/hardware platforms. I think I would describe thing you are calling a platform as a language runtime.

Please could you define what you mean by platform?

draft-ietf-jsonpath-base.md
glyn: I'm not entirely sold on it either. I'll let this PR go cold and then go through the cases and see which I think are improvements.

goessner: RFC 8259 also uses "document" for self reference. I am ok with this change.

draft-ietf-jsonpath-base.md
goessner: I have a slight preference for "language" also.

glyn: @cabo PTAL at the latest version of this PR. If you are still uncomfortable with the changes to the history section, please could you define what you mean by platform in this context?

cabo: The problem is that the variation that we are observing does not come from the programming language only.  The platform (not just OS platform) might provide more variation, e.g., whether RE2 or PCRE or a Boost dialect is available for the RE processing -- all of these could be in a C++ implementation.

If your view of "platform" is more narrowly "OS platform", then we maybe need a term for what I call "platform".

cabo (approved):
Yep, environment is a nice synonym of platform.
Thank you for addressing my concern!

PR #314, opened 2022-11-22
Clean up the term index

glyn:
Avoid the term indexing, so we don't need to worry about whether it applies only to arrays or to arrays and objects.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -204,10 +204,7 @@ Element:
 : A value in a JSON array.
 
 Index:
-: A non-negative integer that identifies a specific element in an array.
-  Note that the term _indexing_ is also used for accessing elements
-  using negative integers ({{index-semantics}}), and for accessing
-  member values in an object using their member name.
+: An integer that identifies a specific element in an array.
 
 Query:
 : Short name for a JSONPath expression.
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -963,7 +960,7 @@ FUNCTION Normalize(i, len):
   END IF
 ~~~~
 
-The result of the array indexing expression `i` applied to an array
+The result of the array index expression `i` applied to an array
```
cabo, line 963: Well, this isn't a pure array index.
I think we still want to distinguish the index (0..n-1) from indexing (which, for instance, can take -2).

cabo (commented):
This may have been a bit too much cleanup, losing the distinction between an array index (0..n-1) and the operation of indexing.

draft-ietf-jsonpath-base.md
glyn: I'll take another look at this PR with that distinction in mind and see how it goes.

glyn: @cabo: re-reading the spec with this PR applied (and after tightening up the semantics of the index selector), I think the generalised index terminology works ok. An index is now defined to be "An integer that identifies a specific element in an array" which includes the case of negative indices. So I don't think we need the rather fine distinction between index and indexing (this term is no longer used). Please could you take a fresh look and see if you agree.

PR #317, opened 2022-11-28
Clarify JSONPath validity requirements

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -487,15 +487,15 @@ A string to be used as a JSONPath query needs to be *well-formed* and
...
-To be valid, integer numbers in the JSONPath query that are relevant
+1. Integer numbers in the JSONPath query that are relevant
 to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
-within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1]).
+within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
-To be valid, strings on the right-hand side of the `=~` regex matching
-operator need to conform to {{-iregexp}}.
+2. Strings on the right-hand side of the `=~` regex matching
```
cabo, line 497: That should be gone soon.

cabo (approved):
I like this structure.
We might come up with more validity requirements, so we may want to keep this even if =~ goes away (which is pretty much what we decided last week).

PR #319, opened 2022-11-28
Editorial changes to semantics section

glyn:
Specifically, use the nodelist term more consistently with elsewhere in the document.


cabo (approved):
LGTM

PR #320, opened 2022-11-29
Function validity

glyn:
Define syntactic validity requirements for function expressions.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1144,6 +1144,7 @@ Children of an array appear in array order in the resultant nodelist.
 {: unnumbered}
 
 A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
+A function expression by itself in a Boolean context is also an existence test which yields true if the function returns a non-empty nodelist or a value and yields false if the function returns an empty nodelist.
 
 Existence tests differ from comparisons in that:
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -261,9 +261,12 @@ Unicode Scalar Value:
   In other words, integers in either of the inclusive base 16 ranges 0 to D7FF and
   E000 to 10FFFF. JSON string values are sequences of Unicode scalar values.
 
+Singular Nodelist:
+: A nodelist containing at most one node.
+
 Singular Path:
 : A JSONPath expression built from segments each of which, regardless of the input value,
-  produces at most one node.
+  produces a Singular Nodelist.
 
 Selector:
 : A single item within a segment that takes the input value and produces a nodelist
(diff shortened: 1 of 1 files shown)


glyn:
@cabo: This draft PR may not be complete, but it's at least a start.

draft-ietf-jsonpath-base.md
```diff
@@ -502,6 +502,14 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
+2. Any function extensions used by the query MUST be supported by the JSONPath
+implementation, MUST have the correct number of arguments of the correct kinds
```
gregsdennis, line 506: I think this is reversed. It sounds like if `foo()` is in the query, then the implementation is required to support it.  But an implementation can't support functions it doesn't know about.

What I think you want is that queries must only use functions that are supported by the implementation.  This way the requirement is on the query author, not the implementation.

draft-ietf-jsonpath-base.md
```diff
@@ -1248,9 +1257,20 @@ function-argument       = comparable / filter-path
 ~~~
 
 Syntactically, a function-expression can occur anywhere where a
-singular-path can occur (exist-expr, comparable).
-A function-expression that employs a function extension that returns a
-nodelist MUST return a singular node to be used in a comparable.
+singular-path can occur (comparable) and in existence tests.
+
+Any function extensions used by the query MUST be supported by the JSONPath
+implementation, MUST have the correct number of arguments of the correct kinds
```
gregsdennis, line 1263: Here too

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
cabo: The query author cannot know what the implementation will be that interprets the query.

The requirement does need to be on the implementation: Detect an unsupported function and handle it as a (static -- before the query is even performed) error.

glyn: Yes, that's was what I intended.

gregsdennis: @glyn, which is what you intended?

As an implementor, I can't know what authors are doing to use in their paths.  All I can do is offer support for specific functions.

The way the text is worded, it sounds like if my implementation doesn't support a function, my implementation is at fault, when it's really the author (who very likely _does_ have some sort of tooling documentation as to which functions are supported, even if they don't know the underlying implementation) who knowingly used an unsupported function.

I don't need users opening bugs saying that I MUST support some function because the spec requires I support any function they decide to put in their queries.

---

What I _think_ this is trying to say is that query authors need to only used functions that are explicitly supported by the tooling they're using.

gregsdennis: > Any function extensions used by the query MUST be supported by the JSONPath implementation.

This text can be read two ways:

1. The query MUST only use functions extensions that are supported.
2. The implementation MUST support any function extensions that may appear in the query.

What I think you mean is option 1.

Option 2 is a backwards requirement and cannot be fulfilled by any implementation.

Regardless, this text is not clear.

gregsdennis: > Detect an unsupported function and handle it as a (static -- before the query is even performed) error.

I don't disagree with you on this, but this text isn't about erroring on unknown functions.

glyn: I get it now. Thanks @gregsdennis. I've started to tease that out with some more text.

glyn:
@cabo: This PR is still very much a work in progress. I've started to explore a type system for function extension arguments and return values. I think a Singular Nodelist subtype of Nodelist may be the way to go. Singular Nodelists then fit where Singular Paths are allowed. (General) Nodelists can go anywhere a path is allowed. We may want to scrap this PR and make clean start, but at least this will serve as working notes for now.

glyn:
I have reworked this PR substantially and addressed the comments. I will submit it as a new PR so the comments can start from scratch.

glyn:
See the new PR https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/330.

PR #328, opened 2022-11-30
Remove =/

glyn:
Merge function expressions into the ABNF productions where they belong. This is more readable than employing =/.


cabo (approved):
Thanks.  More editorials for #280 coming up...

Issue #327, opened 2022-11-30
Cardinality of nodelists returned by function extensions


glyn:
Currently function extensions that return a nodelist must return a nodelist consisting of a single node.

When a function extension is used as a comparable, it should be able to return an empty nodelist.

When a function extension is used in an existence test, it should be able to return any number of nodes in its nodelist. (Currently, such an existence test always yields true since the returned nodelist is non-empty.)

gregsdennis:
Also if a function is used as an input to another function, it should be allowed to return multiple nodes.

glyn:
> Also if a function is used as an input to another function, it should be allowed to return multiple nodes.

Agreed. PR 330 addresses this.

PR #330, opened 2022-12-03
Define syntactic validity of function expressions

glyn:
This PR also addresses https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/327.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -529,7 +529,7 @@ errors.)
 
 ## Syntax
 
-Syntactially, a JSONPath query consists of a root identifier (`$`), which
+Syntactically, a JSONPath query consists of a root identifier (`$`), which
 stands for a nodelist that contains the root node of the argument,
 followed by a possibly empty sequence of *segments*.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -261,9 +261,12 @@ Unicode Scalar Value:
   In other words, integers in either of the inclusive base 16 ranges 0 to D7FF and
   E000 to 10FFFF. JSON string values are sequences of Unicode scalar values.
 
+Singular Nodelist:
+: A nodelist containing at most one node.
+
 Singular Path:
 : A JSONPath expression built from segments each of which, regardless of the input value,
-  produces at most one node.
+  produces a Singular Nodelist.
 
 Selector:
 : A single item within a segment that takes the input value and produces a nodelist
@@ -501,6 +504,11 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
+2. Uses of function extensions must be syntactically valid,
+as described in Section {{<fnex}}.
+
+A JSONPath implementation MUST raise an error for any query which is not
+well-formed and valid.
 The well-formedness and the validity of JSONPath queries are independent of
 the JSON value the query is applied to; no further errors relating to the
 well-formedness and the validity of a JSONPath query can be
@@ -1097,7 +1105,8 @@ comp-op      = "==" / "!=" /                    ; comparison ...
                "<"  / ">"  /                    ; operators
                "<=" / ">="
 
-singular-path     = rel-singular-path / abs-singular-path / function-expression
+singular-path     = rel-singular-path / abs-singular-path /
+                    function-expression
 rel-singular-path = current-node-identifier singular-path-segments
 abs-singular-path = root-identifier singular-path-segments
 singular-path-segments = *(S (name-segment / index-segment))
@@ -1303,8 +1312,9 @@ A function-expression stands for ("returns") a nodelist or a value.
 The arguments are filter expressions that are to be interpreted as
 nodelists or values.
 For each argument and for the result, the function extension defines
-the "kind" of the item, i.e., whether the item is a value or a
-nodelist ("nodes").
+the "kind" of the item, i.e., whether the item is a value, a
+nodelist (i.e., containing any number of nodes), or a Singular Nodelist
+(i.e., containing at most one node).
 
 ~~~ abnf
 function-name           = function-name-first *function-name-char
@@ -1317,10 +1327,25 @@ function-expression     = function-name "(" S [function-argument
 function-argument       = comparable / filter-path
 ~~~
 
-Syntactically, a function-expression can occur anywhere where a
-singular-path can occur (exist-expr, comparable).
-A function-expression that employs a function extension that returns a
-nodelist MUST return a singular node to be used in a comparable.
+Any function expressions in a query must be syntactically valid,
+otherwise the JSONPath implementation MUST raise an error
+(see Section {{<synsem-overview}}).
+
+A function expression is syntactically valid if:
+
+* It conforms to the ABNF syntax above.
+* It occurs as a filter-path in an existence test and the function
+is defined to return either a nodelist or a Singular Nodelist.
+* It occurs as a singular-path on either side of a comparison and the function
+is defined to return either a value or a Singular Nodelist.
+* For the function expression and any function expression it contains,
+each argument of the function matches the defined kind of the argument:
+  * An argument of kind value takes a primitive value or a function expression
+    that returns a value.
+  * An argument of kind nodelist takes a filter-path, a Singular Path, or a
+    function expression that returns a nodelist or a Singular Nodelist.
+  * An argument of kind Singular Nodelist takes a Singular Path or a function
+    expression that returns a Singular Nodelist.
 
 ### `length` Function Extension {#length}
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1060,9 +1060,9 @@ During the iteration process the node of each array element or object member val
 A boolean expression, usually involving the current node, is evaluated and
 the current node is selected if and only if the expression yields true.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1317,10 +1327,25 @@ function-expression     = function-name "(" S [function-argument
...
+A function expression is syntactically valid if:
+
+* It conforms to the ABNF syntax above.
+* It occurs as a filter-path in an existence test and the function
+is defined to return either a nodelist or a Singular Nodelist.
+* It occurs as a singular-path on either side of a comparison and the function
+is defined to return either a value or a Singular Nodelist.
+* For the function expression and any function expression it contains,
+each argument of the function matches the defined kind of the argument:
+  * An argument of kind value takes a primitive value or a function expression
```
cabo, line 1343: This doesn't convert a singular path to a value.
`length($.foo)` would not work?

glyn: Correct, and that's too limiting. I was trying to avoid the issue a function expecting a value but receiving an empty nodelist from a singular path. Here are some alternatives if we allowed a singular path or a function expression returning a singular nodelist:

* We could coerce an empty nodelist to, for example, `null` at runtime, but we've managed to avoid coercions elsewhere in the spec.
* We could cause the filter not to match for the current item. This would impact the Boolean laws which would no longer be true in general.

glyn: Another alternative occurred to me. We could introduce a function extension `value_or_else()` which takes a singular nodelist and a default value and returns a value. The result would be the value of the node if the nodelist is non-empty and the default value otherwise. We could then use `length(value_or_null(<singular path>, ""))`. This is probably too long-winded to be practical.

Another alternative would be to define another `length()` function which takes a singular nodelist and which can be defined to return a suitable default (such as 0 or -1) if the nodelist is empty. Not sure we want to get into overloading of function extension names. That could be an issue for IANA.

goessner: I assume that with `length($.foo)` not working, we imply `$.foo` is taken as an existence test, right?

In the same way as an application needs to distinguish occurences of paths as part of comparisons or just not, they should be able to detect, if a path occurs as a function parameter and might be handed over as it is then. That way the burden/freedom to correctly treat the path lies at the function author (simply taking its value here with `length`).

Have I missed something here?

cabo: > I assume that with `length($.foo)` not working, we imply `$.foo` is taken as an existence test, right?

Oops.  That was not what I intended -- length wants an array, a map or a text string.
So converting this to a Boolean is not what one wants here.
(But for some other function extension, this might exactly be what is wanted -- however, this could then possibly be forced by `!!$.foo`.)
So maybe we need to distinguish function parameters where the arguments are evaluated in a comparison-like context, and parameters where the arguments are evaluated as an existence test.

goessner: Yes, `!!$.foo` would be perfectly fine in case of an existence test as parameter.

But ... exactly what prevents `length($.foo)` from working then?

draft-ietf-jsonpath-base.md
```diff
@@ -1317,10 +1327,25 @@ function-expression     = function-name "(" S [function-argument
...
+otherwise the JSONPath implementation MUST raise an error
+(see Section {{<synsem-overview}}).
+
+A function expression is syntactically valid if:
+
+* It conforms to the ABNF syntax above.
+* It occurs as a filter-path in an existence test and the function
+is defined to return either a nodelist or a Singular Nodelist.
+* It occurs as a singular-path on either side of a comparison and the function
+is defined to return either a value or a Singular Nodelist.
```
gregsdennis, line 1340: So are we not accepting functions which _may_ return a single value but _could_ return multiple values?

This could only be a runtime determination, so static syntactical analysis couldn't find it.  (I realize that's an argument against such functions; just clarifying.)

glyn: Static checks reduce the number of runtime errors we have to cope with in the spec, so I'd prefer to rule out that kind of function signature. However, if we allowed a function to return a list of values (which we currently don't), then such a function could return a list of size one.

draft-ietf-jsonpath-base.md
glyn: Expanding on this functional programming approach, we could define the following types, where `Nodes(l)` is an instance for some nodelist `l` (which could be represented either as a path or as a nodelist):
* `OptionalNodeOrValue = {None, Node(n), Value(v)}`
* `OptionalNode = {None, Node(n)}`
* `OptionalValue = {None, Value(v)}`
* `Value = {Value(v)}`
* `Absent = {None}` (this may not be necessary, but allows, for example, a logging function with no return value)
* `OptionalNodes = {None, Nodes(l)}` (where `None` is equivalent to `Nodes(e)`, where `e` is the empty nodelist)

Each function argument  or return value would have one of the above types. 

The types form a hierarchy (with subtypes indented):
```
OptionalNodeOrValue
    OptionalNode
        Absent
    OptionalValue
        Value
        Absent
    Value
    Absent

OptionalNodes
    OptionalNode
        Absent
```

So, for example, a function with return value of type `Value` could be passed to another function with an argument of type `OptionalValue`.

Note that `OptionalNode` is a subtype of `OptionalNodes` even though the instances of the former are not a subset of the instances of the latter. The `OptionalNode` instance `Node(n)` can be coerced to the `OptionalNodes` instance `Nodes(l)`, where `l` is a nodelist consisting of just the node `n`.

What do people think of this approach?

goessner: This is far too abstract for me and perhaps for someone else.

From the view of an average user of JSONPath - knowing, there is a function `length`, familiar from C-family of languages - I would expect, that the query

`$[?length(@.something] >= 0]`

simply works regardless of the argument.

How would that query look like using your approach then?

cabo: I don't think we can avoid the inherent complexity.
Maybe we can slightly simplify the tree.

Defining this in an abstract way is fine for our specification.
Of course, a few examples can make this more accessible.

glyn: > How would that query look like using your approach then?

The query would be the same. I'm simply describing which function expressions are syntactically valid.

glyn: > Maybe we can slightly simplify the tree.

Note the type hierarchy (if that's the right word) isn't a tree since there are nodes such as `Value` with multiple parents.

glyn: > Of course, a few examples can make this more accessible.

For starters, here are some proposed signatures of the functions that are already in the spec:
* `length(OptionalNodeOrValue) -> OptionalValue`
* `count(OptionalNodes) -> Value`
* `match(Value) -> OptionalValue`
* `search(Value) -> OptionalValue`

For all these that return `OptionalValue`, `None` (i.e. an empty nodelist) is used in error cases (e.g. an empty nodelist argument to `length` or a non-string argument of `match` or `search`).

goessner: This makes it a bit clearer ... thanks. Also my example 

`$[?length(@.something] >= 0]`

seems to be in harmony with our comparison rules now.

gregsdennis:
My [JsonPath.Net library](https://www.nuget.org/packages/JsonPath.Net) now implements everything in the spec, including this PR.

The function infrastructure was a struggle (mostly because I'm in a strongly typed language, I suspect), but once implemented, creating the functions themselves and registering them so that they can be parsed was easy!

With my lib, the user would implement a "function definition" interface which exposes the name, the range of parameters expected (min/max), and the logic for the function itself (which follows a pattern defined in my lib).  Once an instance of their definition is registered, it's available to be parsed.  The parsing logic handles validating that the function receives the correct number of parameters, too, so if you send `length()` seven arguments, it'll fail to parse.

draft-ietf-jsonpath-base.md
```diff
@@ -1420,6 +1507,19 @@ substring that matches the iregexp contained in the string
 that is the second argument.
 The result is `true` if such a substring exists, `false` otherwise.
 
+### Examples
```
cabo, line 1510: this still uses "syntax", we probably want to say "typing" or "types"

draft-ietf-jsonpath-base.md
```diff
@@ -1314,21 +1322,94 @@ LCALPHA                 = %x61-7A  ; "a".."z"
...
+is a subset of the set of instances of the other type.
+
+{{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, and `nl` denotes
+a non-empty nodelist. The table also lists the subtypes of each type.
+
+| Type | Abstract Instances | Subtypes |
+| :--: | :----------------: | :------: |
+| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `None` | `OptionalNode`, `OptionalValue`, `Value`, `Absent` |
+| `OptionalNode` | `Node(n)`, `None` | `Absent` |
+| `OptionalValue` | `Value(v)`, `None` | `Value`, `Absent` |
```
cabo, line 1354: None is used in Python for the JSON value `null`.
We probably want to use a different name so Python users don't think these are the same.

cabo (approved):
I believe this is ready for merging (to serve as the next baseline).
My comments can either be resolved or entered as new issues.

draft-ietf-jsonpath-base.md
glyn: Fixed in 34b224404

draft-ietf-jsonpath-base.md
glyn: Changed to Nothing.

PR #332, opened 2022-12-03
Add string comparison example

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -1279,6 +1279,7 @@ JSON:
 
 | Query | Result | Result Paths | Comment |
 | :---: | ------ | :----------: | ------- |
+| `$.a[@.b == "kilo"]` | `{"b": "kilo"}` | `$['a'][9]` | Member value comparison |
```
gregsdennis, line 1282: Although double quotes are supported, should we use single quotes for `kilo`?  I think we declare they're preferred elsewhere.

cabo: Yes (in the JSONPath query, that is).

```suggestion
| `$.a[@.b == 'kilo']` | `{"b": "kilo"}` | `$['a'][9]` | Member value comparison |
```

goessner (approved):

PR #338, opened 2023-01-06
No negative zero

### LINKED ISSUE #337: Index selector [-0]
According to chapter 2.5.3, using -0 as index is allowed, but no semantics for this is defined (and probably should not be defined).
Proposal would be to exclude -0, i.e. replace the line

int = ["-"] ( "0" / (DIGIT1 *DIGIT) ) ; - optional

by

int = ( "0" / ["-"] (DIGIT1 *DIGIT) ) ; - optional

This also makes sense for all other occurrences of "int" in the document.

cabo:
Close #337

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -18,7 +18,7 @@ abbrev: JSONPath
 area: ART
 wg: JSONPath WG
 kw: JSON
-date: 2022
+date: 2023
 
 author:
 -
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -825,7 +825,7 @@ An index selector `<index>` matches at most one array element value.
 ~~~~ abnf
 index-selector = int                             ; decimal integer
 
-int            = ["-"] ( "0" / (DIGIT1 *DIGIT) ) ; -  optional
+int            = "0" / (["-"] DIGIT1 *DIGIT)     ; -  optional
 DIGIT1         = %x31-39                         ; 1-9 non-zero digit
 ~~~~
 
@@ -1110,7 +1110,7 @@ Alphabetic characters in ABNF are case-insensitive, so "e" can be either "e" or 
 `true`, `false`, and `null` are lower-case only (case-sensitive).
 
 ~~~~ abnf
-number       = int [ frac ] [ exp ]                ; decimal number
+number       = (int / "-0") [ frac ] [ exp ]       ; decimal number
 frac         = "." 1*DIGIT                         ; decimal fraction
 exp          = "e" [ "-" / "+" ] 1*DIGIT           ; decimal exponent
 true         = %x74.72.75.65                       ; true


goessner:
... we would need another `int` then with 

```number       = int [ frac ] [ exp ]                ; decimal number```

in chapter 2.5.5.

JavaScript states 
``` -0 === 0  // true ```

Here is a discussion more in depth about it.

https://stackoverflow.com/questions/3139538/is-minus-zero-0-equivalent-to-zero-0-in-c-sharp

Since JSONPath queries might be automatically generated, an occurence of  `-0` cannot be excluded.

So I am in favour of keeping this more general &ndash; admittedly a bit slanted &ndash; current formulation of  `int`.

**Comment by cabo** [comment]
Indeed, I changed the definition of "number" (line 1113) to re-enable -0 etc.
See comment on https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/338/commits/914856726dd75d4941d86a437b5612e4631eb4e2

**Comment by goessner** [comment]
ok.

Since `int` occurs only one more time, we don't need it anymore then, so we can write more compact with fewer dependencies

```index-selector = ( "0" / ["-"] (DIGIT1 *DIGIT) ) ; - optional ```

and then later 

```number       =  ["-"] ( "0" / (DIGIT1 *DIGIT) ) [ frac ] [ exp ]     ; decimal number```

**Comment by cabo** [comment]
int also occurs in:

```
start          = int       ; included in selection
end            = int       ; not included in selection
step           = int       ; default: 1
```

so I think we should stick with the additional production.

**Comment by goessner** [comment]
yes ... I obviously missed that.

File: draft-ietf-jsonpath-base.md [hunk 1]
```diff
@@ -825,7 +825,7 @@ An index selector `<index>` matches at most one array element value.
 ~~~~ abnf
 index-selector = int                             ; decimal integer
 
-int            = ["-"] ( "0" / (DIGIT1 *DIGIT) ) ; -  optional
+int            = "0" / (["-"] DIGIT1 *DIGIT)     ; -  optional
```
glyn, line 828: I think this deserves another note (#3) below along these lines:

3. The syntax does *not* allow `-0` (which is a valid JSON number).

(Editorial observation: in the base text, it's a little odd that we have the notes below, but we don't have similar notes in the array slice selector syntax section. No need to change this in the current PR, but it wouldn't harm.)

glyn (approved):
I suggest adding a note.

draft-ietf-jsonpath-base.md
cabo: I think I'd go the inverse direction and delete the "; - optional" comment -- the ABNF really should speak for itself.

Note that JSON does not have integers, and we do allow -0 in JSON numbers.
I'm having a hard time imagining anyone being too surprised about not being able to use -0 as an integer.

glyn: Interesting. I'm comfortable with the policy of letting the ABNF speak for itself, but I know of at least one reader of the spec who said that he didn't pay much attention to the ABNF. Anyway, let's not switch policy in this PR. Approving.

glyn (approved):
Approved.

PR #341, opened 2023-01-09
Enable the use of [Optional]Boolean functions without comparison

### LINKED ISSUE #339: Ergonomics of regexp usage
Now we have switched to using function extensions for regexp filter matches, the current syntax requires the result to be compared against a boolean, as shown in this example from PR 330:
```
$[?match(@.timezone, 'Europe/.*') == true]
```

This is consistent with the following statement in the "Existence Tests" section of the spec:
> To test the value of a node selected by a path, an explicit comparison is necessary. 

If we wanted the following syntax to act as a regexp test:
```
```

we would have to allow a function expression (returning a boolean result) to be treated as a `relation-expr` (rather than an `exist-expr` or a `comparable`). This would require (a) considerable rework of filter selectors and (b) making a special case of functions which return booleans. To be clear: I am **not** in favour of such a change.

The reason for raising this issue was so that it doesn't come as a surprise to the WG. Thanks to @cabo who spotted it when reviewing some draft input to PR 330.

cabo:
Fix #339

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -18,7 +18,7 @@ abbrev: JSONPath
 area: ART
 wg: JSONPath WG
 kw: JSON
-date: 2022
+date: 2023
 
 author:
 -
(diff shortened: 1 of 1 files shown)


glyn (approved):
Excellent!

PR #343, opened 2023-01-10
Editorial nits

timbray:
Signed-off-by: Tim Bray <tbray@textuality.com>

Nothing fancy here, just like the title says.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -285,11 +285,11 @@ A node is either the root node or one of its descendants.
 This document models the result of applying a query to the
 argument as a nodelist (a list of nodes).
 
-So nodes are the selectable parts of the argument.
+Nodes are the selectable parts of the argument.
 The only parts of an object that can be selected by a query are the
 member values. Member names and members (name/value pairs) cannot be
 selected.
-So member values have nodes, but members and member names do not.
+Thus, member values have nodes, but members and member names do not.
 Similarly, member values are children of an object, but members and
 member names are not.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -550,7 +550,9 @@ The syntax and semantics of segments are defined in {{segments-details}}.
 
 In this document, the semantics of a JSONPath query define the
 required results and do not prescribe the internal workings of an
-implementation.
+implementation.  This document may describe semantics in a procedural
+step-by-stap fashion, but such descriptions are non-normative and do not
```
cabo, line 554: Well, the descriptions are meant to be normative, but the way they are described is not the only way of implementing them.

cabo (approved):
Thank you.
The text about how the descriptions are not prescribing implementation details but are still normatively describing the semantics could be improved, e.g., using some "same result" language.

draft-ietf-jsonpath-base.md
timbray: Something like "… step-by-step function, which are normative in the sense that any implementation MUST produce an identical result, but not in the sense that implementors are required to use the same algorithms" ?

cabo, line 554:
```suggestion
step-by-step fashion, but such descriptions are normative only in the sense that any implementation MUST produce an identical result, but not in the sense that implementors are required to use the same algorithms.
```

draft-ietf-jsonpath-base.md
```diff
@@ -550,7 +550,9 @@ The syntax and semantics of segments are defined in {{segments-details}}.
...
+constrain implementors.
```
cabo, line 555:
```suggestion

```

glyn: Er, is this suggestion still needed?

glyn: Oh I see.

glyn (approved):
LGTM, thanks.

PR #346, opened 2023-01-10
Close #272

### LINKED ISSUE #272: Unicode escapes in Name Selector
In 3.5.1, Name Selector, it discusses the \\uXXXX escape. Is this to be interpreted exactly as in JSON, i.e. if I want to refer to a character outside the Basic Multilingual Plane, I have to compose two UTF-16 codepoints? Or might I be able to say \\u1F600 for an emoji smiley?

The spec should be specific about this.

cabo:
Add pair of surrogate escapes to note on string format.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -725,7 +725,7 @@ HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 Note: `double-quoted` strings follow the JSON string syntax ({{Section 7 of RFC8259}});
 `single-quoted` strings follow an analogous pattern ({{syntax-index}}).
 No attempt was made to improve on this syntax, so characters with
-scalar values above 0x10000 (such as <u format="num-lit-name">🤔</u>) need to be represented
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -724,6 +724,9 @@ HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 
 Note: `double-quoted` strings follow the JSON string syntax ({{Section 7 of RFC8259}});
 `single-quoted` strings follow an analogous pattern ({{syntax-index}}).
+No attempt was made to improve on this syntax, so characters with
+scalar values above 0x10000 (such as <u format="num-lit-name">🤔</u>) need to be represented
```
glyn, line 728: I'm not keen on the nested parentheses in the rendered output.
```suggestion
scalar values above 0x10000, such as <u format="num-lit-name">🤔</u>, need to be represented
```

glyn (approved):
LGTM, thanks. Just one editorial suggestion.

Issue #344, opened 2023-01-10
Perform a full grammar review and consolidation


cabo:
E.g., <https://mailarchive.ietf.org/arch/msg/jsonpath/ryGaOjcEN8MKVGuxwk4N0LbgTlI>

goessner:
I propose a slightly more compact naming of ABNF of `child`- and `descendant`-segment, which should be more consistent between them two.

### Child Segment
```
child-segment            =  (child /
                                       dot-wildcard /
                                       dot-member-name)
child                           = "[" S selector 1*(S "," S selector) S "]"
dot-wildcard               = "." wildcard
dot-member-name     = "." member-name
member-name            = name-first *name-char
...
```

### Descendant Segment
```
descendant-segment  = (descendant /
                                       ddot-wildcard /
                                       ddot-member-name)
descendant                 = ".." child-segment
ddot-wildcard             = ".." wildcard
ddot-member-name   = ".." member-name
```

`ddot` is well known from LaTeX and means `double-dot` or `dotdot` there.

goessner:
```descendant           = ".." child-segment```

above is wrong (and so it is in the current spec). It needs to read ...

```descendant           = ".." child```

to not allow `...*` triple dot (`dddot` :-)

**Comment by glyn** [comment]
> ddot-wildcard /

@cabo Wouldn't dotdot-wildcard etc. be less cryptic?

**Comment by glyn** [comment]
Here are some quibbles with the current ABNF which we can sort out after The Great Renaming:
```
json-path = root-identifier segments
segments  = *(S segment)


root-identifier  = "$"


selector =  ( name-selector  /
              index-selector /
              slice-selector /
              filter-selector
            )


name-selector       = string-literal

string-literal      = %x22 *double-quoted %x22 /     ; "string" ;;; QUESTION: is an empty string permitted?
                      %x27 *single-quoted %x27       ; 'string' ;;; QUESTION: is an empty string permitted?

double-quoted       = unescaped /
                      %x27      /                    ; '
                      ESC %x22  /                    ; \"
                      ESC escapable

single-quoted       = unescaped /
                      %x22      /                    ; "
                      ESC %x27  /                    ; \'
                      ESC escapable

ESC                 = %x5C                           ; \  backslash

unescaped           = %x20-21 /                      ; s. RFC 8259 ;;; QUESTION: What does "s." mean?
[...]

PR #351, opened 2023-01-10
Close #345

### LINKED ISSUE #345: Clarify behaviour of existence test with OptionalBoolean function
Currently, the existence test description contradicts the intended use of OptionalBoolean functions. We need to be careful if we want to make a special case for OptionalBoolean functions and not open the door to any boolean value being treated similarly.

cabo:
Add the necessary internal references.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1069,10 +1069,14 @@ the current node is selected if and only if the expression yields true.
 
 The current node is accessible via the current node identifier `@`.
 
-An existence expression either tests the existence of a node
-designated by an embedded query (see {{extest}}) or tests whether the
-result of a function expression of type `Boolean` or `OptionalBoolean`
-(see {{fnex}}) is `true`.
+A test expression either tests the existence of a node
+designated by an embedded query (see {{extest}}) or tests the
+result of a function expression (see {{fnex}}).
+In the latter case, if the function expression is of type
+`OptionalBoolean` or one of its subtypes, it tests whether the result
+is `true`; if the function expression is of type `OptionalNodes` or
+one of its subtypes, it tests whether the result is different from
+`Nothing`.
 
 ~~~~ abnf
 boolean-expr      = logical-or-expr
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1069,7 +1069,10 @@ the current node is selected if and only if the expression yields true.
 
 The current node is accessible via the current node identifier `@`.
 
-An existence expression may test the result of a function expression (see {{fnex}}).
+An existence expression either tests the existence of a node
+designated by an embedded query (see {{extest}}) or tests whether the
+result of a function expression of type `Boolean` or `OptionalBoolean`
+(see {{fnex}}) is `true`.
```
glyn, line 1076: The above sentence doesn't cover existence tests of function return values of type `OptionalNodes` or one of its subtypes. I think the solution to that is to beef up the extest section to cover such function return values.

cabo: Yes.  Will update PR later today.

draft-ietf-jsonpath-base.md
```diff
@@ -1069,7 +1069,10 @@ the current node is selected if and only if the expression yields true.
...
```
glyn, line 1075: I don't think of functions expressions of type `Boolean` as existence expressions. We test the return value, but existence is moot. I'd prefer to raise them to another kind of basic-expr.

`OptionalBoolean` function expressions involve both an existence test and, if the return value is not `Nothing`, consideration of the return value.

I wonder if we should try to tease this out in the spec?

glyn (changes requested):
A missing case and an improvement to consider.

draft-ietf-jsonpath-base.md
cabo: I'd like to fix the ABNF to be structured along more logical lines.  This is really #344; I think in this PR we should make sure we have all the material that we then reshuffle.

draft-ietf-jsonpath-base.md
cabo: ... and I resolved the conflict by merging main.

glyn (approved):
Ok, let's refactor the ABNF later.

PR #363, opened 2023-01-24
Make length return nothing when input is invalid

### LINKED ISSUE #361: Why does `length()` return 1 for unsupported inputs?
For `length()`, the spec defines 

> Its only argument is a value (possibly taken from a singular path as in the example above). The result also is a value, an unsigned integer.
> 
> - If the argument value is a string, the result is the number of Unicode scalar values in the string.
> - If the argument value is an array, the result is the number of elements in the array.
> - If the argument value is an object, the result is the number of members in the object.
> - For any other argument value, the result is one.

Why return 1?  That's a possible value for `length()` to return for valid inputs.  Wouldn't it make sense to return a value that _can't_ be returned by a valid input, like -1?

```jsonpath
$[?length(@.authors) == 1]
```

```json
[
  { "data": "no authors here" },
  { "authors": [ "a single author" ] }
]
```

Currently, this returns the both objects in the array, which is counter-intuitive.  The length of the `authors` property in `{"data": "no authors here" }` is nonsensical because it doesn't contain an `authors` property.

gregsdennis:
Resolves #361


gregsdennis:
hmm... didn't expect the other changes.  will close, update my local, and try again.

PR #365, opened 2023-01-24
`match()` and `search()` returns are not comparable

### LINKED ISSUE #360: The examples for `match()` conflate boolean values with boolean expresion results
For filter expressions, we have made a distinction between the JSON literals `true`/`false` and the result of a comparison (i.e. `$[?@.a]` is different from `$[?@.a == true]`).  However the examples for the `match()` function seem to conflict with this idea.

```jsonpath
$[?match(@.timezone, 'Europe/.*') == true]
```

The returns for `match()` and `search()` are described as

> Result: `OptionalBoolean` (`true`, `false`, or `Nothing`)

I think that this implies the latter would be valid, but the former would imply that the `match()` function can operate as an existence test akin to `$[?@.a]`, which doesn't make sense.  I don't think the function is testing whether a match exists, but whether the value is a match.

The only way I can see the former being valid is if the first argument allowed for a nodelist, but then that would make the latter feel weird (and [potentially only fail at runtime](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/359)).

It _could_ be defined so that:

- if a singular path is used for the first argument, then it MUST be compared to a literal
- if a plural (?) path is used for the first argument, then it MUST NOT be compared to a literal (it MUST be an existence test)

This would resolve the ambiguity and allow for parsing to handle things, but context-sensitive grammar is tricky.

gregsdennis:
Resolves #360

Also, I noticed a minor read-flow issue and reorganized a paragraph to match the other section.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1540,9 +1540,6 @@ The "search" function extension provides a way to check whether a
 given string contains a substring that matches a given regular
 expression, which is in {{-iregexp}} form.
 
-The result is `Nothing` if the first argument is not a string or
-the second argument is not a string conforming to {{-iregexp}}.
-
 ~~~ JSONPath
 $[?search(@.author, "[BR]ob")]
 ~~~
@@ -1551,6 +1548,10 @@ Its first argument is a string that is searched for at least one
 substring that matches the iregexp contained in the string
 that is the second argument.
 The result is `true` if such a substring exists, `false` otherwise.
+
+The result is `Nothing` if the first argument is not a string or
+the second argument is not a string conforming to {{-iregexp}}.
+
 
 ### Examples
 {: unnumbered}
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1563,7 +1564,7 @@ The result is `true` if such a substring exists, `false` otherwise.
 | `$[?count(1) == 1]` | Invalid typing since `1` is not a path  |
 | `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `OptionalNodes` and result type `OptionalNodes` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | Valid typing |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | Valid typing |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | Invalid typing since `OptionBoolean` is not a `comparable` |
```
glyn, line 1567: Good.

glyn (approved):
LGTM, thanks.

PR #369, opened 2023-01-31
Clarify omissions in ABNF ranges

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -698,9 +698,12 @@ single-quoted       = unescaped /
 ESC                 = %x5C                           ; \  backslash
 
 unescaped           = %x20-21 /                      ; see RFC 8259
-                      %x23-26 /                      ; omit "
-                      %x28-5B /                      ; omit '
-                      %x5D-10FFFF                    ; omit \
+                                                     ; omit "
```
cabo, line 701: How about `omit 0x22 "` or, if it must, `omit U+0022 "` etc.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: The former reads better IMO. Added. PTAL.

PR #370, opened 2023-01-31
Delete relation-expr

glyn:


gregsdennis (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1101,7 +1101,7 @@ logical-and-expr  = basic-expr *(S "&&" S basic-expr)
                       ; binds more tightly than disjunction
 
 basic-expr        = paren-expr /
-                    relation-expr /
+                    comp-expr /  ; comparison
```
cabo, line 1104: We might call it "comparison-expr" and get rid of the comment.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: I toyed with that. Adopted.

PR #373, opened 2023-01-31
Don't repeat function-expr in singular-path ABNF

glyn:


cabo (approved):
Interesting.  I thought I had already approved this somewhere else...

glyn:
> Interesting. I thought I had already approved this somewhere else...

Yeah, I think I may have screwed up a rebase. Apologies.

PR #376, opened 2023-01-31
Editorial improvements, also close #344

### LINKED ISSUE #344: Perform a full grammar review and consolidation

glyn:


glyn:
@cabo I noticed some ABNF lines were too long, so I added a second commit to this PR to shrink the lines. It's not pretty, so please see what you think.

draft-ietf-jsonpath-base.md
```diff
@@ -661,8 +664,8 @@ or children of either objects or arrays.
 ~~~~ abnf
 selector            = name-selector  /
                       wildcard-selector /
-                      slice-selector /
                       index-selector /
+                      slice-selector /
```
cabo, line 668: I would prefer to keep the original order so parsers based on PEG can parse slice selectors.

cabo: I don't think we have to do this.

About PEG:
https://en.wikipedia.org/wiki/Parsing_expression_grammar#Semantics
https://www.rfc-editor.org/rfc/rfc8610.html#appendix-A

glyn: Ah I see: putting `index-selector` before `slice-selector` could cause a PEG parser to regard a prefix of a slice (e.g. `1:4`) as an index (`e.g. `1`). (I guess it then depends on what happens whether the enclosing expression fails to parse. Will there be another attempt to parse the selector. I guess not.)

I accept that we don't *have* to reorder the sections. I have reverted the reordering in the third commit.

scripts/test-abnf-cts.rb
```diff
@@ -24,6 +24,13 @@
...
   "union array slice, overflowing step",
   "union array slice, underflowing step",
+  "index selector, overflowing index",
+  "slice selector, overflowing to value",
+  "slice selector, underflowing from value",
+  "slice selector, overflowing from value with negative step",
+  "slice selector, underflowing to value with negative step",
+  "slice selector, overflowing step",
+  "slice selector, underflowing step",
 ]
```
glyn, line 34: I'm not sure I follow what's going on here. Overflows/underflows should cause well formed expressions to fail parsing because they are invalid. Is the purpose of this Ruby file to exercise the ABNF alone or the spec more generally?

glyn: @cabo I get it now - I didn't see that the above tests are skipped until now.

cabo: Sorry for pushing this in the middle of your PR...  Opening up a new PR just for this seemed too much work.
The current tests indeed use the CTS (and JPC) only for checking the ABNF.

draft-ietf-jsonpath-base.md
```diff
@@ -391,6 +391,10 @@ of bracket notation. Examples and descriptions use shorthands where convenient.
 
 ### Selectors
 
+A name selector, e.g. `name`, selects a named child of an object.
```
gregsdennis, line 394: Does this need quotes (e.g. `'name'`)?  Might be confusing with the `3` that doesn't have them.

draft-ietf-jsonpath-base.md
```diff
@@ -832,25 +837,29 @@ The following examples show the `wildcard` selector in use by a child segment.
 
 The example above with the query `$.o[*, *]` shows that the wildcard selector may produce nodelists in distinct
 orders each time it appears in the child segment, when it is applied to an object node with two or more
-members (but not when it is applied to object nodes with less than two members or to array nodes).
+members (but not when it is applied to object nodes with fewer than two members or to array nodes).
```
gregsdennis, line 840: 👍🎉

draft-ietf-jsonpath-base.md
glyn: Nice catch! Fixed.

draft-ietf-jsonpath-base.md
```diff
@@ -2178,8 +2196,8 @@ JSONPath:
 * Square brackets in XPath expressions always operate on the *node
   set* resulting from the previous path fragment. Indices always start
   at 1.
-* With JSONPath, square brackets operate on the *object* or *array*
-  addressed by the previous path fragment. Array indices always start
+* With JSONPath, square brackets operate on the *nodelist*
```
cabo, line 2199: This (in particular with the array index note after it) is a bit misleading -- can we say that it operates on each of the nodes in the nodelist?

cabo (approved):
I submitted my own commit, so I like this much more now :-)
(But these were minor editorial anyway.)

draft-ietf-jsonpath-base.md
glyn: Yes. Adopted.

Issue #377, opened 2023-01-31
What is a JSONPath expression?


glyn:
Section 1.4 says:
>A JSONPath expression consists of an identifier followed by a series of zero or more segments each of which contains one or more selectors.

So this includes relative paths (starting with `@`).

However, section 1.4.1 contradicts this:
> Every JSONPath expression begins with the root node identifier.

Similarly, section 2.4 says:
>Every JSONPath query MUST begin with the root identifier `$`.

(Remember that, according to our terminology, "query" is a short name for a JSONPath expression.)

A possible solution is for sections 1.4.1 and 2.4 to talk about JSONPath expressions/queries which do not appear inside a filter expression.

PR #380, opened 2023-01-31
Clarify JSONPath expression syntax

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/377

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -363,7 +363,6 @@ of zero or more segments each of which contains one or more selectors.
 
 The root node identifier `$` refers to the root node of the argument,
 i.e., to the argument as a whole.
-Every JSONPath expression begins with the root node identifier.
 
 The current node identifier `@` refers to the current node in the context
 of the evaluation of a filter expression (described later).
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -622,7 +621,7 @@ of input node.
 ### Syntax
 {: unnumbered}
 
-Every JSONPath query MUST begin with the root identifier `$`.
+Every JSONPath query (except those inside filter expressions, see {{filter-selector}}) MUST begin with the root identifier `$`.
```
cabo, line 624: Do we call the things inside the filter expressions (filter-path/singular-path) "JSONPath query"?

glyn: Yes, I think we do. Section 1.4 says:
> A JSONPath expression consists of an identifier followed by a series of zero or more segments each of which contains one or more selectors.

and "query" is just a short name for "JSONPath expression".

cabo: Hmm, wouldn't it be useful to have a name for the thing that can be a JSONPath query outside a filter expression?

goessner: Hmm ... there is only a single JSONPath expression outside of filter expressions.

We also should obey our own terminology:

Query:   Short name for a JSONPath expression.

cabo (approved):

Issue #383, opened 2023-02-06
Should input to length function be of type OptionalValue instead of Value


glyn:
@cabo raised this in a PR review.

glyn:
I think the answer is "yes" because that allows the `length()` function to be used on a Singular Path rather than on just a literal.

PR #384, opened 2023-02-06
length should take an OptionalValue

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/383

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1471,7 +1471,7 @@ according to one of the following rules:
 ### `length` Function Extension {#length}
 
 Arguments:
-: 1. `Value`
+: 1. `OptionalValue`
 
 Result:
 : `OptionalValue` (unsigned integer or `Nothing`)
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1471,7 +1471,7 @@ according to one of the following rules:
 ### `length` Function Extension {#length}
```
cabo, line 1471: The text of this section still assumes that the argument is a value.  The list of cases works out, but this is still a bit inconsistent.

glyn: Hmmm. It seemed consistent with descriptions of other functions which take optional arguments.

cabo: let's go for it for now, we can take another pass.

cabo (approved):

draft-ietf-jsonpath-base.md
gregsdennis: > Its only argument is a value (possibly taken from a singular path as
> in the example above).  The result also is a value, an unsigned
> integer.

It uses "value" language for both input and output.  Is it safe to assume from 

> For any other argument value, the result is `Nothing`.

and typing it as `OptionalValue`, that `Nothing` is returned when `Nothing` is received?

glyn:
@cabo I pushed cb1cf1 to clarify optional arguments.

cabo (approved):
That should be enough to remove this double-take.

draft-ietf-jsonpath-base.md
glyn: I think that's pretty clear, yes. I've tried a clarification, but we can tweak later if necessary.

Issue #385, opened 2023-02-07
Table 13 does not show that OptionalNode is a subtype of OptionalValue


glyn:
`OptionalNode` is a subtype by coercion of `OptionalValue`, but Table 13 does not include this subtype relationship.

cabo:
Not sure I like the concept of "subtype by coercion". One can be used in place of the other as it can be coerced, but that is not usually what we'd call a subtype.

glyn:
That's a different point really. Table 13 already includes at least one subtype by coercion, so this issue adds another subtype by coercion.

glyn:
@cabo How would you like to proceed with "subtype by coercion"? If we want to explore alternatives, I think it deserves a separate, non-editorial issue.

gregsdennis:
From the spec, just below [Table 14](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-10.html#table-14): 

> `OptionalNode` is a subtype of `OptionalValue` via coercion since the `OptionalNode` instance `Node(n)` can be coerced to the `OptionalValue` instance `Value(v)`, where `v` is the value of the node `n`.

I disagree with the idea that coercion implies subtypes.

If I have integer and floating point numeric types, I can _cast_ from one to the other, but that doesn't mean that either is a subtype of the other.

C# provides "implicit" casts where the compiler can infer a type cast from context.  For example, 

```c#
float f1 = 1.5;
int i1 = (int) f1;  // C# requires that I explicitly cast this because of data loss

int i2 = 42;
float f2 = i;  // This is an implicit cast.  int isn't a subtype, but it is compatible.

I think this is closer to what we're dealing with here.  We're looking at compatibility and conversion between types, not polymorphism (as "subtype" implies).

PR #386, opened 2023-02-07
Add more subtypes to table 13

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/385

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1408,15 +1408,15 @@ is a subset of the set of instances of the other type.
 {{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, and `nl` denotes
 a non-empty nodelist. The table also lists the subtypes of each type.
 
-| Type                  | Abstract Instances                       | Subtypes                                 |
-| :--:                  | :----------------:                       | :------:                                 |
-| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNode`, `OptionalValue`, `Value` |
-| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                          |
-| `OptionalValue`       | `Value(v)`, `Nothing`                    | `Value`, `OptionalBoolean`               |
-| `Value`               | `Value(v)`                               | `Boolean`                                |
-| `OptionalBoolean`     | `Value(true)`, `Value(false)`, `Nothing` | `Boolean`                                |
-| `Boolean`             | `Value(true)`, `Value(false)`            |                                          |
-| `OptionalNodes`       | `Nodes(nl)`, `Nothing`                   | `OptionalNode`                           |
+| Type                  | Abstract Instances                       | Subtypes                                                               |
+| :--:                  | :----------------:                       | :------:                                                               |
+| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNode`, `OptionalValue`, `OptionalBoolean`, `Value`, `Boolean` |
+| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                                                        |
+| `OptionalValue`       | `Value(v)`, `Nothing`                    | `OptionalNode`, `Value`, `OptionalBoolean`, `Boolean`                  |
+| `Value`               | `Value(v)`                               | `Boolean`                                                              |
+| `OptionalBoolean`     | `Value(true)`, `Value(false)`, `Nothing` | `Boolean`                                                              |
+| `Boolean`             | `Value(true)`, `Value(false)`            |                                                                        |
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1408,15 +1408,15 @@ is a subset of the set of instances of the other type.
...
-| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                          |
-| `OptionalValue`       | `Value(v)`, `Nothing`                    | `Value`, `OptionalBoolean`               |
-| `Value`               | `Value(v)`                               | `Boolean`                                |
-| `OptionalBoolean`     | `Value(true)`, `Value(false)`, `Nothing` | `Boolean`                                |
-| `Boolean`             | `Value(true)`, `Value(false)`            |                                          |
-| `OptionalNodes`       | `Nodes(nl)`, `Nothing`                   | `OptionalNode`                           |
+| Type                  | Abstract Instances                       | Subtypes                                                               |
+| :--:                  | :----------------:                       | :------:                                                               |
+| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNode`, `OptionalValue`, `OptionalBoolean`, `Value`, `Boolean` |
```
cabo, line 1413: Of course, we can put the transitive closure into the table, but we also could simply say a subtype of a subtype is a subtype.

draft-ietf-jsonpath-base.md
```diff
@@ -1408,15 +1408,15 @@ is a subset of the set of instances of the other type.
...
+| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                                                        |
+| `OptionalValue`       | `Value(v)`, `Nothing`                    | `OptionalNode`, `Value`, `OptionalBoolean`, `Boolean`                  |
```
cabo, line 1415: see https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/385#issuecomment-1420298763

draft-ietf-jsonpath-base.md
glyn: I thought it was clearer to put the transitive closure into the table. I guess another option is to change the column title to "Immediate subtypes" (or similar) to try to avoid confusion. But since the transitive closure isn't massive, I don't think omitting the transitive closure from the table is worth the extra explanation.

We should, however, check that the spec says that a subtype of a subtype is a subtype!

cabo: To me those additional entries are clutter that make it harder to check and understand the table.
They are probably the reason I didn't see the subtype by coercion thing that I still don't like.

draft-ietf-jsonpath-base.md
```diff
@@ -1408,15 +1408,15 @@ is a subset of the set of instances of the other type.
...
-| :--:                  | :----------------:                       | :------:                                 |
-| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNode`, `OptionalValue`, `Value` |
-| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                          |
-| `OptionalValue`       | `Value(v)`, `Nothing`                    | `Value`, `OptionalBoolean`               |
-| `Value`               | `Value(v)`                               | `Boolean`                                |
-| `OptionalBoolean`     | `Value(true)`, `Value(false)`, `Nothing` | `Boolean`                                |
-| `Boolean`             | `Value(true)`, `Value(false)`            |                                          |
-| `OptionalNodes`       | `Nodes(nl)`, `Nothing`                   | `OptionalNode`                           |
+| Type                  | Abstract Instances                       | Subtypes                                                               |
+| :--:                  | :----------------:                       | :------:                                                               |
```
cabo, line 1412: BTW, is there a reason you want those centered?

glyn: No, it would be fine for them to be left-justified.

cabo:
```suggestion
| :--                  | :----------------                       | :------                                                               |
```

draft-ietf-jsonpath-base.md
glyn: Ok, I'll change the table to include only "direct" subtypes and ensure the text says that the subtype relationship is transitive.

glyn: @cabo Fixed. PTAL.

draft-ietf-jsonpath-base.md
glyn: @cabo Please can we treat the current PR as editorial, i.e. making the table consistent by including the subtypes by coercion. Then we can merge it, @gregsdennis can do the type renaming ("XxxType"), and we can work on subtypes by coercion to see what better alternatives there are in the context of the four recent issues.

cabo: Sure

cabo (approved):
(Modulo the coercion entry on OptionalValue, as discussed.)

PR #391, opened 2023-02-08
added 'Type' to all type names to distinguish from abstractions

### LINKED ISSUE #366: `Value` in Table 13 is confusing
If I'm reading the [table](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-09.html#tbl-types) right, it looks like the term `Value` is being used to mean different things depending on which column it's in.

- In column 1 ("Type") or column 3 ("Subtypes"), it's referring to the type system we're creating.
- In column 2 ("Abstract Instances"), it's referring to the underlying JSON(ish) value kind, i.e. `object`, `boolean`, `number`, etc, plus `Nothing`.

I think having the same word to define these things can be confusing for the reader, and maybe we need something else in one of these places to indicate the distinction.  We've already done this for the `Absent` type to distinguish it from the `Nothing` value.

gregsdennis:
Resolves #366 

This is literally just a renaming of the names in the Type column of Table 13.

I also noticed that the `count()` description doesn't account for non-nodelist inputs.  I'll open another issue for that.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1098,8 +1098,8 @@ A test expression either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
 result of a function expression (see {{fnex}}).
 In the latter case, if the function expression is of type
-`OptionalBoolean` or one of its subtypes, it tests whether the result
-is `true`; if the function expression is of type `OptionalNodes` or
+`OptionalBooleanType` or one of its subtypes, it tests whether the result
+is `true`; if the function expression is of type `OptionalNodesType` or
 one of its subtypes, it tests whether the result is different from
 `Nothing`.
 
@@ -1117,9 +1117,9 @@ basic-expr          = paren-expr /
                       test-expr
 test-expr           = [logical-not-op S] filter-path /
                          ; path existence or non-existence
-                      function-expr ; OptionalBoolean or subtype
+                      function-expr ; OptionalBooleanType or subtype
 filter-path         = rel-path / json-path /
-                      function-expr ; OptionalNodes or subtype
+                      function-expr ; OptionalNodesType or subtype
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
 ~~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1411,25 +1411,25 @@ a non-empty nodelist. The table also lists the (immediate) subtypes of each type
...
 | Type                  | Abstract Instances                       | Subtypes                                   |
 | :--                   | :----------------                        | :------                                    |
-| `OptionalNodeOrValue` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNode`, `OptionalValue`            |
-| `OptionalNode`        | `Node(n)`, `Nothing`                     |                                            |
-| `OptionalValue`       | `Value(v)`, `Nothing`                    | `OptionalNode`, `Value`, `OptionalBoolean` |
-| `Value`               | `Value(v)`                               | `Boolean`                                  |
-| `OptionalBoolean`     | `Value(true)`, `Value(false)`, `Nothing` | `Boolean`                                  |
-| `Boolean`             | `Value(true)`, `Value(false)`            |                                            |
-| `OptionalNodes`       | `Nodes(nl)`, `Nothing`                   | `OptionalNode`                             |
+| `OptionalNodeTypeOrValueType` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNodeType`, `OptionalValueType`            |
```
glyn, line 1414: I wonder if the double "type" is necessary as it doesn't conform to the pattern of adding "Type" to the end of each type name.
```suggestion
| `OptionalNodeOrValueType` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNodeType`, `OptionalValueType`            |
```

gregsdennis: Yeah, that's probably an artifact of replacing strings.  I'll fix.

gregsdennis: ✔️

cabo (approved):
Thanks!

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: already fixed.

Issue #392, opened 2023-02-08
`count()` description doesn't account for non-nodelist inputs


gregsdennis:
All of the other functions we've defined account for input types that they don't expect, and they return `Nothing` in those cases.  We should update `count()` to do so as well.

cabo:
count() has declared type OptionalNodesType for its Argument 1.
How would it get anything but that?

gregsdennis:
`OptionalNodesType` includes `Nothing` as a potential value.

cabo:
Yes. As in Lisp, there is a weird alignment between () and nil...
I'm not even sure we need NodesType to have an "Optional" aspect.
But if we do, I agree we have to say what count makes out of Nothing (either 0 or Nothing), and shape the return type accordingly.

glyn:
PR  #393 fixes this issue in terms of the current type system in which `Nodes(nl)` is represented by a _non-empty_ nodelist and `Nothing` is represented by a( path producing a)n empty nodelist.

PR #393, opened 2023-02-09
Define count(Nothing)

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/392

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1520,6 +1520,9 @@ The result is a value, an unsigned integer, that gives the number of
 nodes in the nodelist.
 Note that there is no deduplication of the nodelist.
 
+If the argument is `Nothing` (which corresponds to an empty nodelist),
+the result is the value `0`.
+
 
 ### `match` Function Extension {#match}
 


draft-ietf-jsonpath-base.md
```diff
@@ -1520,6 +1520,9 @@ The result is a value, an unsigned integer, that gives the number of
 nodes in the nodelist.
 Note that there is no deduplication of the nodelist.
 
+If the argument is `Nothing` (which corresponds to an empty nodelist),
+the result is the value `0`.
```
gregsdennis, line 1524: Should this instead return `Nothing`?  We don't have any other cases where `Nothing` begets non-`Nothing`.  (You can't get something from nothing. 😆)

Also, to @cabo's point [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/392#issuecomment-1423289172), do we instead want to say that the input is just a nodelist (`NodesType`)?  I'm having trouble thinking of a scenario where it could actually get nothing.

glyn: `count`'s return type is `ValueType`, so `Nothing` would be invalid, but I suppose you are suggesting changing the return type to `OptionalValueType`.

However, `count` really should be dealing with nodelists (or functions that return the moral equivalent of a nodelist) and `Nothing` is simply an abstraction of an empty nodelist, so returning `0` seems consistent to me.

gregsdennis: Sorry for the multiple conversations/option going here. I'll try to keep them separate.

---

> I suppose you are suggesting changing the return type to OptionalValueType.

Yes.  I'm suggesting this.

---

> `Nothing` is simply an abstraction of an empty nodelist

I'm not sure I agree with this.  So far, any expressions we have (except those which contain `count()`) that contains a `Nothing` currently evaluate to `Nothing`, which _yields_ an empty nodelist. It yields an empty nodelist because it is not evaluatable, so no nodes are selected.  But I wouldn't say one is an abstraction of the other.

glyn: The abstract vs concrete distinction is deliberate. The types and abstract instances are just that: abstract. They do not appear in code and they are not evaluated. Their purpose is to define the intended behaviour of the spec.

(Of course, implementations need to represent these abstractions somehow and are free to use algebraic data types, enumerated types, or such like that mirror the types and abstract instances in the spec so long as the resultant behaviour is equivalent.)

The spec describes a relationship between types with their abstract instances and concrete representations. It is the concrete representations that appear in code and which are evaluated.

Does that help?

gregsdennis: No, it doesn't help. You're claiming that `Nothing` is an abstraction of the empty nodelist, but the spec doesn't say that, and I think it's incorrect.

`Nothing` was our placeholder for "no value." An empty nodelist is the output of a query that selected no nodes.   They're not the same thing.  They're not related.

glyn: > No, it doesn't help. You're claiming that `Nothing` is an abstraction of the empty nodelist, but the spec doesn't say that, and I think it's incorrect.

Er, the spec does say that. Table 14 says the abstract instance `Nothing` has concrete representations of a Singular Path or `filter-path` resulting in an empty nodelist.

> `Nothing` was our placeholder for "no value." An empty nodelist is the output of a query that selected no nodes. They're not the same thing. They're not related.

`Nothing` (initially named `Absent`) was inspired by earlier discussion of empty nodelists in the WG. [issue 123](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/123) (Is absent equal to absent?) gave rise to text describing [comparisons involving empty nodelists](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/207/files#diff-aca3bb236d11f768483f8a1153cb9030b67d9cd6a39f012e8e4bfec2c6275165R1227). That's why we chose those concrete representations in Table 14.

cabo: (Just checking if this alternative world is maybe more acceptable for someone who did not evolve Lactose tolerance by continually being fed with Lisp.)

gregsdennis: Table 13 defines `OptionalNodesType` as a nodelist or `Nothing`.  Why have this type at all if `Nothing` is just the empty nodelist?

If `Nothing` is the same as the empty nodelist, then `Nothing` is already a member of `Nodes(nl)`, and we should just have a `NodesType` for nodelists that's represented by `Nodes(nl)` without any mention of `Nothing`.

cabo: > Table 13 defines `OptionalNodesType` as a nodelist or `Nothing`. Why have this type at all if `Nothing` is just the empty nodelist?
> 
> If `Nothing` is the same as the empty nodelist, then `Nothing` is already a member of `Nodes(nl)`, and we should just have a `NodesType` for nodelists that's represented by `Nodes(nl)` without any mention of `Nothing`.

Actually, `Nodes(nl)` is defined as a non-empty nodelist.

```
| `Nodes(nl)`       | `filter-path` resulting in the non-empty nodelist `nl`             |
```

So Nothing is needed to complement this to a Type that includes all nodelists `OptionalNodeType`.
Not sure that this little piece of gymnastic really helps a lot, except that it underscores the Nothingness of the empty list...

gregsdennis: Okay.  The distinction is found just above the row in Table 14 that you referenced:

> | `Nothing` | Singular Path or filter-path resulting in an empty nodelist |

`Nothing` _results_ in an empty nodelist, but it is not itself an empty nodelist.  You're saying that this means `Nothing` is _representative_ of an empty nodelist.

I could be okay with this if the text (above, in the PR) said something more along the lines of 

> If the argument is `Nothing` (which can represent an empty nodelist), the result is the value `0`.

But I'm still a little bothered that we're going from `Nothing` to a value.  This doesn't happen anywhere else, not even the logical NOT operator `!` does this.  (It reinterprets `Nothing` as TestBoolean-false and inverts that to TestBoolean-true.)

cabo (approved):

draft-ietf-jsonpath-base.md
goessner: I would be very happy, if we could get rid of that 'Nothing' thing somehow.

Issue #387, opened 2023-02-07
`Boolean` and `Value` (and their optional variants) should be distinct


gregsdennis:
I think this relates to #366.

An expression boolean (e.g. the result of `@.a==2`) is not a JSON value.  This is something that we decided long ago.  However Table 13 describes it as "`Value(true)` or `Value(false)`".  This is wrong.

If we allow `OptionalBoolean` to be a subtype of `OptionalValue`, then paths like `$[?length(1)]` and `$[?match(@.a,@.b)=="string"]` will be valid and produce no parsing errors.  In #360 (see also #365), we decided that this wouldn't be the case. `match()` only operates as a boolean, and `length()` only operates as a value.

`OptionalBoolean` and `OptionalValue` MUST be distinct in order for this to work.

### Edit

This issue is about the _role_ a function plays within an expression.  That role is determined by its return value.

cabo:
Since we don't allow `comparison` as a function argument, I don't think `1==2` *has* a type.

gregsdennis:
I don't see how that's relevant.  I'm not talking about argument types.  I'm talking about return types and the role of functions within expressions.

`match()` returns an expression boolean (not a JSON boolean), which is valid where booleans are valid in expressions, i.e. _not_ in comparisons.

`length()` returns a JSON-like value, which is valid where values are valid in expressions, i.e. _only_ in comparisons.

gregsdennis:
This distinction is implied because of the decisions that we've already made

- in the issue and PR linked above
- that we aren't going to support type coercion

gregsdennis:
From [Section 2.6.2](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-10.html#name-type-correctness-of-functio):

> - If it occurs as a `filter-path` in a test expression, the function is defined to have result type `OptionalNodes` or one of its subtypes, or to have result type `OptionalBoolean` or one of its subtypes.
> - If it occurs as a comparable in a comparison, the function is defined to have result type `OptionalNodeOrValue` or one of its subtypes.

If we allow `OptionalBoolean` to be a subtype of `OptionalNodeOrValue`, then the above decisions are contradicted.

gregsdennis:
In the interim we discussed that a function can return one of three kinds of data:

- a value (e.g. `length()`)
- a boolean (e.g. `match()`)
- a nodelist (e.g. a hypothetical `distinct()` function)

### Value

[...]

PR #394, opened 2023-02-09
Close #388: Clean up grammar around function-expr

### LINKED ISSUE #388: Ambiguity of parsing through comparable and filter-path

cabo:
(No semantic change)

This removes the weird diamonds we have been tangled in.


draft-ietf-jsonpath-base.md
```diff
@@ -1115,11 +1115,10 @@ logical-and-expr    = basic-expr *(S "&&" S basic-expr)
...
                       comparison-expr /
                       test-expr
-test-expr           = [logical-not-op S] filter-path /
-                         ; path existence or non-existence
-                      function-expr ; OptionalBooleanType or subtype
-filter-path         = rel-path / json-path /
-                      function-expr ; OptionalNodesType or subtype
+test-expr           = [logical-not-op S]
+                      (filter-path / ; path existence or non-existence
+                       function-expr) ; OptionalBooleanType or subtype
```
glyn, line 1120: The above function-expr is now the only one in test-expr, so it's not limited to OptionalBooleanType or subtype. It needs to include OptionalNodesType or subtype too.

Approving as I expect these comments to reworked soon.

glyn (approved):
Please adjust ABNF comment.

draft-ietf-jsonpath-base.md
cabo: I just added the notation ST(T), which makes it easier to compactly speak about these things.
Also, some editorial cleanup.

draft-ietf-jsonpath-base.md
```diff
@@ -1456,19 +1459,19 @@ The type correctness of function expressions can now be defined in terms of this
 
 A function expression is correctly typed if all the following are true:
 
-* If it occurs as a `filter-path` in a test expression, the function
-is defined to have result type `OptionalNodesType` or one of its subtypes,
-or to have result type `OptionalBooleanType` or one of its subtypes.
+* If it occurs directly in a test expression, the function
+is defined to have a result type out of `ST(OptionalNodesType)`.
+or to have a result type out of `ST(OptionalBooleanType)`.
```
gregsdennis, line 1464: I think "out" is redundant here.  "Result" implies output.

draft-ietf-jsonpath-base.md
```diff
@@ -1456,19 +1459,19 @@ The type correctness of function expressions can now be defined in terms of this
...
 * If it occurs as a `comparable` in a comparison, the function
-is defined to have result type `OptionalNodeOrValueType` or one of its subtypes.
+is defined to have a result type out of `ST(OptionalNodeOrValueType)`.
```
gregsdennis, line 1466: "out" here as well.

draft-ietf-jsonpath-base.md
gregsdennis, line 1464: > or to have a result type out of `ST(OptionalBooleanType)`.

This is the contradiction.  A boolean cannot rightly occur in a test expression.  The grammar and the type system misalign here.

cabo: "out of" as in "two out of three".  Do you have a better way to say this?

cabo: I wasn't trying to change the type system here, just to make the grammar more transparent.
Again, the grammar is not defining the type system (these bullet points are trying to), but what can actually be written.  We don't allow literals to be notated in a test-expr, outside of a comparison-expr or the arguments for a function-expo.  That doesn't mean all the values that these literals can have are invalid as a function return that goes into a test-expr.

gregsdennis: I see what you're getting at now.  I would say "defined by".

> the function has a result type defined by `ST(OptionalNodesType)`

with `ST(T)` being the set of the type T and all of its subtypes (as you've defined).

cabo: Hmm, "take a lolly defined by this jar" doesn't work for me, and I think most people wouldn't think that a set "defines" its members.  But maybe "in" is better than "out of"...

gregsdennis: > the grammar is not defining the type system

I'm not saying that one has to define the other.  I'm saying that what they each define needs to align, and right now they don't.

> That doesn't mean all the values that these literals can have are invalid as a function return that goes into a test-expr.

Maybe this isn't coming across right, but I think you're saying that because `match()` can potentially return `Nothing` that it should be allowed to act as `test-expr`.  Is that what you're saying?

I disagree with this (as mentioned in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/387#issuecomment-1424866809): 

> Because `match()` has the _potential_ to return these JSON values, the function MUST be restricted to locations where those values are valid, namely `comparable`.

draft-ietf-jsonpath-base.md
```diff
@@ -1456,19 +1459,19 @@ The type correctness of function expressions can now be defined in terms of this
 
 A function expression is correctly typed if all the following are true:
 
-* If it occurs as a `filter-path` in a test expression, the function
-is defined to have result type `OptionalNodesType` or one of its subtypes,
-or to have result type `OptionalBooleanType` or one of its subtypes.
+* If it occurs directly in a test expression, the function
+is defined to have a result type in `ST(OptionalNodesType)`.
+or to have a result type in `ST(OptionalBooleanType)`.
```
gregsdennis, line 1464: I still disagree with this, but as this is editorial, we'll continue.

gregsdennis (approved):

draft-ietf-jsonpath-base.md
cabo: > I'm saying that what they each define needs to align

I am not aware of a reason for this.
(I'm hearing you loud and clear that you *want* them to align, and I don't agree.)

Since this discussion really is one of the other issues, I'll merge this PR now, and we can continue the discussion there.

PR #395, opened 2023-02-14
Remove `BooleanType` as a type for consistent expression grammar

gregsdennis:
This is an option for fixing the grammar issues explained in #387 around functions and their role in expressions.

- Removes the `BooleanType` type and makes `match()` and `search()` merely produce `OptionalValueType`.  That value is either JSON `true` or `false`.
- Removes the special handling for functions that can return a JSON `true` or `false` result.

A fallout of this is that `$[?match(@.timezone, 'Europe/.*')]` is not valid/well-formed.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1095,13 +1095,9 @@ filter-selectors enclosing the filter-selector that is directly
 enclosing the identifier).
 
 A test expression either tests the existence of a node
-designated by an embedded query (see {{extest}}) or tests the
-result of a function expression (see {{fnex}}).
-In the latter case, if the function expression is of type
-`ST(OptionalBooleanType)` (see {{typesys}}), it tests whether the result
-is `true`; if the function expression is of type
-`ST(OptionalNodesType)`, it tests whether the result is different from
-`Nothing`.
+designated by an embedded query (see {{extest}}) or function expression
+(see {{fnex}}).  A function is only valid as a test expression if
+it is of the type `ST(OptionalNodesType)`.
 
 ~~~~ abnf
 boolean-expr        = logical-or-expr
@@ -1117,8 +1113,7 @@ basic-expr          = paren-expr /
                       test-expr
 test-expr           = [logical-not-op S]
                       (filter-path / ; path existence or non-existence
-                       function-expr) ; ST(OptionalBooleanType) or
-                                      ; ST(OptionalNodesType)
+                       function-expr) ; ST(OptionalNodesType)
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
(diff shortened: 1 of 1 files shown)


glyn (commented):
I could live with this because it's similar to where we were before the special treatment of functions returning `ST(OptionalBoolean)` was introduced. However, I think the usability of having to compare the result of `match` and `search` is poor and I prefer the explicit approach of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/396 overall.

gregsdennis:
> I think the usability of having to compare the result of `match` and `search` is poor

I agree that it's more clunky, but returning a value (thereby requiring the comparison) does have the advantage of providing support for `$[?match(@.a, 'a.*')==@.b]`.

I'm not tied to it though.

cabo:
The current status of the document has solved the clunkiness.
I still don't know why we need this regression.

gregsdennis:
> The current status of the document has solved the clunkiness.

No.  I will not debate this here.  I've made the case perfectly clear in #387 over multiple comments that the current solution is subpar at best.  _The document is wrong._

cabo:
Well, it seems the document is not describing the solution very well.  Let's work on that.

gregsdennis:
> Well, it seems the document is not describing the solution very well. Let's work on that.

Make a proposal, but the solution is wrong.  The document presents the solution.  And that's why the document is wrong.

draft-ietf-jsonpath-base.md
```diff
@@ -1095,13 +1095,9 @@ filter-selectors enclosing the filter-selector that is directly
...
-designated by an embedded query (see {{extest}}) or tests the
-result of a function expression (see {{fnex}}).
-In the latter case, if the function expression is of type
-`ST(OptionalBooleanType)` (see {{typesys}}), it tests whether the result
-is `true`; if the function expression is of type
-`ST(OptionalNodesType)`, it tests whether the result is different from
-`Nothing`.
+designated by an embedded query (see {{extest}}) or function expression
+(see {{fnex}}).  A function is only valid as a test expression if
+it is of the type `ST(OptionalNodesType)`.
```
cabo, line 1100: This is not a good match for match() and search().

cabo (changes requested):

draft-ietf-jsonpath-base.md
gregsdennis: Yes it is.  It follows perfectly from the return value of `match()` and `search()`: a JSON `true` or `false`, or `Nothing`.  I'm just saying that such a value cannot be used in a `test-expr`... because it's meaningless there.

cabo:
> > Well, it seems the document is not describing the solution very well. Let's work on that.
> 
> Make a proposal, but the solution is wrong. The document presents the solution. And that's why the document is wrong.

You have worked yourself into this fervor.  I'm going to stop commenting on this issue.

cabo:
This is superseded by #403.

PR #396, opened 2023-02-14
`BooleanType` should be logical, not a value, for consistent expression grammar

### LINKED ISSUE #387: `Boolean` and `Value` (and their optional variants) should be distinct

gregsdennis:
This is another option to resolve #387 (see #395 for the first option).

For this, `BooleanType` is not a value but represents the logical result of `test-expr`.

- Provides for functions that can return a true or false result by declaring that such a result is logical, not JSON.
- Removes `OptionalBooleanType` is no longer necessary, as `Nothing` in a `test-expr` behaves the same as logical _false_.

A fallout of this is that `$[?match(@.timezone, 'Europe/.*')==false]` is not valid/well-formed.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1098,7 +1098,7 @@ A test expression either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
 result of a function expression (see {{fnex}}).
 In the latter case, if the function expression is of type
-`ST(BooleanType)` (see {{typesys}}), it tests whether the result
+`ST(PropositionType)` (see {{typesys}}), it tests whether the result
 is a logical "true;" if the function expression is of type
 `ST(OptionalNodesType)`, it tests whether the result is different from
 `Nothing`.
@@ -1117,7 +1117,7 @@ basic-expr          = paren-expr /
                       test-expr
 test-expr           = [logical-not-op S]
                       (filter-path / ; path existence or non-existence
-                       function-expr) ; ST(BooleanType) or
+                       function-expr) ; ST(PropositionType) or
                                       ; ST(OptionalNodesType)
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
@@ -1418,8 +1418,8 @@ a non-empty nodelist. The table also lists the (immediate) subtypes of each type
 | `OptionalNodeOrValueType` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNodeType`, `OptionalValueType`        |
 | `OptionalNodeType`        | `Node(n)`, `Nothing`                     |                                                |
 | `OptionalValueType`       | `Value(v)`, `Nothing`                    | `OptionalNodeType`, `ValueType`                |
-| `ValueType`               | `Value(v)`                               | `BooleanType`                                  |
-| `BooleanType`             | `Logic`                                  |                                                |
+| `ValueType`               | `Value(v)`                               |                                                |
+| `PropositionType`         | `PropositionTrue`, `PropositionFalse`    |                                                |
 | `OptionalNodesType`       | `Nodes(nl)`, `Nothing`                   | `OptionalNodeType`                             |
 {: #tbl-types title="Function extension type system"}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1417,10 +1417,9 @@ a non-empty nodelist. The table also lists the (immediate) subtypes of each type
 | :--                       | :----------------                        | :------                                        |
 | `OptionalNodeOrValueType` | `Node(n)`, `Value(v)`, `Nothing`         | `OptionalNodeType`, `OptionalValueType`        |
 | `OptionalNodeType`        | `Node(n)`, `Nothing`                     |                                                |
-| `OptionalValueType`       | `Value(v)`, `Nothing`                    | `OptionalNodeType`, `ValueType`, `OptionalBooleanType` |
+| `OptionalValueType`       | `Value(v)`, `Nothing`                    | `OptionalNodeType`, `ValueType`                |
 | `ValueType`               | `Value(v)`                               | `BooleanType`                                  |
-| `OptionalBooleanType`     | `Value(true)`, `Value(false)`, `Nothing` | `BooleanType`                                  |
-| `BooleanType`             | `Value(true)`, `Value(false)`            |                                                |
+| `BooleanType`             | `Logic`                                  |                                                |
```
glyn, line 1422: 1. `BooleanType` should be renamed to `PropositionType` to further distance the term from the JSON `true` and `false` values which people (and RFC 8259, in the introduction) sometimes refer to as booleans. We should also rename `boolean-expr` to `proposition` in the ABNF to emphasise the correspondence.

2. There need to be two abstract instances rather than one. If we adopt the renaming above, the abstract instances could be  `PropositionTrue` and `PropositionFalse`.

glyn (changes requested):
I like this direction as it gets rid of the apparent special case and thereby makes the spec easier to read. But I made some suggestions which I think will really help. (There are a few editorial changes to make too, but let's agree on the overall approach first.)

draft-ietf-jsonpath-base.md
gregsdennis: I'm happy with this change. I can do that shortly.

Is there evidence of the term "proposition" used in this way elsewhere? Will this term be understood?

glyn: The term is often used in the context of proposition logic (or [propositional calculus](https://en.wikipedia.org/wiki/Propositional_calculus)), but it's also quite common in CompSci. The term won't necessarily be familiar to all programmers, but at least from their perspective, "proposition" would then be just an unfamiliar but distinct term in the spec.

cabo: We could call it LogicalFalse.  But we don't need it!

gregsdennis:
> There are a few editorial changes to make too

Yeah, I noticed few unrelated editorials I'd like to rework as well.

draft-ietf-jsonpath-base.md
```diff
@@ -1590,7 +1585,7 @@ the second argument is not a string conforming to {{-iregexp}}.
 | `$[?count(1) == 1]` | Invalid typing since `1` is not a path  |
 | `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `OptionalNodesType` and result type `OptionalNodesType` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | Valid typing |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | Invalid typing since `OptionalBooleanType` is not a `comparable` |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | Invalid typing since `PropositionType` can only appear in `test-expr` |
```
cabo, line 1588: I think that this is a worse outcome than the current solution.

cabo (changes requested):

draft-ietf-jsonpath-base.md
gregsdennis: The current solution is inconsistent and contradictory.  Literally _any_ solution that is consistent is better.

gregsdennis:
Closing this in favor of ☝️ as it's more complete.

PR #397, opened 2023-02-15
Fix match example

glyn:
The typing is valid according to the validity rules for functions.


cabo (approved):

cabo:
The document was indeed wrong...

PR #402, opened 2023-02-19
changed type system to just three types

### LINKED ISSUE #387: `Boolean` and `Value` (and their optional variants) should be distinct

gregsdennis:
Resolves #387 (specifcally implements https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/387#issuecomment-1436079071)

This is effectively an extension of #396 (though not a continuation of that git branch).  I purposely didn't use `PropositionType` because earlier in the text (under `#filter-selector`), it says

> A boolean expression, usually involving the current node, is evaluated and the current node is selected if and only if the expression yields true.

Since the "boolean" language is here, I just stuck with it.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1102,11 +1102,10 @@ enclosing the identifier).
 A test expression either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
 result of a function expression (see {{fnex}}).
-In the latter case, if the function expression is of type
-`ST(OptionalBooleanType)` (see {{typesys}}), it tests whether the result
-is `true`; if the function expression is of type
-`ST(OptionalNodesType)`, it tests whether the result is different from
-`Nothing`.
+In the latter case, the function expression MUST be of type
+`BooleanType` (see {{typesys}}) or of type `NodelistType` to appear
+in a test expression.  The expression tests whether the result
+is or can be converted to `PropositionTrue` (see {{type-conv}}).
 
 ~~~~ abnf
 boolean-expr        = logical-or-expr
@@ -1122,8 +1121,8 @@ basic-expr          = paren-expr /
                       test-expr
 test-expr           = [logical-not-op S]
                       (filter-path / ; path existence or non-existence
-                       function-expr) ; ST(OptionalBooleanType) or
-                                      ; ST(OptionalNodesType)
+                       function-expr) ; BooleanType or
+                                      ; NodelistType
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1140,8 +1139,8 @@ logical-not-op      = "!"               ; logical NOT operator
 Comparisons are between primitive values (that is, numbers, strings, `true`, `false`,
 and `null`),
 Singular Paths, each of which selects at most one node, and
-function expressions (see {{fnex}}) that
-return a primitive value or at most one node.
+function expressions (see {{fnex}}) of type `ValueType` or
+`NodelistType` (see {{type-conv}}).
 
 ~~~~ abnf
 comparison-expr     = comparable S comparison-op S comparable
@@ -1149,7 +1148,7 @@ literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
                       singular-path /   ; Singular Path value
-                      function-expr  ; ST(OptionalNodeOrValueType)
+                      function-expr  ; ValueType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
                       "<"  / ">"
@@ -1322,7 +1321,7 @@ The second set of examples shows some complete JSONPath queries that make use
 of filter selectors, and the results of evaluating these queries on a
 given JSON value as input.
 (Note that two of the queries employ function extensions; please see
-Sections {{<match}} and {{<search}} below for details about these.)
+Sections {{match}} and {{search}} below for details about these.)
 
 JSON:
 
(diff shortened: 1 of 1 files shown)


cabo:
Thanks.  As a general observation, I prefer to use a different name for JSONPath-level logicals from JSON-level Booleans.  In my draft, I have settled for "logical" for this.

Have you checked what functionality is lost with this approach?

gregsdennis:
> Thanks. As a general observation, I prefer to use a different name for JSONPath-level logicals from JSON-level Booleans. In my draft, I have settled for "logical" for this.

I'm not hung up on naming things.

> Have you checked what functionality is lost with this approach?

Because `match()` and `search()` MUST be used as "logicals," they can't be used as `comparable`s.  While you get `$[?match(@.a, 'a.*')]` and `$[?!match(@.a, 'a.*')]`, you don't get `$[?match(@.a, 'a.*') == @.b]`.  But I think that has a very niche use case that I'm sure could be covered with a custom function.

cabo:
I agree that `$[?match(@.a, 'a.*') == @.b]` can be expressed without Booleanness:
`$[?(match(@.a, 'a.*') && true == @.b) || (!match(@.a, 'a.*') && false == @.b)]`

cabo:
I wrote the 44 lines of code needed to check well-typedness in this type system (with the four predefined functions) and things do look good to me.  More testing and word-smithing required...

Sample (error messages precede the example; the right column is a JSON representation of the query):

```
$[?match(@.a, "x*")]                                                           ["$", ["filt", ["func", "match", ["@", "a"], "x*"]]]
*** Cannot use ["func", "match", ["@", "a"], "a.*"] with declared_type logical for required type value in comparable
$[?match(@.a, 'a.*') == @.b]                                                   ["$", ["filt", ["==", ["func", "match", ["@", "a"], "a.*"], ["@", "b"]]]]
*** Cannot use ["func", "match", ["@", "a"], "a.*"] with declared_type logical for required type value in comparable
$[?match(@.a, 'a.*') == @[1]]                                                  ["$", ["filt", ["==", ["func", "match", ["@", "a"], "a.*"], ["@", 1]]]]
$[?(match(@.a, 'a.*') && true == @.b) || (!match(@.a, 'a.*') && false == @.b)] ["$", ["filt", ["or", ["and", ["func", "match", ["@", "a"], "a.*"], ["==", true, ["@", "b"]]], ["and", ["not", ["func", "match", ["@", "a"], "a.*"]], ["==", false, ["@", "b"]]]]]]
*** Cannot use ["func", "length", ["@", "a"]] with declared_type value for required type logical in test expression
$[?length(@.a)]                                                                ["$", ["filt", ["func", "length", ["@", "a"]]]]
$[?length(@.a)==5]                                                             ["$", ["filt", ["==", ["func", "length", ["@", "a"]], 5]]]
$[?count(@.a)==5]                                                              ["$", ["filt", ["==", ["func", "count", ["@", "a"]], 5]]]
*** Cannot use 4 with declared_type value for required type nodes in ["func", "count", 4]
$[?count(4)==5]                                                                ["$", ["filt", ["==", ["func", "count", 4], 5]]]
$[?(@['key']==count(@))]                                                       ["$", ["filt", ["==", ["@", "key"], ["func", "count", ["@"]]]]]
```

gregsdennis:
@cabo although we haven't defined one, a nodelist-function is the outlier.  Assuming the follow definition for @glyn's ad-hoc proposal for `distinct()`, could you run the same tests, please?

Name: `distinct`

Arguments:
- `NodelistType` - Likely the result of a subquery, like `@.*`

Returns `NodelistType` - A copy of the argument nodelist with duplicates removed.  A duplicate (for the purpose of this comment) is the same JSON value regardless of its location in the data.

| Path to test | Validity | Node selection | Note |
| :- | :-: | :- | :- |
| `$[?distinct(@.*)]` | valid | selects the node if it has children | This example is not really useful as it should evaluate the same as `$[?@.*]` because removing duplicates has no effect on empty-ness |
| `$[?distinct(@.*)==42]` | valid | selects the node if the resulting nodelist is just a collection of one or more `42`s | Again, maybe not too useful |

I'm not looking for evaluation here, just parsing, as you have above.

draft-ietf-jsonpath-base.md
```diff
@@ -1102,11 +1102,10 @@ enclosing the identifier).
...
 result of a function expression (see {{fnex}}).
-In the latter case, if the function expression is of type
-`ST(OptionalBooleanType)` (see {{typesys}}), it tests whether the result
-is `true`; if the function expression is of type
-`ST(OptionalNodesType)`, it tests whether the result is different from
-`Nothing`.
+In the latter case, the function expression MUST be of type
+`BooleanType` (see {{typesys}}) or of type `NodelistType` to appear
+in a test expression.  The expression tests whether the result
+is or can be converted to `PropositionTrue` (see {{type-conv}}).
```
glyn, line 1108: `PropositionTrue` should be `BooleanTrue`.

draft-ietf-jsonpath-base.md
```diff
@@ -1411,82 +1410,101 @@ a type system is first introduced.
...
-| `Nothing`         | Singular Path or `filter-path` resulting in an empty nodelist      |
-| `Nodes(nl)`       | `filter-path` resulting in the non-empty nodelist `nl`             |
+| `Nothing`         | A representation of the absence of a JSON value, distinct from the JSON literal `null` |
+| `Nodes(nl)`       | `filter-path` resulting in the nodelist `nl`, which may or may not be empty  |
 {: #tbl-typerep title="Concrete representations of abstract instances"}
 
-The following subtype relationships depend on coercion:
+### Type Conversions {#type-conv}
+
+The following type conversions may occur:
```
glyn, line 1448: Type conversions seem to have to same downsides as type coercions in the current spec. We might be able to get away from conversions/coercions altogether by listing equivalent types where they are used.

draft-ietf-jsonpath-base.md
```diff
@@ -1411,82 +1410,101 @@ a type system is first introduced.
...
-is defined to have a result type in `ST(OptionalNodeOrValueType)`.
-* For it and any function expression it contains,
-each argument of the function matches the defined type of the argument
-according to one of the following rules:
-  * The argument is a function expression with defined result type
-    that is the same as, or a subtype of, the defined type of the argument.
-  * The argument is a literal primitive value and the defined type of the  argument is `Value` or any type of which `Value` is a subtype.
-  * The argument is a Singular Path and the defined type of the argument is `OptionalNodeType` or any type of which `OptionalNodeType` is a subtype.
-  * The argument is a `filter-path` (which includes Singular Paths) and the defined type of the argument is `OptionalNodesType`.
+A function expression is typed based on its return, and may only appear in expressions as follows:
```
glyn, line 1461: There are now no typing rules for functions calling other functions. Surely the result type of a function used as a function argument should line up with the argument type.

draft-ietf-jsonpath-base.md
```diff
@@ -1411,82 +1410,101 @@ a type system is first introduced.
...
+
+* If the function return is of type `ValueType`, it may only
+  appear as a `comparable` in a comparison.
+* If the function return is of type `BooleanType`, it may
+  only appear in a `test-expr`.
+* If the function return is of type `NodelistType`, it may
+  appear either as a `comparable` in a comparison or in
+  a `test-expr` as members of `NodelistType` can be converted
+  to either `ValueType` or `BooleanType`.
```
glyn, line 1470: I'd prefer `NodelistType` to be usable in comparisons and test expressions without conversion.

draft-ietf-jsonpath-base.md
```diff
@@ -1411,82 +1410,101 @@ a type system is first introduced.
...
+
+| Function return | Converted Value | Converted Boolean |
+| :-: | :-: | :-: |
+| empty nodelist | `Nothing` | `BooleanFalse` |
+| single-node nodelist | the node's JSON value | `BooleanTrue` |
+| multiple-node nodelist | `Nothing` | `BooleanTrue` |
+{: #tbl-typeconv title="Conversion of nodelists to values and booleans"}
+
+For example, given a `NodelistType` function `myfunc()`, the following paths
+are both valid but can have different and unexpected results:
```
glyn, line 1486: Why are the results unexpected for someone who has read the spec?

draft-ietf-jsonpath-base.md
```diff
@@ -1411,82 +1410,101 @@ a type system is first introduced.
...
+| function return | `myfunc(@.a)` evaluation (as a test) | `myfunc(@.a)==42` evaluation (in a comparison) |
+| :-: | :- | :- |
+| empty nodelist | The nodelist is converted to `BooleanFalse`.<br>The expression result is false. | The empty nodelist is converted to `Nothing`.<br>The expression result is false. |
+| single-node nodelist| The nodelist is converted to `BooleanTrue`.<br>The expression result is true. | The nodelist is converted to the node's value.<br>The expression result is true. |
+| mutliple-node nodelist| The nodelist is converted to `BooleanTrue`.<br>The expression result is true. | The nodelist is converted to `Nothing`.<br>The expression result is false. |
+
+The expression outcome is different between the test expression and comparison
+in the multiple-node nodelist case.
+This is a result of the type system and conversions, and it cannot be resolved
+in a way that keeps the type system consistent.
```
glyn, line 1497: Perhaps removing conversions would resolve the issue. However, I don't think the issue is type system inconsistency. The type system seems consistent to me. It's just that the behaviour might be a little surprising for people who are not familiar with the spec.

draft-ietf-jsonpath-base.md
```diff
@@ -1551,23 +1566,20 @@ string matches a given regular expression, which is in {{-iregexp}} form.
...
-Its first argument is an optional string that is matched against the iregexp
+Its first argument is a string that is matched against the iregexp
 contained in the string that is the second argument.
-The result is `true` if the string matches the iregexp and `false`
-otherwise.
-
-The result is `Nothing` if the first argument is not a string or
-the second argument is not a string conforming to {{-iregexp}}.
+The result is `BooleanTrue` if the string matches the iregexp
+and `BooleanFalse` otherwise.
```
glyn, line 1572: This doesn't say what happens if the first argument is not a string or if the second argument does not conform to i-regexp.

draft-ietf-jsonpath-base.md
```diff
@@ -1595,7 +1605,7 @@ the second argument is not a string conforming to {{-iregexp}}.
 | `$[?length(@.*) < 3]` | Invalid typing since `@.*` is a non-singular path |
 | `$[?count(@.*) == 1]` | Valid typing |
 | `$[?count(1) == 1]` | Invalid typing since `1` is not a path  |
-| `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `OptionalNodesType` and result type `OptionalNodesType` |
+| `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `NodelistType` and result type `NodelistType` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | Valid typing |
 | `$[?match(@.timezone, 'Europe/.*') == true]` | Valid typing |
```
glyn, line 1610: Shouldn't this now fail to type check? I thought `BooleanType` was not comparable to JSON true/false.

draft-ietf-jsonpath-base.md
```diff
@@ -1911,9 +1921,6 @@ IANA is requested to register the following media type {{RFC6838}}:
 Type name:
 : application
 
-Subtype name:
-: jsonpath
-
 Required parameters:
```
glyn, line 1924: Not sure why this change is needed for this PR.

draft-ietf-jsonpath-base.md
```diff
@@ -2019,7 +2026,7 @@ Column "Change Controller" always has the value "IESG" and the column
 
 | Function Name | Brief description                  | Input                          | Output            |
 | length        | length of array                    | `OptionalValueType`                | `OptionalValueType`   |
-| count         | size of nodelist                   | `OptionalNodesType`                | `Value`           |
+| count         | size of nodelist                   | `NodelistType`                | `Value`           |
 | match         | regular expression full match      | `OptionalNodeOrValueType`, `Value` | `OptionalBooleanType` |
 | search        | regular expression substring match | `OptionalNodeOrValueType`, `Value` | `OptionalBooleanType` |
 {: #pre-reg title="Initial Entries in the Function Extensions Subregistry"}
```
glyn, line 2032: This table still refers to some of the types which have gone away.

glyn (commented):
Thanks for this PR. It's going going in a reasonable direction although there are some rough edges which need sorting out. Please see my detailed comments.

I'd like to see a clearer terminological distinction between the propositional/logical type and JSON true/false.

Let's try and avoid type conversions/coercions as they seem to be more trouble than they are worth. (Also, the rest of the spec has studiously avoided implicit conversions, for similar reasons.)

@gregsdennis It may be worth deferring fixing up this PR if @cabo is about to provide something similar but with the rough edges removed.

cabo:
> could you run the same tests, please

```
$ cat test-data/distinct.jpl
$[?distinct(@.*)]
$[?distinct(@.*)==42]
$ jpt -l -fdistinct-nn test-data/distinct.jpl
$[?distinct(@.*)]     ["$", ["filt", ["func", "distinct", ["@", ["wild"]]]]]
$[?distinct(@.*)==42] ["$", ["filt", ["==", ["func", "distinct", ["@", ["wild"]]], 42]]]
$
```

So the well-typedness comes out well with these examples for a distinct(nodes) ➔ nodes (`distinct-nn`).
Distinct of course can generate general nodelists, not just singular ones, so this is a bit of the violation of the signature of `==`, but then we don't try to distinguish 0-or-1 nodelists from general nodelists in the simple type system we seem to be converging on.

(Get the JPT tool via `gem install jpt`.)

draft-ietf-jsonpath-base.md
gregsdennis: Thanks.  I had changed to "proposition" as in another PR, but then decided against it.  Must've missed this one coming back.

draft-ietf-jsonpath-base.md
gregsdennis: I listed these as "conversion" because I tend to see "coercion" as a value adjustment within the same time.  For instance, if a value should be constrained betwee 0 and 100, but -5 is submitted, the app could error or _coerce_ the value to 0.

But a conversion exists between values of different types.

I think the conversions are necessary.  The _types_ aren't equivalent, and neither are their values.  For example, you can't go from `BooleanFalse` to an empty nodelist (nor do we have any reason to).  It's a one-way conversion to make nodelists operate properly within expressions.

draft-ietf-jsonpath-base.md
gregsdennis: > Surely the result type of a function used as a function argument should line up with the argument type.

Yes.  I think this is implied by typing the function.  If function `foo()` is used as an argument for function `bar()`, i.e. `bar(foo())`, then `foo()` MUST return the same type that `bar()` expects to receive.

We can explicitly state this if you like, but I'm not sure it's necessary to.

draft-ietf-jsonpath-base.md
gregsdennis: Please explain how this can be?  Comparisons have values as operands and output booleans.  Test expressions have booleans has operands and output booleans.  Without conversion, nodelists would not be valid.

draft-ietf-jsonpath-base.md
gregsdennis: "otherwise" covers those cases.

draft-ietf-jsonpath-base.md
gregsdennis: As stated before, the conversions are necessary, and "removing" them is just describing them without naming them.  If you don't think list 1496 and 1497 are needed, I can remove them, but I thought it'd be fair to offer some explanation as to why the type system is expressed the way it is.

draft-ietf-jsonpath-base.md
gregsdennis: "Subtype" is no longer used, so the definition isn't needed.  We don't have subtypes.  We have conversions.

"Subtype" implies an inheritance model.  We have three disparate types.

draft-ietf-jsonpath-base.md
gregsdennis: Yes.  Will update.

gregsdennis:
> Distinct of course can generate general nodelists, not just singular ones, so this is a bit of the violation of the signature of == - @cabo 

First, thanks for runnig that test and sharing the results.

The type conversion from `NodelistType` to `ValueType` (which is accepted by `==`) now includes multiple nodelists as being converted to `Nothing`.  The need for this was explained in the final section of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/387#issuecomment-1436079071.

Since the ABNF only allows Singluar Paths and explicit JSON values to be used with `==`, functions become the only way to have a multiple nodelist in a comparison, so I think this conversion is suitable.

draft-ietf-jsonpath-base.md
cabo: This is a subtype of a media type in an IANA registration.
Fixed in #403.

glyn:
Yes, thanks @gregsdennis.

PR #403, opened 2023-02-21
Filter selector cleanup

### LINKED ISSUE #387: `Boolean` and `Value` (and their optional variants) should be distinct

### LINKED ISSUE #389: TestBoolean vs. JSONBoolean -- are they the same or different?

### LINKED ISSUE #401: Function return types
I notice that the match() function extension has return type `OptionalBooleanType (true, false, or Nothing)`

This seems wrong to me?  A regexp is provided and either there is a match or there isn't.  if the argument yields nothing then `match` yields false. How can `Nothing` be returned?  (Same is true for `search` obviously).

cabo:
General simplification and cleanup of the type system used in filter selectors.

Intended to supersede #402, which this is based on.
Intended to supersede #395.
Closes #387.
Closes #389.
Closes #401.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1093,7 +1093,7 @@ of each array element or object member value of the structured value being
 filtered; this element or member value is then known as the *current node*.
 
 The current node can be used as the start of one or more JSONPath
-queries used inside subexpressions of the filter expression, notated
+queries in subexpressions of the filter expression, notated
 via the current-node-identifier `@`.
 Each JSONPath query can be used either for testing existence of a
 result of the query, for obtaining a specific JSON value resulting
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1464,7 +1464,7 @@ a type system is first introduced.
 Each parameter and the result of a function extension must have a declared type.
 
 A type is a set of instances.
-Declared types enable testing a JSONPath query for well-typedness
+Declared types enable checking a JSONPath query for well-typedness
 independent of any argument the JSONPath query is applied to.
 
 {{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, and `nl` denotes
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -406,7 +406,7 @@ elements from an array, giving a start position, an end position, and
 an optional step value that moves the position from the start to the
 end.
 
-Filter expressions `?<expr>` select certain children of an object or array, as in:
+Filter expressions `?<logical-expr>` select certain children of an object or array, as in:
 
 ~~~~ JSONPath
 $.store.book[?@.price < 10].title
@@ -430,7 +430,7 @@ $.store.book[?@.price < 10].title
 | `*`                 | [wildcard selector](#name-selector): selects all children of a node                                                    |
 | `3`                 | [index selector](#index-selector): selects an indexed child of an array (from 0)                                        |
 | `0:100:5`           | [array slice selector](#slice): start:end:step for arrays                                                               |
-| `?<expr>`           | [filter selector](#filter-selector): selects particular children using a logical expression                      |
+| `?<logical-expr>`   | [filter selector](#filter-selector): selects particular children using a logical expression                      |
 | `length(@.foo)`     | [function extension](#fnex): invokes a function in a filter expression                                                  |
 {: #tbl-overview title="Overview of JSONPath syntax"}
 
@@ -1117,7 +1117,7 @@ function expressions are not in use.
 #### Syntax
 {: unnumbered}
 
-The filter selector has the form `?<expr>`.
+The filter selector has the form `?<logical-expr>`.
 
 ~~~~ abnf
 filter-selector     = "?" S logical-expr
@@ -2246,7 +2246,7 @@ The descendant operators, starting with `..`, borrowed from {{E4X}}, are similar
 The array slicing construct `[start:end:step]` is unique to JSONPath,
 inspired by {{SLICE}} from ECMASCRIPT 4.
 
-Filter expressions are supported via the syntax `?<expr>` as in
+Filter expressions are supported via the syntax `?<logical-expr>` as in
 
 ~~~~ JSONPath
 $.store.book[?@.price < 10].title
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1664,7 +1664,8 @@ If the first argument is not a string or the second argument is not a
 string conforming to {{-iregexp}}, the result is `LogicalFalse`. <!-- XXX -->
 Otherwise, the string that is the first argument is searched for at
 least one substring that matches the iregexp contained in the string
-that is the second argument; the result is `LogicalTrue` if
+that is the second argument; the result is `LogicalTrue` if such a
+substring exists and `LogicalFalse` otherwise.
 
 
 ### Examples
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1480,7 +1480,7 @@ a nodelist.
 Notes:
 
 * `ValueType` is an abstraction of a JSON value or `Nothing`.
-* `LogicalType` is an abstraction of a the result of a `test-expr`.
+* `LogicalType` is an abstraction of the value of a `logical-expr`.
   Its `LogicalTrue` and `LogicalFalse` values are not relatable to
   the JSON literals `true` and `false` and have no syntactical representation in JSON Path.
 * `NodesType` is an abstraction of a `filter-path` (which appears
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1085,7 +1085,7 @@ For each iteration (element/member), a logical expression, the *filter expressio
 is evaluated which decides whether the node of
 the element/member is selected.
 (While a logical expression evaluates to what mathematically is a
-Boolean value, we use the term *logical* to maintain a distinction from
+Boolean value, this specification uses the term *logical* to maintain a distinction from
 the Boolean values that JSON can represent.)
 
 During the iteration process, the filter expression receives the node
@@ -1535,7 +1535,7 @@ A function expression is well-typed if all of the following are true:
    * The argument is a Singular Path or `filter-path` (which includes
      Singular Paths), or a function expression with declared result
      type `NodesType`.  Where the declared type of the parameter is
-     not `NodeType`, a conversion applies.
+     not `NodesType`, a conversion applies.
 
 #### Conversion example
 {:unnumbered}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -517,7 +517,7 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
-2. Uses of function extensions must be correctly typed,
+2. Uses of function extensions must be *well-typed*,
 as described in {{fnex}}.
 
 A JSONPath implementation MUST raise an error for any query which is not
@@ -1112,6 +1112,8 @@ function is declared.
 The type system is limited in scope; its function is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
+The type system also guides conversions ({{type-conv}}) that are
+implicit in the grammar without the use of function expressions.
 
 #### Syntax
 {: unnumbered}
@@ -1457,10 +1459,10 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
 or a `comparable`.
 
 Any function expressions in a query must be well-formed (by conforming to the above ABNF)
-and correctly typed,
+and well-typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
-To define which function expressions are correctly typed,
+To define which function expressions are well-typed,
 a type system is first introduced.
 
 ### Type System for Function Expressions {#typesys}
@@ -1516,9 +1518,9 @@ The following type conversions may occur:
     is `LogicalTrue`.
   * If the nodelist is empty, the conversion result is `LogicalFalse`.
 
-The type correctness of function expressions can now be defined in terms of this type system.
+The well-typedness of function expressions can now be defined in terms of this type system.
 
-### Type Correctness of Function Expressions
+### Well-Typedness of Function Expressions
 
 A function expression is well-typed if all of the following are true:
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -2082,7 +2082,7 @@ Brief description:
 : a brief description
 
 Parameters:
-: A comma-separated list of zero or more declared types, each for one of the
+: A comma-separated list of zero or more declared types, one for each of the
   arguments expected for this function extension
 
 Result:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1468,7 +1468,7 @@ Declared types enable checking a JSONPath query for well-typedness
 independent of any argument the JSONPath query is applied to.
 
 {{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, and `nl` denotes
-a non-empty nodelist.
+a nodelist.
 
 | Type                 | Abstract Instances                       |
 | :--                  | :----------------                        |
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1107,7 +1107,7 @@ functions predefined, see {{fnex}}, and the ability to register further
 functions provided by the Function Extensions sub-registry ({{iana-fnex}}).
 When a function is defined, it is given a unique name, and its return value and each of its arguments is given a
 *declared type*.
-The type system is limited in scope; its function is to express
+The type system is limited in scope; its purpose is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
 The type system also guides conversions ({{type-conv}}) that are
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1679,7 +1679,7 @@ substring exists and `LogicalFalse` otherwise.
 | `$[?count(1) == 1]` | not well-typed since `1` is not a path  |
 | `$[?count(foo(@.*)) == 1]` | well-typed, where `foo` is a function extension with a parameter of type `NodesType` and result type `NodesType` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | well-typed |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals and JSON booleans do not mix |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals may not be used in comparisons |
 {: title="Function expression examples"}
 
 ## Segments  {#segments-details}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -406,7 +406,7 @@ elements from an array, giving a start position, an end position, and
 an optional step value that moves the position from the start to the
 end.
 
-Filter expressions `?<boolean expr>` select certain children of an object or array, as in:
+Filter expressions `?<expr>` select certain children of an object or array, as in:
 
 ~~~~ JSONPath
 $.store.book[?@.price < 10].title
@@ -430,7 +430,7 @@ $.store.book[?@.price < 10].title
 | `*`                 | [wildcard selector](#name-selector): selects all children of a node                                                    |
 | `3`                 | [index selector](#index-selector): selects an indexed child of an array (from 0)                                        |
 | `0:100:5`           | [array slice selector](#slice): start:end:step for arrays                                                               |
-| `?<expr>`           | [filter selector](#filter-selector): selects particular children using a boolean expression                             |
+| `?<expr>`           | [filter selector](#filter-selector): selects particular children using a logical expression                      |
 | `length(@.foo)`     | [function extension](#fnex): invokes a function in a filter expression                                                  |
 {: #tbl-overview title="Overview of JSONPath syntax"}
 
@@ -1139,7 +1139,7 @@ any other than the directly enclosing filter-selector (i.e., of
 filter-selectors enclosing the filter-selector that is directly
 enclosing the identifier).
 
-Logical expressions offer the usual boolean operators (`||` for OR,
+Logical expressions offer the usual Boolean operators (`||` for OR,
 `&&` for AND, and `!` for NOT).
 Parentheses MAY be used within `logical-expr` for grouping.
 
@@ -1165,17 +1165,12 @@ A test expression is shorthand for a comparison that
 either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
 result of a function expression (see {{fnex}}).
-In the latter case, if the function expression is of type
-`ST(OptionalBooleanType)` (see {{typesys}}), it tests whether the result
-is `true`; if the function expression is of type
-`ST(OptionalNodesType)`, it tests whether the result is different from
-`Nothing`.
-
-| Test Expression                              | result   | comparison | with    |
-|----------------------------------------------|----------|------------|---------|
-| embedded query (`filter-path`)               | nodelist | !=         | Nothing |
-| `function-expr` <br> ST(OptionalNodesType)   | nodelist | !=         | Nothing |
-| `function-expr` <br> ST(OptionalBooleanType) | value    | ==         | true    |
+In the latter case, if the function result type is declared as
+`LogicalType` (see {{typesys}}), it tests whether the result
+is `LogicalTrue`; if the function result type is declared as
+`NodesType`, it tests whether the result is non-empty.
+If the declared function result type is `ValueType`, its use in a
+test expression is not well-typed.
 
 ~~~ abnf
 
@@ -1259,7 +1254,7 @@ Children of an array appear in array order in the resultant nodelist.
 ##### Existence Tests {#extest}
 {: unnumbered}
 
-A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
+A path by itself in a Logical context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
 
 Existence tests differ from comparisons in that:
 
@@ -1316,7 +1311,7 @@ compared is an object, array, boolean, or `null`.
 * The comparison `a > b` yields true if and only if `b < a` yields true.
 * The comparison `a >= b` yields true if and only if `b < a` yields true or `a == b` yields true.
 
-##### Boolean Operators
+##### Logical Operators
 {: unnumbered}
 
 The logical AND, OR, and NOT operators have the normal semantics of Boolean algebra and
@@ -1547,14 +1542,14 @@ A function expression is well-typed if all of the following are true:
 
 While functions returning `NodesType` can appear in both comparisons and test expressions,
 path authors must be aware of the conversions and the potential outcomes.
-Specifically, nodelists convert to values and booleans according to {{tbl-typeconv}}:
+Specifically, nodelists convert to values and logicals according to {{tbl-typeconv}}:
 
 | Function return        | Converted as Value    | Converted as Logical |
 | :-:                    | :-:                   | :-:                  |
 | empty nodelist         | `Nothing`             | `LogicalFalse`       |
 | single-node nodelist   | the node's JSON value | `LogicalTrue`        |
 | multiple-node nodelist | `Nothing`             | `LogicalTrue`        |
-{: #tbl-typeconv title="Conversion of nodelists to values and booleans"}
+{: #tbl-typeconv title="Conversion of nodelists to values and logicals"}
 
 For example, given a function `myfunc()` of declared result type
 `NodesType`, the following filter expressions are both valid but can
@@ -1678,13 +1673,13 @@ that is the second argument; the result is `LogicalTrue` if
 
 | Query | Comment |
 | :---: | ------- |
-| `$[?length(@) < 3]` | Valid typing |
-| `$[?length(@.*) < 3]` | Invalid typing since `@.*` is a non-singular path |
-| `$[?count(@.*) == 1]` | Valid typing |
-| `$[?count(1) == 1]` | Invalid typing since `1` is not a path  |
-| `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `NodesType` and result type `NodesType` |
-| `$[?match(@.timezone, 'Europe/.*')]`         | Valid typing |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | Valid typing |
+| `$[?length(@) < 3]` | well-typed |
+| `$[?length(@.*) < 3]` | not well-typed since `@.*` is a non-singular path |
+| `$[?count(@.*) == 1]` | well-typed |
+| `$[?count(1) == 1]` | not well-typed since `1` is not a path  |
+| `$[?count(foo(@.*)) == 1]` | well-typed, where `foo` is a function extension with argument of type `NodesType` and result type `NodesType` |
+| `$[?match(@.timezone, 'Europe/.*')]`         | well-typed |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals and JSON booleans do not mix |
 {: title="Function expression examples"}
 
 ## Segments  {#segments-details}
@@ -2086,12 +2081,12 @@ Function Name:
 Brief description:
 : a brief description
 
-Input:
-: A comma-separated list of zero or more types of the
+Parameters:
+: A comma-separated list of zero or more declared types, each for one of the
   arguments expected for this function extension
 
-Output:
-: The type of the result for this function extension
+Result:
+: The declared type of the result for this function extension
 
 Change Controller:
 : (see {{Section 2.3 of -ianacons}})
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
-| `OptionalValueType`       | `Value(v)`, `Nothing`                    | `OptionalNodeType`, `ValueType`, `OptionalBooleanType` |
-| `ValueType`               | `Value(v)`                               | `BooleanType`                                  |
-| `OptionalBooleanType`     | `Value(true)`, `Value(false)`, `Nothing` | `BooleanType`                                  |
-| `BooleanType`             | `Value(true)`, `Value(false)`            |                                                |
-| `OptionalNodesType`       | `Nodes(nl)`, `Nothing`                   | `OptionalNodeType`                             |
+a non-empty nodelist.
+
+| Type                 | Abstract Instances                       |
+| :--                  | :----------------                        |
+| `ValueType`          | `Value(v)`, `Nothing`                    |
```
glyn, line 1476: I prefer `OptionalValueType` as this gives a hint of the possibility of the value `Nothing`.

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
+| `LogicalType`        | `LogicalTrue`, `LogicalFalse`            |
+| `NodesType`       | `Nodes(nl)`                              |
 {: #tbl-types title="Function extension type system"}
 
 Notes:
 
-* `OptionalNodeOrValueType` is an abstraction of a `comparable` (which may appear on either side of a comparison or as a function argument).
-* `OptionalNodeType` is an abstraction of a Singular Path.
```
glyn, line 1437: This may still be needed as an analog of Singular Path in comparisons.

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
-* `ValueType` is an abstraction of a primitive value.
```
glyn, line 1438: We may actually still need this type. For example, `count()` never returns `Nothing`, so the return type doesn't accurately describe the possible return values.

draft-ietf-jsonpath-base.md
```diff
@@ -1496,9 +1582,9 @@ filter expression:
 $[?length(@.authors) >= 5]
 ~~~
 
-Its only argument is an optional value (possibly taken from a singular path as
-in the example above).  The result also is an optional value, an unsigned
-integer.
+Its only argument is an instance of `ValueType` (possibly taken from a
+singular path as in the example above).  The result also is an
```
glyn, line 1586: We also have to allow the argument to be a function expression, in which case it should be an analog of a Singular Path like `OptionalNodeOrValueType`.

draft-ietf-jsonpath-base.md
```diff
@@ -1512,7 +1598,7 @@ integer.
 ### `count` Function Extension {#count}
 
 Arguments:
-: 1. `OptionalNodesType`
+: 1. `NodesType`
 
 Result:
 : `ValueType` (unsigned integer)
```
glyn, line 1604: We should probably use a type which doesn't include `Nothing`. Failing that, the spec for this function should make it clear that `Nothing` is never returned.

draft-ietf-jsonpath-base.md
```diff
@@ -1551,23 +1634,23 @@ string matches a given regular expression, which is in {{-iregexp}} form.
...
 ~~~
 
-Its first argument is an optional string that is matched against the iregexp
-contained in the string that is the second argument.
-The result is `true` if the string matches the iregexp and `false`
-otherwise.
-
-The result is `Nothing` if the first argument is not a string or
-the second argument is not a string conforming to {{-iregexp}}.
+Its arguments are instances of ValueType.
```
glyn, line 1637: We should mention that a Singular Path is also acceptable as the first argument.

draft-ietf-jsonpath-base.md
```diff
@@ -1577,27 +1660,26 @@ expression, which is in {{-iregexp}} form.
...
 ~~~
 
-Its first argument is an optional string that is searched for at least one
-substring that matches the iregexp contained in the string
-that is the second argument.
-The result is `true` if such a substring exists, `false` otherwise.
-
-The result is `Nothing` if the first argument is not a string or
-the second argument is not a string conforming to {{-iregexp}}.
+Its arguments are instances of ValueType.
```
glyn, line 1663: We should mention that a Singular Path is also acceptable as the first argument.

draft-ietf-jsonpath-base.md
```diff
@@ -1577,27 +1660,26 @@ expression, which is in {{-iregexp}} form.
...
+If the first argument is not a string or the second argument is not a
+string conforming to {{-iregexp}}, the result is `LogicalFalse`. <!-- XXX -->
+Otherwise, the string that is the first argument is searched for at
+least one substring that matches the iregexp contained in the string
+that is the second argument; the result is `LogicalTrue` if
```
glyn, line 1668: The above sentence was truncated.

draft-ietf-jsonpath-base.md
```diff
@@ -1577,27 +1660,26 @@ expression, which is in {{-iregexp}} form.
...
-| `$[?count(foo(@.*)) == 1]` | Valid typing, where `foo` is a function extension with argument of type `OptionalNodesType` and result type `OptionalNodesType` |
-| `$[?match(@.timezone, 'Europe/.*')]`         | Valid typing |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | Valid typing |
+| `$[?length(@) < 3]` | well-typed |
+| `$[?length(@.*) < 3]` | not well-typed since `@.*` is a non-singular path |
+| `$[?count(@.*) == 1]` | well-typed |
+| `$[?count(1) == 1]` | not well-typed since `1` is not a path  |
+| `$[?count(foo(@.*)) == 1]` | well-typed, where `foo` is a function extension with argument of type `NodesType` and result type `NodesType` |
+| `$[?match(@.timezone, 'Europe/.*')]`         | well-typed |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals and JSON booleans do not mix |
```
glyn, line 1682:
```suggestion
| `$[?match(@.timezone, 'Europe/.*') == true]` | invalid as JSONPath logicals are not well-typed in comparisons |
```

draft-ietf-jsonpath-base.md
```diff
@@ -1999,12 +2081,12 @@ Function Name:
 Brief description:
 : a brief description
 
-Input:
-: A comma-separated list of zero or more types of the
+Parameters:
+: A comma-separated list of zero or more declared types, each for one of the
```
glyn, line 2085:
```suggestion
: A comma-separated list of zero or more declared types, one for each of the
```

draft-ietf-jsonpath-base.md
```diff
@@ -402,7 +406,7 @@ elements from an array, giving a start position, an end position, and
 an optional step value that moves the position from the start to the
 end.
 
-Filter expressions `?<boolean expr>` select certain children of an object or array, as in:
+Filter expressions `?<expr>` select certain children of an object or array, as in:
```
glyn, line 409: Why not use `logical-expr` instead and elsewhere?
```suggestion
Filter expressions `?<logical-expr>` select certain children of an object or array, as in:
```

glyn (changes requested):
I like the general approach, although, after further consideration, I think we need to have more than three types. I've made various comments and suggestions, at various levels. Some of the editorial changes can easily be adopted in this PR. The removal of type conversions is a larger change and could be deferred to a follow-on PR.

draft-ietf-jsonpath-base.md
```diff
@@ -1120,36 +1155,49 @@ logical-and-expr    = basic-expr *(S "&&" S basic-expr)
...
-and `null`),
-Singular Paths, each of which selects at most one node, and
-function expressions (see {{fnex}}) that
-return a primitive value or at most one node.
+Comparison expressions are available for comparisons between primitive
+values (that is, numbers, strings, `true`, `false`, and `null`).
+These can be obtained via literal values; Singular Paths, each of
+which selects at most one node the value of which is then used; and
+function expressions (see {{fnex}}) of type `ValueType` or
+`NodesType` (see {{type-conv}}).
```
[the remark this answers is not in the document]
gregsdennis: I disagree.  The grammar (I assume you mean the ABNF exclusively) is augmented by the type system.  The type system provides additional "well-formedness" checks that haven't been (or can't be) expressed in the ABNF.

I see them as each defining a set of restrictions, and the set of valid expressions lies in the union of those restrictions.

draft-ietf-jsonpath-base.md
```diff
@@ -1202,7 +1254,7 @@ Children of an array appear in array order in the resultant nodelist.
 ##### Existence Tests {#extest}
 {: unnumbered}
 
-A path by itself in a Boolean context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
+A path by itself in a Logical context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
```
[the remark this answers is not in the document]
gregsdennis: As I described in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/402#discussion_r1113518922, this requires that you describe the conversion without calling it a "conversion."

draft-ietf-jsonpath-base.md
gregsdennis: I tend to not get hung up on naming, and while I don't fully disagree with you, I think `ValueType` is considerably less clunky than `OptionalValueType`.  `ValueType` being defined here in this way is sufficient for me.

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
+  expression, and the following rules for function arguments apply to
+  its declared result type.
+* Each argument of the function can be used for the declared type of the corresponding declared
+  parameter according to one of the following rules:
+   * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
+   * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
+   * The argument is a Singular Path or `filter-path` (which includes
+     Singular Paths), or a function expression with declared result
+     type `NodesType`.  Where the declared type of the parameter is
+     not `NodeType`, a conversion applies.
```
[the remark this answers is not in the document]
gregsdennis: I agree with the correction, but I don't see any use of a `NodeType`.  The only places we deal with a single node are in defining `@` and converting a single nodelist into a value.  Neither of these cases warrant a type being defined for them.

draft-ietf-jsonpath-base.md
gregsdennis: The only kind of function that is valid here would be one that returns `ValueType`.  "An instance of `ValueType`" includes this.

The result of a Singular Path is also convertible to `ValueType`, so no issues there.

draft-ietf-jsonpath-base.md
gregsdennis: Again, I think "instances of `ValueType` includes these.

gregsdennis, line 1637:
```suggestion
Its arguments are instances of `ValueType`.
```

draft-ietf-jsonpath-base.md
gregsdennis, line 1663:
```suggestion
Its arguments are instances of `ValueType`.
```

draft-ietf-jsonpath-base.md
```diff
@@ -2164,7 +2246,7 @@ The descendant operators, starting with `..`, borrowed from {{E4X}}, are similar
 The array slicing construct `[start:end:step]` is unique to JSONPath,
 inspired by {{SLICE}} from ECMASCRIPT 4.
 
-Filter expressions are supported via the syntax `?<boolean expr>` as in
+Filter expressions are supported via the syntax `?<expr>` as in
```
gregsdennis, line 2249: This would need the same treatment as https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403/files#r1113477827 if we decide to do that.

gregsdennis (changes requested):
I like this.  The essentials are the same as #402, but with more description and some renamings.

I largely agree with @glyn's wordsmithing suggestions, but I disagree with his opinion that we don't need conversions ([here's why](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/402#discussion_r1113518922)).

draft-ietf-jsonpath-base.md
```diff
@@ -1072,22 +1076,58 @@ The following examples show the array slice selector in use by a child segment.
...
+Filter selectors are used to iterate over the elements or members of
+structured values, i.e., JSON arrays and objects.
+The structured values are identified in the nodelist offered by the
+child or descendant segment the filter selector is used in.
+
+For each iteration (element/member), a logical expression is
+evaluated, the *filter expression*, which decides whether the node of
+the element/member is selected.
+(While a logical expression evaluates to what mathematically is a
+Boolean value, we use the term *logical* to maintain a distinction from
```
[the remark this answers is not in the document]
cabo: Passive is really bad here.
Fixed in b18e22e

draft-ietf-jsonpath-base.md
```diff
@@ -1072,22 +1076,58 @@ The following examples show the array slice selector in use by a child segment.
...
+The set of available functions is extensible, with a number of
+functions predefined, see {{fnex}}, and the ability to register further
+functions in the Function Extensions sub-registry ({{iana-fnex}}).
+When a function is defined, it is given a unique name, and the
+*declared type*
+of each of its arguments as well as the declared type returned from the
+function is declared.
+The type system is limited in scope; its function is to express
+restrictions that, without functions, are implicit in the grammar of
+filter expressions.
```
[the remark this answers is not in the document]
gregsdennis: Maybe this case can be added to the [conversion section](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403/files#diff-aca3bb236d11f768483f8a1153cb9030b67d9cd6a39f012e8e4bfec2c6275165R1538) as another "gotcha" example.

While I agree that it's surprising, the spec doesn't claim to have a way to compare nodelists, and I think, once thought through, a result of `true` for your example can be expected.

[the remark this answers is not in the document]
cabo: I'm actually leaning in the inverse direction: getting rid of "singular types" in the grammar.
But it would be good to have a proposal for going in the direction you suggest.
Not in this PR, though...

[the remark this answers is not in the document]
gregsdennis: To be clear, what do you mean by "singular types?"

draft-ietf-jsonpath-base.md
```diff
@@ -1072,22 +1076,58 @@ The following examples show the array slice selector in use by a child segment.
...
+The type system also guides conversions ({{type-conv}}) that are
+implicit in the grammar without the use of function expressions.
```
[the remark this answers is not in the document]
gregsdennis: Sure it does.  `@.a == 42` contains an implicit conversion from the singular nodelist returned by `@.a` to the value that the single node contains.  `NodesType` -> `ValueType`.

Without this implicit conversion, you have to define what `==` means for

- nodelist == nodelist
- nodelist == value
- value == value

But with the conversion, you only need to define "value == value."

draft-ietf-jsonpath-base.md
```diff
@@ -1072,22 +1076,58 @@ The following examples show the array slice selector in use by a child segment.
...
-filter-selector     = "?" S boolean-expr
+filter-selector     = "?" S logical-expr
 ~~~~
 
-During the iteration process the node of each array element or object member value being visited is known as the current node.
-A boolean expression, usually involving the current node, is evaluated and
-the current node is selected if and only if the expression yields true.
-
-As the expression is composed of side-effect free components,
+As the filter expression is composed of side-effect free components,
```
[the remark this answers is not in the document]
cabo: More global attempt at reducing the use of "component" in f7ea77d

draft-ietf-jsonpath-base.md
```diff
@@ -1120,36 +1155,49 @@ logical-and-expr    = basic-expr *(S "&&" S basic-expr)
 basic-expr          = paren-expr /
                       comparison-expr /
                       test-expr
+
+paren-expr          = [logical-not-op S] "(" S logical-expr S ")"
+                                        ; parenthesized expression
+logical-not-op      = "!"               ; logical NOT operator
+~~~~
+
+A test expression is shorthand for a comparison that
```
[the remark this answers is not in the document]
cabo: Fixed in f111cff

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: This is an interesting discussion, but I think no action is needed here in this PR.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: No action needed, I think?

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
-* `OptionalNodeOrValueType` is an abstraction of a `comparable` (which may appear on either side of a comparison or as a function argument).
-* `OptionalNodeType` is an abstraction of a Singular Path.
-* `ValueType` is an abstraction of a primitive value.
-* `BooleanType` is an abstraction of a primitive value that is either
-  `true` or `false`.
-* `OptionalValueType` is an abstraction of a primitive value that may
-  alternatively be absent (`Nothing`).
-* `OptionalNodesType` is an abstraction of a `filter-path` (which appears
+* `ValueType` is an abstraction of a JSON value or `Nothing`.
+* `LogicalType` is an abstraction of a the result of a `test-expr`.
```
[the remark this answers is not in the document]
gregsdennis: This suggestion works if you define the "value of a `logical-expr`" to be its result, but I don't think we've done that.  I'd also prefer to leave "value" out of the logical context, to be honest.

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
+  Its `LogicalTrue` and `LogicalFalse` values are not relatable to
```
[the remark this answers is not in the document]
gregsdennis: No, relatable is correct, but they cannot "be related" (verb).

draft-ietf-jsonpath-base.md
gregsdennis: "(unsigned integer)" as compared with `length()`'s "(unsigned integer or `Nothing`)" makes that clear, I think.

draft-ietf-jsonpath-base.md
gregsdennis:
```suggestion
that is the second argument; the result is `LogicalTrue` if
 the string contains a match the iregexp and `LogicalFalse` otherwise.
```

(referenced from `match()` definition)

cabo: Fixed in 3e1265a

draft-ietf-jsonpath-base.md
cabo: I tried to be a bit informal in this example...
(And I certainly don't want to regress from "well-typed" to "invalid" (which has been rooted out, not the least as it can be considered a non-word these days.)

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1454,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
+### Well-Typedness of Function Expressions
+
+A function expression is well-typed if all of the following are true:
+
+* If it occurs directly in a test expression, the function is declared
+  to have a result type of `LogicalType`, or (conversion applies)
+  `NodesType`.
+* If it occurs directly as a `comparable` in a comparison, the
+  function is declared to have a result type of `ValueType`, or
+  (conversion applies) `NodesType`.
```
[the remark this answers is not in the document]
gregsdennis: If you want to do this, you'd have to define a new type for single-nodelists (separate from `NodesType`) so that functions can specify that this is what they return and can be used in comparisons.  While you could then restrict `NodesType` functions to only appearing in test expressions, I'm not convinced this new type is any more useful than the function just returning the value that the node contains (`ValueType`).

draft-ietf-jsonpath-base.md
gregsdennis:
```suggestion
| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals may not be used in comparisons |
```

draft-ietf-jsonpath-base.md
cabo: I liked the reduced weight, but went with the suggestion (throughout) in 58e7ed0

draft-ietf-jsonpath-base.md
cabo: I don't understand what you mean by "do that" (and the line has already changed into ?<logical-expr>)...

gregsdennis: "do that" == "follow Glyn's suggestion," which you did, and you changed it here, too.  This is resolved.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: `singular-path` as a feeble duplication of `filter-path`.
(We also need to fix the suffix "path" in these things.  Later...)

cabo:
> I like the general approach, although, after further consideration, I think we need to have more than three types. I've made various comments and suggestions, at various levels. Some of the editorial changes can easily be adopted in this PR. The removal of type conversions is a larger change and could be deferred to a follow-on PR.

So I think what we want to do is get this PR consistent, and then open the floodgates on 3 vs. 4 types.
At least I now know that you can implement this spec in a day :-)

gregsdennis:
> At least I now know that you can implement this spec in a day

I think that likely depends on the language.

cabo:
On 2023-02-21, at 23:13, Greg Dennis ***@***.***> wrote:
> 
>  I'd also prefer to leave "value" out of the logical context, to be honest.

Indeed.
708def9

gregsdennis (approved):

glyn (changes requested):
I've reviewed the unresolved comment threads and they represent what I think remains to be discussed. Thanks @cabo for sorting out the rest (with @gregsdennis's input too). Here's a summary:

1. Whether or not functions (and/or paths) returning non-singular nodelists should be allowed in comparisons. I don't think they should be allowed, so I regard the PR as a regression in this respect from the current spec and I don't want to merge it.

2. Whether the spec would be clearer if it avoided type conversions. This is an editorial concern and it shouldn't block merging this PR.

3. Whether we need a subtype of the PR's `ValueType` without `Nothing`.  This is a relatively minor point I would be happy to cover in an issue. It shouldn't block merging this PR.

I'd like to request that item 1 be addressed (to avoid "non-singular comparisons") in this PR, which can then be merged. There can then be a follow-on issue to explore the general topic of non-singular comparisons. My recollection is that there has already been considerable discussion in the WG on this topic, so I would definitely like further discussion to occur in its own issue for better visibility.

gregsdennis:
> Whether or not functions (and/or paths) returning non-singular nodelists should be allowed in comparisons.

From the ABNF, paths that return multiple nodes are not allowed in comparisons.  I agree with this, and this PR is not changing it.

From the type system, as described here, functions that return multiple nodelists are allowed in comparisons.  I think this is fine _with the defined conversions_.  To change this, we would need to (one of):

- disallow nodelist functions in comparisons at all
- define a new type that represents a single-node nodelist that _is_ suitable for comparisons
- (there's also a third option)

## Disallow nodelist function in comparisons

I'm okay with this.  I'm not sure if this would limit any functionality, though.  It would limit functions that return `NodesType` to logicals because the nodelist is still convertible to (or "has meaning in," or however you want to say it) a logical.

The [examples](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/402#issuecomment-1437796107) I put up for the hypothetical `distinct()` function were contrived to endure @cabo's tests.  I certainly can't think of a practical use case to put a nodelist in a comparison.

This does raise the question of whether to even have `NodesType`, though.  If a function that returns `NodesType` is only valid as a logical, then it might as well return `BooleanType`.

## Define a new type for single-node nodelists

A function that returns a single-node nodelist would be valid in a comparison, but a function that returns `NodesType` would not.

This is really an extension of disallowing nodelist functions in comparisons, so all of that logic still applies (i.e. is `NodesType` still useful?).

Conversely, a function that returns the new single-node nodelist type would only be useful in a comparison (otherwise, we run into the same inconsistencies that I've been highlighting this whole time).  Because it's only useful in a comparison, it might as well just return the value of the single node, which is `ValueType`.  So why have it?

## The third option

Only allow nodelist functions to be parameters for other functions, i.e. disallow them from both logicals and comparisons.

This would only allow something like `$[?count(distinct(@.*))==42]`, which is arguably useful, but both `$[?distinct(@.*)]` and `$[?distinct(@.*)==42]` would be invalid.

---

> Whether the spec would be clearer if it avoided type conversions.

We cannot avoid type conversions.  They are implicit in the grammar.

To prove this, I'll exclude functions.

**All paths return nodelists.  They don't return logicals, and they don't return values.**  Singular paths, specifically, return a nodelist with a single value.

Paths are always allowed to act as logicals.  That is, there is a _conversion_ from a nodelist to a logical.  Namely, if a nodelist has one or more nodes, it is treated as (converted to) `LogicalTrue`; if it has no nodes, it is treated as (converted to) `LogicalFalse`.  If I were to write `$[?@.a]` as a test expression explicitly (in a C-ish syntax), it would be:

```
$[?(LogicalType)@.a]
```

where I'm "casting" the nodelist returned by `@.a` to `LogicalType`.

Paths are allowed to act as values if they're singular.  That is, there is a _conversion_ from a nodelist to a value if the nodelist contains at most single node.  If the nodelist were to contain more than one node, the conversion would be undefined, and a runtime error would occur.  To avoid the potential error, the ABNF applies an additional restriction, and we can catch it at parse time.  Writing `$[?@.a==42]` explicitly would be:

```
$[?((ValueType)@.a)==42]  // extra parentheses not necessary, but can help readability
```

where I'm "casting" the nodelist returned by `@.a` to a `ValueType`.

We have specified these conversions by describing their behavior, but that doesn't detract from the fact that they're there.

Type conversions are unavoidable.  We might as well state them.  I think it helps us in the end.

glyn:
> > Whether or not functions (and/or paths) returning non-singular nodelists should be allowed in comparisons.
> 
> From the ABNF, paths that return multiple nodes are not allowed in comparisons. I agree with this, and this PR is not changing it.
> 
> From the type system, as described here, functions that return multiple nodelists are allowed in comparisons. I think this is fine _with the defined conversions_ (also see below).

I don't think it's fine as it introduces (more) comparisons with surprising results, such as:
```
<function returning two nodes> == <function returning three nodes>

<path producing an empty nodelist> == <function returning two nodes>
```
which both yield true.

> To change this, we would need to (one of):
> 
>     * disallow nodelist functions in comparisons at all
> 
>     * define a new type that represents a single-node nodelist that _is_ suitable for comparisons
> 
>     * (there's also a third option)
> 
> 
> ## Disallow nodelist function in comparisons
> 
> I'm okay with this. I'm not sure if this would limit any functionality, though. It would limit functions that return `NodesType` to logicals because the nodelist is still convertible to (or "has meaning in," or however you want to say it) a logical.
> 
> The [examples](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/402#issuecomment-1437796107) I put up for the hypothetical `distinct()` function were contrived to endure @cabo's tests. I certainly can't think of a practical use case to put a nodelist in a comparison.
> 
> This does raise the question of whether to even have `NodesType`, though. If a function that returns `NodesType` is only valid as a logical, then it might as well return `BooleanType`.

Not so. Such functions can be passed as arguments to other functions, e.g. `count(distinct($.a.*)) > 1`.

> ## Define a new type for single-node nodelists
> 
> A function that returns a single-node nodelist would be valid in a comparison, but a function that returns `NodesType` would not.

This is currently my preferred option.

> 
> This is really an extension of disallowing nodelist functions in comparisons, so all of that logic still applies (i.e. is `NodesType` still useful?).
> 
> Conversely, a function that returns the new single-node nodelist type would only be useful in a comparison (otherwise, we run into the same inconsistencies that I've been highlighting this whole time). Because it's only useful in a comparison, it might as well just return the value of the single node, which is `ValueType`. So why have it?

Again, for passing as an argument to another function, e.g. `count(distinct(append(head($.a.*), $.b.*))) > 1`.

> ## The third option
> 
> Only allow nodelist functions to be parameters for other functions, i.e. disallow them from both logicals and comparisons.
> 
> This would only allow something like `$[?count(distinct(@.*))==42]`, which is arguably useful, but both `$[?distinct(@.*)]` and `$[?distinct(@.*)==42]` would be invalid.
> 
> > Whether the spec would be clearer if it avoided type conversions.
> 
> We cannot avoid type conversions. They are implicit in the grammar.
> 
> To prove this, I'll exclude functions.
> 
> **All paths return nodelists. They don't return logicals, and they don't return values.** Singular paths, specifically, return a nodelist with a single value.
> 
> Paths are always allowed to act as logicals. That is, there is a _conversion_ from a nodelist to a logical. Namely, if a nodelist has one or more nodes, it is treated as (converted to) `LogicalTrue`; if it has no nodes, it is treated as (converted to) `LogicalFalse`. If I were to write `$[?@.a]` as a test expression explicitly (in a C-ish syntax), it would be:
> 
> ```
> $[?(LogicalType)@.a]
> ```
> 
> where I'm "casting" the nodelist returned by `@.a` to `LogicalType`.
> 
> Paths are allowed to act as values if they're singular. That is, there is a _conversion_ from a nodelist to a value if the nodelist contains at most a single node. If the nodelist were to contain more than one node, the conversion would be undefined, and a runtime error would occur. To avoid the potential error, the ABNF applies an additional restriction to identify paths which are guaranteed to return at most a single node, and we can catch it at parse time. Writing `$[?@.a==42]` explicitly would be:
> 
> ```
> $[?((ValueType)@.a)==42]  // extra parentheses not necessary, but can help readability
> ```
> 
> where I'm "casting" the nodelist returned by `@.a` to a `ValueType`.
> 
> We have specified these conversions by instead describing their behavior, but that doesn't detract from the fact that they're there.
> 
> Type conversions are unavoidable. We might as well state them. I think it helps us in the end.

Please note that I'm not proposing any change of semantics, merely an editorial change to improve readability. I suggest we defer discussion of the (de)merits of such a change until/unless we have the proposed text before us. If I attempt such a change, I'll take into account the way some readers may think of it in terms of type conversion.

cabo:
> > To change this, we would need to (one of):
> > ```
> > * disallow nodelist functions in comparisons at all
> > 
> > * define a new type that represents a single-node nodelist that _is_ suitable for comparisons
> > 
> > * (there's also a third option)
> > ```
> > 
> > 
Clearly, this needs to be decided, but we should first complete this PR so we have a stable basis for that.

cabo:
I added issue #404 so we can continue on this after completing #403.

glyn:
> I added issue #404 so we can continue on this after completing #403.

@cabo I'm not sure what you are proposing. From my perspective, #404 represents the only blocker to merging #403.

cabo:
> > I added issue #404 so we can continue on this after completing #403.
> 
> @cabo I'm not sure what you are proposing. From my perspective, #404 represents the only blocker to merging #403.

In the spirit of continuous improvement, we can merge #403 and work on #404 from that as a stable basis.
This is not a "blocker", just another work item.

glyn (approved):
Approving with the caveat that #404 is addressed soon.

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1453,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
-* `ValueType` is an abstraction of a primitive value.
-* `BooleanType` is an abstraction of a primitive value that is either
-  `true` or `false`.
-* `OptionalValueType` is an abstraction of a primitive value that may
-  alternatively be absent (`Nothing`).
-* `OptionalNodesType` is an abstraction of a `filter-path` (which appears
+* `ValueType` is an abstraction of a JSON value or `Nothing`.
+* `LogicalType` is an abstraction of the result of a `logical-expr`.
+  Its two instances, `LogicalTrue` and `LogicalFalse`, are not related to
+  the JSON literals `true` and `false` and have no syntactical representation in JSONPath.
```
glyn, line 1485: Post-merge query:

@cabo I'm trying to understand the intent behind the statement "and have no syntactical representation in JSONPath". Why aren't `1 == 1` and `1 == 0` valid representations of `LogicalTrue` and `LogicalFalse`?

draft-ietf-jsonpath-base.md
```diff
@@ -1402,91 +1453,125 @@ According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
...
+* `NodesType` is an abstraction of a `filter-path` (which appears
   in a test expression or as a function argument).
+  Members of `NodesType` have no syntactical representation in JSONPath.
```
glyn, line 1488: Post-merge query:

@cabo Similarly, why isn't a query that always returns an empty nodelist (e.g. `$[?1 == 0]`) a valid syntactical representation of `Nothing` (a member of `NodesType` )?

draft-ietf-jsonpath-base.md
cabo: They semantically yield LT/LF, but aren't syntactical representations.

Add "direct" if that wording isn't sharp enough.

draft-ietf-jsonpath-base.md
cabo: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#discussion_r1122634114

draft-ietf-jsonpath-base.md
glyn: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/413

draft-ietf-jsonpath-base.md
glyn: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/413

PR #406, opened 2023-02-23
Clarify resulting order of nested list operations

cabo:
Add clarifications that the input nodelist provides the outer order, the segments provide a middle-level order, and selectors (if multiple) provide the inner order.

This may have been obvious to some readers, but now should be obvious to more, including yours truly.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -580,7 +580,8 @@ representation as desired by the implementation.
 Note that an empty nodelist is a valid query result.
 
 A segment operates on each of the nodes in its input nodelist in turn,
-and the resultant nodelists are concatenated to produce
+and the resultant nodelists are concatenated in the order of the input
+nodelist they were derived from to produce
 the result of the segment. A node may be selected more than once and
 appears that number of times in the nodelist. Duplicate nodes are not removed.
 
@@ -1684,7 +1685,11 @@ substring exists and `LogicalFalse` otherwise.
 
 ## Segments  {#segments-details}
 
-Segments apply one or more selectors to an input value and concatenate the results into a single nodelist.
+For each node in an input nodelist,
+segments apply one or more selectors to the node and concatenate the
+results of each selector into per-input-node nodelists, which are then
+concatenated in the order of the input nodelist to form a single
+segment result nodelist.
 
 It turns out that the more segments there are in a query, the greater the depth in the input value of the
 nodes of the resultant nodelist:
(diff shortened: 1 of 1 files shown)


glyn (approved):
LGTM thanks. Semantics unchanged, but hopefully a lot clearer.

cabo:
I don't think the example for `$..[*]` in Table 17 is right.

https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-10.html#section-2.7.2.3-3

It says:

``` 
[{"j"=>1, "k"=>2}, [5, 3, [{"j"=>4}, {"k"=>6}]], 1, 2, 5, 3, [{"j"=>4}, {"k"=>6}], {"j"=>4}, 4, {"k"=>6}, 6]
``` 

It should be:

```
[{"j"=>1, "k"=>2}, [5, 3, [{"j"=>4}, {"k"=>6}]], 1, 2, 5, 3, [{"j"=>4}, {"k"=>6}], {"j"=>4}, {"k"=>6}, 4, 6]
```

Note that the descendants list puts out ` [{"j"=>4}, {"k"=>6}]` before `{"j"=>4}` and `{"k"=>6}`.
The wildcard selector then has no choice but outputting `{"j"=>4}`, `{"k"=>6}`, `4`, `6`.

glyn:
> I don't think the example for `$..[*]` in Table 17 is right.
> 
> https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-10.html#section-2.7.2.3-3
> 
> It says:
> 
> ```
> [{"j"=>1, "k"=>2}, [5, 3, [{"j"=>4}, {"k"=>6}]], 1, 2, 5, 3, [{"j"=>4}, {"k"=>6}], {"j"=>4}, 4, {"k"=>6}, 6]
> ```
> 
> It should be:
> 
> ```
> [{"j"=>1, "k"=>2}, [5, 3, [{"j"=>4}, {"k"=>6}]], 1, 2, 5, 3, [{"j"=>4}, {"k"=>6}], {"j"=>4}, {"k"=>6}, 4, 6]
> ```
> 
> Note that the descendants list puts out ` [{"j"=>4}, {"k"=>6}]` before `{"j"=>4}` and `{"k"=>6}`. The wildcard selector then has no choice but outputting `{"j"=>4}`, `{"k"=>6}`, `4`, `6`.

Nice catch! The example needs fixing as does the ordering note below Table 17.

cabo:
I let the computer do the catching :-)
Do you want to make these changes?
I'd do a new issue then.
(And wait a bit for Greg to catch up.)

glyn:
@cabo Yes, I'm happy to do that and please do create a new issue. I won't be able to fix it before the weekend at the earliest though.

PR #409, opened 2023-02-25
Close #400

### LINKED ISSUE #400: Mandatory function implementation
I'm commenting on the editor's draft. Section 2.6 says, of len(), count(), match(), and search(), that  “they are an integral part of the JSONPath specification and are mandatory to implement.”

My intuition is that this is not consistent with the WG charter. I think we're supposed to be specifying the GCD of JSONPath implementations, not enriching the language. The actual words of the charter are “based on the common semantics
and other aspects of existing implementations”. Do len(), count(), match(), and search() qualify?

cabo:
Fix a right margin incursion, as well.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1174,7 +1174,7 @@ test expression is not well-typed.
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
-                      (filter-path / ; path existence or non-existence
+                      (filter-path /  ; path existence/non-existence
                        function-expr) ; LogicalType or
                                       ; NodesType
 filter-path         = rel-path / json-path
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1420,7 +1420,8 @@ This section defines the extension point as well as four function
 extensions that use this extension point.
 While these mechanisms are designed to use the extension point,
 they are an integral part of the JSONPath specification and are
-mandatory to implement.
+expected to be implemented like any other integral part of this
+specification.
```
glyn, line 1424: The change is a bit nuanced from my perspective. What was the rewording trying to get across? It reads about the same to me.

glyn (approved):
Approved with a small query.

draft-ietf-jsonpath-base.md
cabo: Avoiding the big word "mandatory".
(There is a long discussion about compliance vs. interoperability in the IETF.)

glyn: Ok, thanks. Maybe I don't need to know about that discussion, but if you have a link handy, that would be good.

cabo: If you are into internet history, you want to be on the Internet History list: https://elists.isoc.org/mailman/listinfo/internet-history

There are probably dozens, but here is one recent thread that mentioned how interoperability won over compliance:
https://elists.isoc.org/pipermail/internet-history/2022-March/007880.html

Issue #405, opened 2023-02-22
Avoid type conversion language


glyn:
I intend to attempt to build on the merge of #403 with an editorial change (i.e. no syntactic or semantic impact) to avoid the use of type conversion language. The aim is to make the spec clearer, but we can judge the success of that once I've made the attempt.

See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#pullrequestreview-1308453021 and the following short thread for context.

gregsdennis:
@glyn can you please explain your objection to the conversions?  I think they're perfectly clear.  (I'm biased as I wrote it.)  But you seem opposed to the idea of them rather than how they're specified.

Type conversions are a necessary mechanic for comparisons.  I haven't seen a language that doesn't employ conversions in some way.

Even Javascript, a notoriously loosely typed language, has them.  Consider `"42" == 42`.  Javascript allows you to do this, and it will return true, but under the hood it's still doing a type conversion.  Whether it attempts to convert the string to a number or the number to a string, it MUST do some kind of conversion on one or both of the values before comparing the values in order for this to return true.

gregsdennis:
I'd like to explore this by expressing `@.a && @.b == 42` functionally, without "conversion" language.

- `x == y` is replaced with `equals(x, y)`, which takes two `ValueType` parameters and returns `LogicalType`.
- `x && y` is replace with `both(x, y)`, which takes two `LogicalType` parameters and returns `LogicalType`.

Thus, we would write the expression functionally as 

```
both(@.a, equals(@.b, 42))
```

Except we have a type mismatch because `@.a` and `@.b` are `NodesType`, so these functions are not receiving the right types of parameters.

To fix this, we'll need `LogicalType` and `ValueType` from those `NodesType`s, so let's introduce new functions (as @cabo suggested in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404#issuecomment-1440291255): 

- `value(x)` takes a `NodesType` and returns a `ValueType`.
- `logical(x)` takes a `NodesType` and returns a `LogicalType`.

Now we can write the functional version as

```
both(logical(@.a), equals(value(@.b), 42))
```

All of the types align, and this can be executed.

Now, let's remove the first two functions to get back to the expression syntax:

```
logical(@.a) && value(@.b) == 42
```

All of the types align, and this can still be evaluated.

[...]

PR #410, opened 2023-02-25
Potential draft to close #404 -- explicit conversions only

### LINKED ISSUE #404: Confusion by having functions returning nodelists in comparisons
https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#issuecomment-1439581061

Followup issue from #403

cabo:
For discussion as one approach to solve #404

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1113,8 +1113,10 @@ When a function is defined, it is given a unique name, and its return value and 
 The type system is limited in scope; its purpose is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
-The type system also guides conversions ({{type-conv}}) that mimic the
-way different kinds of expressions are handled in the grammar when
+The type system is supported by the two conversion functions `exist`
+and `value` (see Sections {{<exist-fn}} and {{<value-fn}}), which enable
+explicitly performing transitions between different types, mimicking
+the way different kinds of expressions are handled in the grammar when
 function expressions are not in use.
 
 #### Syntax
@@ -1166,20 +1168,18 @@ logical-not-op      = "!"               ; logical NOT operator
 A test expression
 either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
-result of a function expression (see {{fnex}}).
-In the latter case, if the function result type is declared as
-`LogicalType` (see {{typesys}}), it tests whether the result
-is `LogicalTrue`; if the function result type is declared as
-`NodesType`, it tests whether the result is non-empty.
-If the declared function result type is `ValueType`, its use in a
-test expression is not well-typed.
+result of a function expression of declared type `LogicalType` (see
+{{fnex}}).
+In the latter case, the test expression tests whether the function
+result is `LogicalTrue`.
+If the declared function result type is any other than `LogicalType`,
+its use in a test expression is not well-typed.
 
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType or
-                                      ; NodesType
+                       function-expr) ; LogicalType
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1190,15 +1190,16 @@ Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Paths, each of
 which selects at most one node the value of which is then used; and
-function expressions (see {{fnex}}) of type `ValueType` or
-`NodesType` (see {{type-conv}}).
+function expressions (see {{fnex}}) of type `ValueType`.
+If the declared function result type is any other than `ValueType`,
+its use in a comparison expression is not well-typed.
 
 ~~~~ abnf
 comparison-expr     = comparable S comparison-op S comparable
 literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
-                      singular-path /   ; Singular Path value
+                      singular-path /   ; Singular Path -> value
                       function-expr  ; ValueType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
@@ -1451,10 +1452,13 @@ function-argument   = literal /
                       function-expr
 ~~~
 
-A function argument is a `filter-path` or a `comparable`.
+A function argument is a `filter-path` (which yields a `NodesType`), a
+literal (which yields a `ValueType`), or a function expression (which
+yields the declared type of the function named).
 
-According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
-or a `comparable`.
+According to {{filter-selector}}, a `function-expr` is valid as part of
+a test expression (if the declared type is `NodesType`) or as a
+`comparable` (if the declared type is `ValueType`).
 
 Any function expressions in a query must be well-formed (by conforming to the above ABNF)
 and well-typed,
@@ -1465,7 +1469,7 @@ a type system is first introduced.
 
 ### Type System for Function Expressions {#typesys}
 
-Each parameter and the result of a function extension must have a declared type.
+Each parameter as well as the result of a function extension must have a declared type.
 
 A type is a set of instances.
 Declared types enable checking a JSONPath query for well-typedness
@@ -1500,22 +1504,6 @@ The abstract instances above can be obtained from the concrete representations i
 | `Nodes(nl)`       | A list of zero or more nodes, e.g., from a `filter-path` resulting in the nodelist `nl`, which may or may not be empty  |
 {: #tbl-typerep title="Concrete representations of abstract instances"}
 
-### Type Conversions {#type-conv}
-
-The following type conversions may occur:
-
-* Where an expression with a declared type of `NodesType` needs to be
-  converted to a `ValueType`, the conversion proceeds as follows:
-  * If the nodelist contains a single node, the conversion result is
-    the value of the node.
-  * If the nodelist is empty or contains multiple nodes, the
-    conversion result is `Nothing`.
-* Where a member of `NodesType` needs to be converted to a
-  `LogicalType`, the conversion proceeds as follows:
-  * If the nodelist contains one or more nodes, the conversion result
-    is `LogicalTrue`.
-  * If the nodelist is empty, the conversion result is `LogicalFalse`.
-
 The well-typedness of function expressions can now be defined in terms of this type system.
 
 ### Well-Typedness of Function Expressions
@@ -1523,11 +1511,9 @@ The well-typedness of function expressions can now be defined in terms of this t
 A function expression is well-typed if all of the following are true:
 
 * If it occurs directly in a test expression, the function is declared
-  to have a result type of `LogicalType`, or (conversion applies)
-  `NodesType`.
+  to have a result type of `LogicalType`.
 * If it occurs directly as a `comparable` in a comparison, the
-  function is declared to have a result type of `ValueType`, or
-  (conversion applies) `NodesType`.
+  function is declared to have a result type of `ValueType`.
 * Otherwise, it is occurring as an argument to a further function
   expression, and the following rules for function arguments apply to
   its declared result type.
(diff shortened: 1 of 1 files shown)


gregsdennis:
What is the intent behind this edit? Is this final, or a stepping stone toward something else?

I think the edits are fine. They don't change the functionality.  But I still don't see the need for these explicit functions.

cabo:
The intent of this proposal for a technical change is to find out whether we can agree on not having implicit conversions on the function side, but explicit conversions only.  This is intended to solve #404 (and #405 as well).  The explicit conversion functions are needed to replace the implicit conversions.  Of course, we can always decide to change the functionality later, but I wanted to propose a 1:1 drop-in first.

draft-ietf-jsonpath-base.md
```diff
@@ -1449,10 +1452,13 @@ function-argument   = literal /
...
 
-A function argument is a `filter-path` or a `comparable`.
+A function argument is a `filter-path` (which yields a `NodesType`), a
+literal (which yields a `ValueType`), or a function expression (which
+yields the declared type of the function named).
 
-According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
-or a `comparable`.
+According to {{filter-selector}}, a `function-expr` is valid as part of
+a test expression (if the declared type is `NodesType`) or as a
```
glyn, line 1460: Should this say `LogicalType`?

draft-ietf-jsonpath-base.md
```diff
@@ -1535,36 +1523,80 @@ A function expression is well-typed if all of the following are true:
    * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
    * The argument is a Singular Path or `filter-path` (which includes
      Singular Paths), or a function expression with declared result
-     type `NodesType`.  Where the declared type of the parameter is
-     not `NodesType`, a conversion applies.
+     type `NodesType`.
```
glyn, line 1526: This bullet is confusing because it doesn't way what parameter type(s) accept paths and `NodesType` functions. Should it say that the parameter type is `NodesType`?

draft-ietf-jsonpath-base.md
```diff
@@ -1535,36 +1523,80 @@ A function expression is well-typed if all of the following are true:
...
 
-| Function return        | Converted as Value    | Converted as Logical |
-| :-:                    | :-:                   | :-:                  |
-| empty nodelist         | `Nothing`             | `LogicalFalse`       |
-| single-node nodelist   | the node's JSON value | `LogicalTrue`        |
-| multiple-node nodelist | `Nothing`             | `LogicalTrue`        |
-{: #tbl-typeconv title="Conversion of nodelists to values and logicals"}
+Note that a multiple-node nodelist result is deemed `LogicalTrue` in
+the test expression, but cannot be used for obtaining a single
+value and thus feeds the comparison expression with `Nothing`.
```
glyn, line 1547: This is confusing. Shouldn't this text be part of the spec of the `value()` function instead?

draft-ietf-jsonpath-base.md
```diff
@@ -1535,36 +1523,80 @@ A function expression is well-typed if all of the following are true:
...
+node in a single-element nodelist and make that available for further
+processing in the filter expression:
+
+~~~ JSONPath
+$[?value(fn(@.*.price)) >= 7.99]
+~~~
+
+Its only argument is a nodelist.
+If the nodelist contains a single node, the result is the value of the node.
+If the nodelist is empty or contains multiple nodes, the result is `Nothing`.
```
glyn, line 1591: I'm not sure converting multiple nodes to `Nothing` is the best approach. I would prefer to have a type for single node nodelists. Then `value()` could take a parameter of that type and attempting to pass a filter-path` to `value()` would not be well-typed.

draft-ietf-jsonpath-base.md
```diff
@@ -1535,36 +1523,80 @@ A function expression is well-typed if all of the following are true:
...
+
+The effect of the `value` function is similar to the extraction of a
+node value that is performed when a query expression is used as either
+side of a comparison operator.  These query expressions, however, are
+syntactically limited to singular paths.  Here, the argument might be
+a nodelist with multiple nodes: it would be confusing to notate
+queries such as `fn(@..a) == fn(@..b)` without explicit indication
+that an extraction of a single value takes place at either side.
```
glyn, line 1599: This paragraph highlights the difference between the grammar and the type system which I think needs to be addressed by adding a single node nodelist type.

glyn (commented):
This PR is an interesting idea. I like avoiding implicit conversions. I've pointed out what appear to be some inconsistencies. I've also suggested adding a single node nodelist type which I think would add further clarity. Perhaps that should be explored in a separate PR as it's somewhat orthogonal to making all function conversations explicit.

cabo:
> This PR is an interesting idea. I like avoiding implicit conversions. I've pointed out what appear to be some inconsistencies. I've also suggested adding a single node nodelist type which I think would add further clarity. Perhaps that should be explored in a separate PR as it's somewhat orthogonal to making all function conversations explicit.

I thought about adding a fourth type.  I don't think there is any benefit to be expected from that.

The reason we needed a type system mainly is that we need to resolve the ambiguity between looking at a query with respect to the set of nodes produced in order to make statements that are based on that set (e.g., exist()), or whether we really want to access the value of of of these nodes.  Adding a SingularType mostly would need a new conversion function to get there, which I'll call tcboo() [there can only be one], but add nothing to resolving this ambiguity.  Functions that are interested in values already have to deal with the possibility that the nodelist is empty; handling multi-node lists is not a significant added burden.

We needed [did we?] to add singular queries ("paths",  which is wrong terminology) to make comparisons less surprising -- by simply removing the ability to use results from a potentially multi-node query in a comparison, because we have no good default conversion.
With function expressions, we have more abilities, so we no longer need to restrict ourselves.

value(general-query) is a simple escape path if what we are looking for in a comparison is the case where the interesting case of a general-query ("filter-path") would yield one node, plugging that hole for those cases.
Nothing is added by requiring to notate value(tcboo(general-query))...

draft-ietf-jsonpath-base.md
cabo: Right.  Cleaned this up some (the third bullet also was half-redundant with the first).

draft-ietf-jsonpath-base.md
cabo: Went from "the" to "a".

draft-ietf-jsonpath-base.md
cabo: See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/410#issuecomment-1445311058 for my view of this.

draft-ietf-jsonpath-base.md
cabo: (Ditto)

goessner:
I do like the idea of explicit conversions via functional syntax, especially if they "make comparisons less surprising"(cabo). I don't think, we should eliminate all implicit conversions. At least we might keep easily comprehensible ones.

For me, in an ideal world, functions would simply work by JSON in / JSON out. I understand, that with existence tests complexity was introduced.

If having "a type for single node nodelists"(glyn) would eliminate the need for that "Nothing" type I would support this. To me it seems to be a helpless attempt to invent a substitute for JSON's absent `undefined` type.

cabo:
> I do like the idea of explicit conversions via functional syntax, especially if they "make comparisons less surprising"(cabo). 

Good.

> I don't think, we should eliminate all implicit conversions. 

They are still there outside the function extension mechanism.  This PR just removes them from the function extensions.

> At least we might keep easily comprehensible ones.

That is the part that is so hard to predict -- who will think of the implicit conversions when looking at a JSONPath query?

> For me, in an ideal world, functions would simply work by JSON in / JSON out. I understand, that with existence tests complexity was introduced.
> 
> If having "a type for single node nodelists"(glyn) would eliminate the need for that "Nothing" type I would support this. To me it seems to be a helpless attempt to invent a substitute for JSON's absent `undefined` type.

There is no way to have embedded queries without taking account of the fact that queries may come out empty.
So there is no way around something like Nothing (except maybe munching that up with `false` or `null` or `-1` in some way).
I don't think that the way we have found to cope with this is helpless; it is actually quite good.
(BTW, if JSON had `undefined`, we'd need to invent another one...)

gregsdennis:
> They are still there outside the function extension mechanism. This PR just removes them from the function extensions.

This is what I would consider an inconsistency (though less severe than before) in the syntax.  Why allow implicit conversions for @-paths, but not for functions?  Either the implicit conversion exists for a member of one type to a member of another, or it doesn't.  It shouldn't care whether that member came from a function or a path or was explicit.  These should all be treated equally:

```
@.path == 42
functionThatReturnsANodelist(@.a) == 42
<nodelist> == 42  // I know we don't have a way to represent a nodelist; use your imagination
```

If you require an explicit conversion function for one, then you should require an explicit conversion function for all of them.

> That is the part that is so hard to predict -- who will think of the implicit conversions when looking at a JSONPath query?

I suppose people who are familiar with JSON Path syntax.  I've been coding C# for a while, and I "read in" all of the implicit conversions that are present there.  I've even defined a few.  But if I try to read a language I'm unfamiliar with, I'll likely see something that doesn't make sense, so I'll have to go read the docs on it to figure out what is going on.  Once I've learned that, I can see the conversions.

cabo:
On 2023-02-27, at 21:14, Greg Dennis ***@***.***> wrote:
> 
> @.path == 42
> ***@***.***) == 42
> <nodelist> == 42  // I know we don't have a way to represent a nodelist; use your imagination
> 
> If you require an explicit conversion function for one, then you should require an explicit conversion function for all of them.

This is an innocuous inconsistency: If you forget about it, you get an error message and not a silently failing query.

Grüße, Carsten

gregsdennis:
> This is an innocuous inconsistency: If you forget about it, you get an error message and not a silently failing query.

If the problem is the implicit nature, then I suggest we require the conversion function everywhere a conversion takes place.  Don't allow paths to be implicitly converted if functions aren't allowed to be implicitly converted.

I'm just saying that we need to be consistent.  It's not fair to give candy to one toddler and not the others.

cabo:
> On 2023-02-27, at 21:49, Greg Dennis ***@***.***> wrote:
> 
> I'm just saying that we need to be consistent.

No we don’t.  Consistency is the hobgoblin...

> It's not fair to give candy to one toddler and not the others.

That is most definitely not an applicable metaphor here.
Besides, one toddler already had the candy.
I don’t think we want to break the existence tests that are out there.

Grüße, Carsten

hiltontj:
Hey, I've been working on an implementation of this spec in Rust. Working with the specification has been a great experience. Now, I'm trying to follow these issues to see where things will land with the Functions Extension before I implement that part of the spec. I hope you don't mind me chiming in.

@gregsdennis - regarding the treatment of `@.path == 42` and `funcProducingNodelist(@.a) == 42` being equal...

My interpretation of the spec was that explicit conversion is not needed in the former case, because you are only allowed singular paths in comparisons.

So, a JSONPath containing `@.path[*]` or `@.path[1:10]`, or any other non-singular path in a comparison, would be rejected as invalid.

Writing a non-singular path into a comparison would require an explicit function invocation that produces a singular optional node from a nodelist, e.g., `first(@.path[*])` or `only_one(@.path[*])`. The same would be said for `funcProducingNodelist(...)`.

---

On the subject of this PR, I am okay with the idea that existence is implicit in cases like `$.path[?@.foo[*]]`, and forcing the use of a function like `exist` seems a bit awkward. This particular case is much like truthiness in JavaScript and a concept that I think is familiar to developers. Introducing the `exist` function seems to eliminate the need for *Existence Tests*.

An evolved function set, containing explicitly named functions that can perform these conversions, e.g., `is_empty`, `is_single_node`, etc., would allow users to make their queries more specific without any ambiguity.

timbray:
Hey @hiltontj, speaking under my co-chair hat, we do NOT mind you chiming in, this kind of input is super valuable. More, please.

gregsdennis:
> My interpretation of the spec was that explicit conversion is not needed in the former case, because you are only allowed singular paths in comparisons.
> 
> So, a JSONPath containing @.path[*] or @.path[1:10], or any other non-singular path in a comparison, would be rejected as invalid.

Yes, but we don't have a "singular path" or "single nodelist" type (what @glyn wants).  We only have `NodelistType` which is what a path, singular or otherwise, returns.  Because of this, we still need to define a conversion from `NodelistType` to `ValueType` (for comparisons) or `BooleanType` (for tests).

If you want to create a "single nodelist" type, then you can start to have functions that return multiple nodelists and functions that return single nodelists.  Then you can restrict comparisons to only support functions that return this "single nodelist" type.  _Then_ you get a conversion from single-nodelist to value, and because a single-nodelist is also a general nodelist, you get the single-nodelist to boolean for free.

I'd much rather see this worked up than accept these explicit functions.

hiltontj:
> If you want to create a "single nodelist" type, then you can start to have functions that return multiple nodelists and functions that return single nodelists. Then you can restrict comparisons to only support functions that return this "single nodelist" type. Then you get a conversion from single-nodelist to value, and because a single-nodelist is also a general nodelist, you get the single-nodelist to boolean for free.

I think having the single-nodelist type helps clear things up when defining functions for the reasons you described here. 

Similarly, I think specifying a singular path that can only produce a single-nodelist, i.e. a path that is only composed of name and index selectors, is also helpful.

With both defined, implementations can catch when a user tries something like,
```
$.foo[? @.bar[*] == 42 ]
```
and provide an error, e.g.,
```
$.foo[? @.bar[*] == 42 ]
              ^
Error: expected singular path, i.e., one containing only name or
index selectors, and no wildcard, slice, or filter selectors.
```
While at the same time, allowing when a user tries something like,
```
$.books[? @.author == 'Herman Melville']
```
In that latter case, it feels intuitive to use just write `@.author`, because as the person composing the JSONPath, I know the structure of my JSON data.

cabo:
> ```
> $.foo[? @.bar[*] == 42 ]
> ```
> 
> and provide an error, e.g.,

Hi Trevor,

what you describe is already the case -- the grammar ensures that you only can use "singular path" constructions in a comparison.

The discussion we are having is about function extensions.
It occurs in a bit of a vacuum, because we don't have any functions that return nodelists.
We don't want to invent the kitchen sink just to have any; the functions are are defined in the draft and this PR are meant to be the minimum we'll need to get the function extension implemented (and get some other grammar issues moot).

I don't think that a singular nodelist declared type for a function helps at all.  You'd use that in a position where a JSON value is needed -- why not define the function to return that value in the first place?  The two proposed new functions `exist` and `value` are meant to round out the type system, not to be ones you'd use all the time (and certainly not until someone has registered a function with a declared type to return a nodelist).

gregsdennis:
> I don't think that a singular nodelist declared type for a function helps at all. You'd use that in a position where a JSON value is needed -- why not define the function to return that value in the first place?

I raised this before.  The benefit is that a single-nodelist is still a nodelist, so the function would be valid as a test expression as well.

cabo:
> > I don't think that a singular nodelist declared type for a function helps at all. You'd use that in a position where a JSON value is needed -- why not define the function to return that value in the first place?
> 
> I raised this before. The benefit is that a single-nodelist is still a nodelist, so the function would be valid as a test expression as well.

Which is a great reason to use exist()/value() then.

draft-ietf-jsonpath-base.md
glyn: This was fixed, thanks.

glyn:
> > This PR is an interesting idea. I like avoiding implicit conversions. I've pointed out what appear to be some inconsistencies. I've also suggested adding a single node nodelist type which I think would add further clarity. Perhaps that should be explored in a separate PR as it's somewhat orthogonal to making all function conversations explicit.
> 
> I thought about adding a fourth type. I don't think there is any benefit to be expected from that.
> 
> The reason we needed a type system mainly is that we need to resolve the ambiguity between looking at a query with respect to the set of nodes produced in order to make statements that are based on that set (e.g., exist()), or whether we really want to access the value of these nodes.

We added a type system so that certain usages of function expressions would be invalid and raise an error. This then avoids having to define a (non-error) query result which (some) people would find surprising.

We've outlawed non-singular paths in comparisons, so I'm suggesting we do similar with function expressions.

> Adding a SingularType mostly would need a new conversion function to get there,

I'm not sure where this is trying to get. I'd like to avoid the surprise factor of a multi-node list being treated the same as if it was empty.

> which I'll call tcboo() [there can only be one], but add nothing to resolving this ambiguity. Functions that are interested in values already have to deal with the possibility that the nodelist is empty; handling multi-node lists is not a significant added burden.
> 
> We needed [did we?] to add singular queries ("paths", which is wrong terminology) to make comparisons less surprising -- by simply removing the ability to use results from a potentially multi-node query in a comparison, because we have no good default conversion. With function expressions, we have more abilities, so we no longer need to restrict ourselves.

I disagree. Allowing `value()` to take a multi-node argument produces  the same surprise that allowing comparisons of non-singular paths would produce.

> 
> value(general-query) is a simple escape path if what we are looking for in a comparison is the case where the interesting case of a general-query ("filter-path") would yield one node, plugging that hole for those cases. Nothing is added by requiring to notate value(tcboo(general-query))...

I am not proposing a way of feeding a general query into a comparison. I'd like to make that case invalid. We don't need such a `tcboo()` function.

hiltontj:
> what you describe is already the case -- the grammar ensures that you only can use "singular path" constructions in a comparison.

@cabo - thanks for clearing that up. That was my main concern. Also, I think I understand your point about single-nodelist - it is basically just `Option<Value>`.

gregsdennis:
I prefer #412 over this.

glyn (commented):
See my appraisal here: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404#issuecomment-1465175176.

glyn:
Closing. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404#issuecomment-1465175176 for the rationale behind this.

PR #412, opened 2023-02-28
Apply the same rules to function expressions as to non-function expressions

glyn:
By consistently applying the same rules for non-function expressions to function expressions, implicit type conversion (or at least such language) of function expressions and type conversion functions are not needed.

The rules are:
* Non-emptiness of nodelists constitutes an existence test in a test expression.
* A single-node nodelist represents the value of its node when used in comparisons.

This change increases referential transparency in that certain kinds of expressions with the same result may be used interchangeably.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/405

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1113,8 +1113,10 @@ When a function is defined, it is given a unique name, and its return value and 
 The type system is limited in scope; its purpose is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
-The type system also guides conversions ({{type-conv}}) that mimic the
-way different kinds of expressions are handled in the grammar when
+The type system is supported by the two conversion functions `exist`
+and `value` (see Sections {{<exist-fn}} and {{<value-fn}}), which enable
+explicitly performing transitions between different types, mimicking
+the way different kinds of expressions are handled in the grammar when
 function expressions are not in use.
 
 #### Syntax
@@ -1166,20 +1168,18 @@ logical-not-op      = "!"               ; logical NOT operator
 A test expression
 either tests the existence of a node
 designated by an embedded query (see {{extest}}) or tests the
-result of a function expression (see {{fnex}}).
-In the latter case, if the function result type is declared as
-`LogicalType` (see {{typesys}}), it tests whether the result
-is `LogicalTrue`; if the function result type is declared as
-`NodesType`, it tests whether the result is non-empty.
-If the declared function result type is `ValueType`, its use in a
-test expression is not well-typed.
+result of a function expression of declared type `LogicalType` (see
+{{fnex}}).
+In the latter case, the test expression tests whether the function
+result is `LogicalTrue`.
+If the declared function result type is any other than `LogicalType`,
+its use in a test expression is not well-typed.
 
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType or
-                                      ; NodesType
+                       function-expr) ; LogicalType
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1190,15 +1190,16 @@ Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Paths, each of
 which selects at most one node the value of which is then used; and
-function expressions (see {{fnex}}) of type `ValueType` or
-`NodesType` (see {{type-conv}}).
+function expressions (see {{fnex}}) of type `ValueType`.
+If the declared function result type is any other than `ValueType`,
+its use in a comparison expression is not well-typed.
 
 ~~~~ abnf
 comparison-expr     = comparable S comparison-op S comparable
 literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
-                      singular-path /   ; Singular Path value
+                      singular-path /   ; Singular Path -> value
                       function-expr  ; ValueType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
@@ -1450,10 +1451,13 @@ function-argument   = literal /
                       function-expr
 ~~~
 
-A function argument is a `filter-path` or a `comparable`.
+A function argument is a `filter-path` (which yields a `NodesType`), a
+literal (which yields a `ValueType`), or a function expression (which
+yields the declared type of the function named).
 
-According to {{filter-selector}}, a `function-expr` is valid as a `filter-path`
-or a `comparable`.
+According to {{filter-selector}}, a `function-expr` is valid as part of
+a test expression (if the declared type is `NodesType`) or as a
+`comparable` (if the declared type is `ValueType`).
 
 Any function expressions in a query must be well-formed (by conforming to the above ABNF)
 and well-typed,
@@ -1499,22 +1503,6 @@ The abstract instances above can be obtained from the concrete representations i
 | `Nodes(nl)`       | A list of zero or more nodes, e.g., from a `filter-path` resulting in the nodelist `nl`, which may or may not be empty  |
 {: #tbl-typerep title="Concrete representations of abstract instances"}
 
-### Type Conversions {#type-conv}
-
-The following type conversions may occur:
-
-* Where an expression with a declared type of `NodesType` needs to be
-  converted to a `ValueType`, the conversion proceeds as follows:
-  * If the nodelist contains a single node, the conversion result is
-    the value of the node.
-  * If the nodelist is empty or contains multiple nodes, the
-    conversion result is `Nothing`.
-* Where a member of `NodesType` needs to be converted to a
-  `LogicalType`, the conversion proceeds as follows:
-  * If the nodelist contains one or more nodes, the conversion result
-    is `LogicalTrue`.
-  * If the nodelist is empty, the conversion result is `LogicalFalse`.
-
 The well-typedness of function expressions can now be defined in terms of this type system.
 
 ### Well-Typedness of Function Expressions
@@ -1522,11 +1510,9 @@ The well-typedness of function expressions can now be defined in terms of this t
 A function expression is well-typed if all of the following are true:
 
 * If it occurs directly in a test expression, the function is declared
-  to have a result type of `LogicalType`, or (conversion applies)
-  `NodesType`.
+  to have a result type of `LogicalType`.
 * If it occurs directly as a `comparable` in a comparison, the
-  function is declared to have a result type of `ValueType`, or
-  (conversion applies) `NodesType`.
+  function is declared to have a result type of `ValueType`.
 * Otherwise, it occurs as an argument in another function
   expression, and the following rules for function arguments apply to
   its declared result type.
@@ -1536,36 +1522,80 @@ A function expression is well-typed if all of the following are true:
    * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
    * The argument is a Singular Path or `filter-path` (which includes
      Singular Paths), or a function expression with declared result
-     type `NodesType`.  Where the declared type of the parameter is
-     not `NodesType`, a conversion applies.
+     type `NodesType`.
 
-#### Conversion example
-{:unnumbered}
+{{tab-exist-value}} summarizes how the two conversion functions can be
+used to make use of functions returning nodelists, using a
+hypothetical function `fn` of declared type `NodesType`: The table
+gives the result of the conversion function, and the effect of that
+result on the expression given. (`fn` is a hypothetical function
+extension that takes and results in `NodesType`; this function will
+also be used in example further below.)
 
-While functions returning `NodesType` can appear in both comparisons and test expressions,
-path authors must be aware of the conversions and the potential outcomes.
-Specifically, nodelists convert to values and logicals according to {{tbl-typeconv}}:
 
-| Function return        | Converted as Value    | Converted as Logical |
-| :-:                    | :-:                   | :-:                  |
-| empty nodelist         | `Nothing`             | `LogicalFalse`       |
-| single-node nodelist   | the node's JSON value | `LogicalTrue`        |
-| multiple-node nodelist | `Nothing`             | `LogicalTrue`        |
-{: #tbl-typeconv title="Conversion of nodelists to values and logicals"}
-
-For example, given a function `myfunc()` of declared result type
-`NodesType`, the following filter expressions are both valid but can
-have different and unexpected results:
-
-| function result        | `myfunc(@.a)` evaluation (as a test expression)                              | `myfunc(@.a)==42` evaluation (in a comparison)                                |
-| :-:                    | :-                                                                           | :-                                                                            |
-| empty nodelist         | The nodelist is converted to `LogicalFalse`.<br>The expression result is false. | The empty nodelist is converted to `Nothing`.<br>The expression result is false. |
-| single-node nodelist   | The nodelist is converted to `LogicalTrue`.<br>The expression result is true.   | The nodelist is converted to the node's value.<br>The expression result is true. |
-| multiple-node nodelist | The nodelist is converted to `LogicalTrue`.<br>The expression result is true.   | The nodelist is converted to `Nothing`.<br>The expression result is false.       |
+| fn(...) result         | `exist(fn(@..a))` evaluation (as a test expression) | `value(fn(@..a))==42` evaluation (in a comparison)             |
+| :-:                    | :-                                                  | :-                                                             |
+| empty nodelist         | `LogicalFalse`.<br>The test fails.                     | `Nothing`.<br>The expression result is false.                     |
+| single-node nodelist   | `LogicalTrue`.<br>The test succeeds.                   | The node's value.<br>The expression result is true if that is 42. |
+| multiple-node nodelist | `LogicalTrue`.<br>The test succeeds.                   | `Nothing`.<br>The expression result is false.                     |
+{: #tab-exist-value title="Using a function returning a `NodesType` in
+test expressions and comparisons"}
 
 Note that a multiple-node nodelist result is deemed `LogicalTrue` in
-the test expression, but cannot be used for conversion to a single
-value and thus feeds a comparison with `Nothing`.
+the test expression, but cannot be used for obtaining a single
+value and thus feeds the comparison expression with `Nothing`.
+
+
+### `exist` Function Extension {#exist-fn}
+
+Parameters:
+: 1. `NodesType`
+
+Result:
+: `LogicalType`
+
+The "exist" function extension returns `LogicalTrue` if the nodelist
+given consists of at least one node, `LogicalFalse` otherwise:
+
+~~~ JSONPath
+$[?exist(fn(@.*.price))]
+~~~
+
+The effect of the `exist` function is similar to the existence test
+that is performed when a query expression is directly used in a test
+expression:
+If the nodelist contains one or more nodes, the result is
+`LogicalTrue`.
+If the nodelist is empty, the result is `LogicalFalse`.
+
+
+### `value` Function Extension {#value-fn}
+
+Parameters:
+: 1. `NodesType`
+
+Result:
+: `ValueType`
+
+The "value" function extension provides a way to obtain the value of a
+node in a single-element nodelist and make that available for further
+processing in the filter expression:
+
+~~~ JSONPath
+$[?value(fn(@.*.price)) >= 7.99]
+~~~
+
+Its only argument is a nodelist.
+If the nodelist contains a single node, the result is the value of the node.
+If the nodelist is empty or contains multiple nodes, the result is `Nothing`.
+
+The effect of the `value` function is similar to the extraction of a
+node value that is performed when a query expression is used as either
+side of a comparison operator.  These query expressions, however, are
+syntactically limited to singular paths.  Here, the argument might be
+a nodelist with multiple nodes: it would be confusing to notate
+queries such as `fn(@..a) == fn(@..b)` without explicit indication
+that an extraction of a single value takes place at either side.
 
 
 ### `length` Function Extension {#length}
@@ -1680,9 +1710,9 @@ substring exists and `LogicalFalse` otherwise.
 | `$[?length(@.*) < 3]` | not well-typed since `@.*` is a non-singular path |
 | `$[?count(@.*) == 1]` | well-typed |
 | `$[?count(1) == 1]` | not well-typed since `1` is not a path  |
-| `$[?count(foo(@.*)) == 1]` | well-typed, where `foo` is a function extension with a parameter of type `NodesType` and result type `NodesType` |
+| `$[?count(fn(@.*)) == 1]`  | well-typed, where `fn` is a function extension with a parameter of type `NodesType` and result type `NodesType` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | well-typed |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals may not be used in comparisons |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as `LogicalType` may not be used in comparisons |
 {: title="Function expression examples"}
 
 ## Segments  {#segments-details}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1456,7 +1456,7 @@ literal (which yields a `ValueType`), or a function expression (which
 yields the declared type of the function named).
 
 According to {{filter-selector}}, a `function-expr` is valid as part of
-a test expression (if the declared type is `NodesType`) or as a
+a test expression (if the declared type is `LogicalType`) or as a
 `comparable` (if the declared type is `ValueType`).
 
 Any function expressions in a query must be well-formed (by conforming to the above ABNF)
@@ -1519,10 +1519,9 @@ A function expression is well-typed if all of the following are true:
 * Each argument of the function can be used for the declared type of the corresponding declared
   parameter according to one of the following rules:
    * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
-   * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
+   * The argument is a literal primitive value and the declared type of the parameter is `ValueType`.
    * The argument is a Singular Path or `filter-path` (which includes
-     Singular Paths), or a function expression with declared result
-     type `NodesType`.
+     Singular Paths) and the declared type of the parameter is `NodesType`.
 
 {{tab-exist-value}} summarizes how the two conversion functions can be
 used to make use of functions returning nodelists, using a
@@ -1542,8 +1541,8 @@ also be used in example further below.)
 test expressions and comparisons"}
 
 Note that a multiple-node nodelist result is deemed `LogicalTrue` in
-the test expression, but cannot be used for obtaining a single
-value and thus feeds the comparison expression with `Nothing`.
+a test expression, but cannot be used for obtaining a single
+value and thus feeds a comparison expression with `Nothing`.
 
 
 ### `exist` Function Extension {#exist-fn}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1113,11 +1113,6 @@ When a function is defined, it is given a unique name, and its return value and 
 The type system is limited in scope; its purpose is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
-The type system is supported by the two conversion functions `exist`
-and `value` (see Sections {{<exist-fn}} and {{<value-fn}}), which enable
-explicitly performing transitions between different types, mimicking
-the way different kinds of expressions are handled in the grammar when
-function expressions are not in use.
 
 #### Syntax
 {: unnumbered}
@@ -1166,20 +1161,22 @@ logical-not-op      = "!"               ; logical NOT operator
 ~~~~
 
 A test expression
-either tests the existence of a node
-designated by an embedded query (see {{extest}}) or tests the
+either tests the existence of nodes
+designated by an embedded query or by a function expression of declared type
+`SingleNodeType` or `NodesType` (see {{extest}}) or tests the
 result of a function expression of declared type `LogicalType` (see
 {{fnex}}).
 In the latter case, the test expression tests whether the function
 result is `LogicalTrue`.
-If the declared function result type is any other than `LogicalType`,
+If a declared function result type is any other than `SingleNodeType`, `NodesType`, or `LogicalType`,
 its use in a test expression is not well-typed.
 
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType
+                       function-expr) ; SingleNodeType/NodesType
+                                      ; existence or LogicalType
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1190,8 +1187,8 @@ Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Paths, each of
 which selects at most one node the value of which is then used; and
-function expressions (see {{fnex}}) of type `ValueType`.
-If the declared function result type is any other than `ValueType`,
+function expressions (see {{fnex}}) of type `ValueType` or `SingleNodeType`.
+If the declared function result type is any other than `ValueType` or `SingleNodeType`,
 its use in a comparison expression is not well-typed.
 
 ~~~~ abnf
@@ -1200,7 +1197,7 @@ literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
                       singular-path /   ; Singular Path -> value
-                      function-expr  ; ValueType
+                      function-expr  ; ValueType/SingleNodeType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
                       "<"  / ">"
@@ -1258,13 +1255,16 @@ Children of an array appear in array order in the resultant nodelist.
 {: unnumbered}
 
 A path by itself in a Logical context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
+A function expression with result type `SingleNodeType` or `NodesType` in a Logical context is an existence test
+which yields true if the function returns a non-empty nodelist and yields false if the function returns an empty node-list.
 
 Existence tests differ from comparisons in that:
 
 * they work with arbitrary relative or absolute paths (not just Singular Paths).
 * they work with paths that select structured values.
+* they work with function expressions with result type `SingleNodeType` or `NodesType` (not just `SingleNodeType`).
 
-To examine the value of a node selected by a path, an explicit comparison is necessary.
+To examine the value of a node selected by a path or returned by a function expression, an explicit comparison is necessary.
 For example, to test whether the node selected by the path `@.foo` has the value `null`, use `@.foo == null` (see {{null-semantics}})
 rather than the negated existence test `!@.foo` (which yields false if `@.foo` selects a node, regardless of the node's value).
 
@@ -1456,8 +1456,8 @@ literal (which yields a `ValueType`), or a function expression (which
 yields the declared type of the function named).
 
 According to {{filter-selector}}, a `function-expr` is valid as part of
-a test expression (if the declared type is `LogicalType`) or as a
-`comparable` (if the declared type is `ValueType`).
+a test expression (if the declared type is `SingleNodeType`, `NodesType`, or `LogicalType`) or as a
+`comparable` (if the declared type is `ValueType` or `SingleNodeType`).
 
 Any function expressions in a query must be well-formed (by conforming to the above ABNF)
 and well-typed,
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1463,19 +1470,21 @@ a type system is first introduced.
...
-{{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, and `nl` denotes
+{{tbl-types}} defines the available types in terms of abstract instances, where `n` denotes a node, `v` denotes a value, `nl1` denotes a nodelist
+with at most one node, and `nl` denotes
 a nodelist.
 
 | Type                 | Abstract Instances                       |
 | :--                  | :----------------                        |
 | `ValueType`          | `Value(v)`, `Nothing`                    |
 | `LogicalType`        | `LogicalTrue`, `LogicalFalse`            |
+| `SingleNodeType`     | `Nodes(nl1)`                             |
```
goessner, line 1487: perhaps pragmatically name it `NodeType`.

glyn: Perhaps. I did consider that but I thought it was too easy to confuse `NodeType` and `NodesType`. I also liked having some clue that the type is related to `singular-path`.

goessner: hmm ... the reader is sensitised then and will pay attention to the difference.

But the relation to `singular-path` might be a good reason to keep the term of your choice.

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,44 @@ Notes:
...
-  (conversion applies) `NodesType`.
+  function is declared to have a result type of `ValueType` or `SingleNodeType`.
 * Otherwise, it is occurring as an argument to a further function
   expression, and the following rules for function arguments apply to
   its declared result type.
 * Each argument of the function can be used for the declared type of the corresponding declared
   parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
-   * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
+   * The argument is a function expression with declared result type that is the same as,
```
cabo, line 1529: How does this work without exist()/value()?

draft-ietf-jsonpath-base.md
```diff
@@ -1255,13 +1255,16 @@ Children of an array appear in array order in the resultant nodelist.
...
+which yields true if the function returns a non-empty nodelist and yields false if the function returns an empty node-list.
 
 Existence tests differ from comparisons in that:
 
 * they work with arbitrary relative or absolute paths (not just Singular Paths).
 * they work with paths that select structured values.
+* they work with function expressions with result type `SingleNodeType` or `NodesType` (not just `SingleNodeType`).
 
-To test the value of a node selected by a path, an explicit comparison is necessary.
+To test the value of a node selected by a path or returned by a function expression, an explicit comparison is necessary.
```
goessner, line 1267: That function expression might return either

* a value
* a singular-path

right ?

draft-ietf-jsonpath-base.md
```diff
@@ -1270,14 +1273,15 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
 
 The comparison operators `==` and `<` are defined first and then these are used to define `!=`, `<=`, `>`, and `>=`.
 
-When a path resulting in an empty nodelist appears on either side of a comparison:
+When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two paths each of which result in an empty nodelist.
+is between two paths or function expressions each of which result in an empty nodelist.
```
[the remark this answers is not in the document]
goessner: yes ... I am also stumbling over it ...

draft-ietf-jsonpath-base.md
glyn: Sorry, but the bullet on line 1529 is talking about passing function expressions as arguments, so I don't see the relevance of `exist()`/`value()`. Were you asking about a different bullet(s)?

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: How about "involves only paths or functions which result..."?

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,44 @@ Notes:
 * `LogicalType` is an abstraction of the result of a `logical-expr`.
   Its two instances, `LogicalTrue` and `LogicalFalse`, are not related to
   the JSON literals `true` and `false` and have no syntactical representation in JSONPath.
+* `SingleNodeType` is an abstraction of a `singular-path` (which appears in a comparison,
+  in a test expression, or as a function argument).
 * `NodesType` is an abstraction of a `filter-path` (which appears
   in a test expression or as a function argument).
   Members of `NodesType` have no syntactical representation in JSONPath.
+* `SingleNodeType` is a subtype of `NodesType`. An instance of `SingleNodeType` is well-typed
+  wherever an instance of `NodesType` is well-typed.
```
gregsdennis, line 1503: I don't like the "subtype" language.  It implies type inheritance, which is not a feature that our type system has.

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,44 @@ Notes:
...
 
 The abstract instances above can be obtained from the concrete representations in {{tbl-typerep}}.
 
 | Abstract Instance | Concrete Representations                                           |
 | :---------------: | :----------------------:                                           |
 | `Value(v)`        | JSON value `v`                                                     |
 | `Nothing`         | A representation of the absence of a JSON value, distinct from the JSON literal `null`, e.g., from a Singular Path or `filter-path` resulting in an empty nodelist   |
-| `Nodes(nl)`       | A list of zero or more nodes, e.g., from a `filter-path` resulting in the nodelist `nl`, which may or may not be empty  |
+| `Nodes(nl)`       | A list of zero or more nodes, e.g., from a `filter-path` or a `singular-path` resulting in the nodelist `nl`, which may or may not be empty  |
```
gregsdennis, line 1511: We may need to add `Nodes(nl1)` (or maybe just `nl1`) since that's used as the abstraction for `SingleNodeType`

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,44 @@ Notes:
...
+  function is declared to have a result type of `ValueType` or `SingleNodeType`.
 * Otherwise, it is occurring as an argument to a further function
   expression, and the following rules for function arguments apply to
   its declared result type.
 * Each argument of the function can be used for the declared type of the corresponding declared
   parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
-   * The argument is a literal primitive value and the defined type of the parameter is `ValueType`.
+   * The argument is a function expression with declared result type that is the same as,
+     or a subtype of, the declared type of the parameter.
```
gregsdennis, line 1530: Again, I dislike the "subtype" term.

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,44 @@ Notes:
...
-
-For example, given a function `myfunc()` of declared result type
-`NodesType`, the following filter expressions are both valid but can
-have different and unexpected results:
-
-| function result        | `myfunc(@.a)` evaluation (as a test expression)                              | `myfunc(@.a)==42` evaluation (in a comparison)                                |
-| :-:                    | :-                                                                           | :-                                                                            |
-| empty nodelist         | The nodelist is converted to `LogicalFalse`.<br>The expression result is false. | The empty nodelist is converted to `Nothing`.<br>The expression result is false. |
-| single-node nodelist   | The nodelist is converted to `LogicalTrue`.<br>The expression result is true.   | The nodelist is converted to the node's value.<br>The expression result is true. |
-| multiple-node nodelist | The nodelist is converted to `LogicalTrue`.<br>The expression result is true.   | The nodelist is converted to `Nothing`.<br>The expression result is false.       |
```
gregsdennis, line 1563: I encourage you to do an exploration of these cases with the system you've defined.  I expect that the table will be the same.  We may still want the table.

gregsdennis (changes requested):
I like this better than #410, though I still have a few comments.

gregsdennis:
> The rules are:
> 
> - Non-emptiness of nodelists constitutes an existence test in a test expression.
> - A single-node nodelist represents the value of its node when used in comparisons.

I would say these rules define conversions.  I don't see á difference except that you're not using the word "conversion."

gregsdennis:
Also, kudos for the branch name.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: Fixed in 318825601e534bffc86fa1b57b3afb76cc8909a2.

draft-ietf-jsonpath-base.md
glyn: In test expressions, valid function expressions return either a LogicalType or a nodelist. They don't return a (JSON) value. The function expressions considered by this sentence are those that return a singular nodelist.

draft-ietf-jsonpath-base.md
glyn: Thanks @gregsdennis. I agree with the conclusions of your table, but I'm trying to avoid conversion language and make the function side of the spec depend on, _and consistent with_, the semantics already defined by the non-function side of the spec.

glyn:
> > The rules are:
> > 
> > * Non-emptiness of nodelists constitutes an existence test in a test expression.
> > * A single-node nodelist represents the value of its node when used in comparisons.
> 
> I would say these rules define conversions. I don't see á difference except that you're not using the word "conversion."

Ok, I think implementation considerations may be confusing things, but let's look at the following text, for example:

> When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
> * a comparison using the operator == yields true if and only if the comparison is between two expressions, each of which is a path or a function expression, and each of which result in an empty nodelist.

This defines the behaviour of a comparison involving at least one empty nodelist _directly_. There is no conversion necessarily involved, although an implementation could perform a conversion if that produces the same behaviour. 

Essentially, it says that comparisons of the form `<empty nodelist> == ...` or `... == <empty nodelist>` yield true if and only if both sides are `<empty nodelist>`.

glyn:
> > The rules are:
> > 
> > * Non-emptiness of nodelists constitutes an existence test in a test expression.
> > * A single-node nodelist represents the value of its node when used in comparisons.
> 
> I would say these rules define conversions. I don't see á difference except that you're not using the word "conversion."


Please see my other [comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/412#issuecomment-1448789303) on this.

draft-ietf-jsonpath-base.md
glyn: I used the term "subtype" as it's a pithy way of saying that `SingleNodeType` is acceptable everywhere `NodesType` is acceptable. For readers who think in terms of inheritance, a `SingleNodeType` _is a_ `NodesType` (i.e. any value of type `SingleNodeType` is also a value of `NodesType`).

What aspects of inheritance do you think are not applicable to the type system in this PR?

draft-ietf-jsonpath-base.md
```diff
@@ -1276,7 +1276,7 @@ The comparison operators `==` and `<` are defined first and then these are used
 When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two paths or function expressions each of which result in an empty nodelist.
+is between two expressions, each of which is a path or a function expression, and each of which result in an empty nodelist.
```
gregsdennis, line 1279: Is a path unto itself an expression, though?

draft-ietf-jsonpath-base.md
gregsdennis: I admit this is better than the usage than with the previous `Optional*` types.  Meh.  I'll let this go.

draft-ietf-jsonpath-base.md
glyn: I tried adding another row along those lines but I found it very confusing. A `filter-path` can produce a nodelist with one node, so an abstract instance of the form `Nodes(nl1)` doesn't necessarily belong to the `SingleNodeType`. We use types to work out what values can occur, not vice versa.

I'm concerned when you say "... `Nodes(nl1)` ... [is] used as the abstraction for `SingleNodeType`

draft-ietf-jsonpath-base.md
gregsdennis: I removed all of the conversion language from the table.

> ... make the function side of the spec depend on, and consistent with, the semantics already defined by the non-function side of the spec.

So let's replace the headers so that they're not function-centric.

| expression result        | evaluation as a test expression                              | evaluation in a comparison                                |
| :-:                    | :-                                                                           | :-                                                                            |
| empty nodelist         | Interpreted as `LogicalFalse`.<br>The expression result is false. | Interpreted as `Nothing`.<br>The expression result is false. |
| single-node nodelist   | Interpreted as `LogicalTrue`.<br>The expression result is true.   | Interpreted as the node's value.<br>The expression result is true. |
| multiple-node nodelist | Interpreted as `LogicalTrue`.<br>The expression result is true.   | _not well-formed_ |

glyn: I don't think it's necessary to talk about interpreting an empty nodelist as `LogicalFalse` or a non-empty nodelist as `LogicalTrue`. We can simply say that an empty nodelist evaluated as a test expression yields false and the other cases yeild true.

I'm confused by the phrases "The expression result is [false/true]". I _think_ you mean the comparison result is false/true. Right?

draft-ietf-jsonpath-base.md
glyn: No, you're right. `comparable` is a better term. Fixed in 6c62f1e0c8b5d48b745c2a132fb1ebcdc2494203.

draft-ietf-jsonpath-base.md
gregsdennis: Those phrases can be removed if you like, but I think it ties in with the text for filter expressions in general:

> A boolean expression, usually involving the current node, is evaluated and the current node is selected if and only if the expression yields true.

And later, in your edit,

> In the latter case, the test expression tests whether the function result is `LogicalTrue`.

It's assumed from the second statement that there's a distinction between the "true" from the first statement and `LogicalTrue`, but really I think they're the same thing; there's no reason to make a distinction between them.  Maybe that needs to be cleaned up later.

draft-ietf-jsonpath-base.md
```diff
@@ -1276,7 +1276,7 @@ The comparison operators `==` and `<` are defined first and then these are used
 When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two expressions, each of which is a path or a function expression, and each of which result in an empty nodelist.
+is between two `comparable`s, each of which is a path or a function expression, and each of which result in an empty nodelist.
 
 * a comparison using the operator `<` yields false.
```
gregsdennis, line 1281: Editorial:  Does this mean that `>=` yields true?  Can be solved later.

draft-ietf-jsonpath-base.md
```diff
@@ -1270,14 +1273,15 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
 
 The comparison operators `==` and `<` are defined first and then these are used to define `!=`, `<=`, `>`, and `>=`.
 
-When a path resulting in an empty nodelist appears on either side of a comparison:
+When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two paths each of which result in an empty nodelist.
+is between two `comparable`s, each of which is a path or a function expression, and each of which result in an empty nodelist.
```
gregsdennis, line 1279: I think the context of line 1276 allows us to simplify this

> When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
> 
> * a comparison using the operator `==` yields true if and only if both operands, whether path or function expression, result in an empty nodelist.

We already have from the 1276 the condition that _one_ side is an empty nodelist.  We just need to say that the other one is as well.

Nit: I don't really care for "function _expression_."  I much prefer just "function."  The whole thing is an expression that can contain a function.

draft-ietf-jsonpath-base.md
cabo: This is confused.  The table is about values, not types.

cabo:
> Essentially, it says that comparisons of the form `<empty nodelist> == ...` or `... == <empty nodelist>` yield true if and only if both sides are `<empty nodelist>`.

You never get to compare nodelists.  You always compare comparables, which are ValueType.  These have "Nothing", not "empty nodelist".

draft-ietf-jsonpath-base.md
cabo: There are 16 combinations.  5 of them are explained.  You need functions like exist() and value() to make a few more work.

gregsdennis:
> You never get to compare nodelists. You always compare comparables, which are ValueType. These have "Nothing", not "empty nodelist".

The simple solution is to make the empty nodelist and the single nodelist `comparable`...

draft-ietf-jsonpath-base.md
cabo: It probably helps if you discuss the declared types, which is what the type system constrains.  So you are saying that test expressions take declared types of L, S, and N.  Not V.

draft-ietf-jsonpath-base.md
cabo: > I used the term "subtype" as it's a pithy way of saying that `SingleNodeType` is acceptable everywhere `NodesType` is acceptable. 

I think I also would feel more comfortable if we said just that.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
cabo: Thanks!

draft-ietf-jsonpath-base.md
cabo: We have defined `>=` as `==` || `>`, where `>` is `swap and <`.
Yes, it hurts that `>=` ≠ `! <`, but that's where we are.

gregsdennis: I actually think `swap and <` is good.

`"foo" < 4` should yield false because they're incompatible.  Similarly, `"foo" >= 4` should also yield false for the same reason.  Replacing `"foo"` with `Nothing` should give the same results for the same reasons.

Conversely, if the operands are compatible then `>=` does imply `! <`.

This is set up right.

draft-ietf-jsonpath-base.md
gregsdennis: The table is about whatever we want it to be about.  Its sole purpose is clarification.

draft-ietf-jsonpath-base.md
glyn: That's true and the other combinations are not well-typed.

We don't need to define `exist()` and `value()` now because the pressing use cases for these functions have been dispensed with by this PR.

Also it's important to observe that such conversion functions are just ordinary functions that obey the above rules for well-typedness. Adding those functions wouldn't actually change the rules.

draft-ietf-jsonpath-base.md
glyn: Yes, that may turn out to be necessary, but currently I am limiting the use of types to the function side of the spec. Rolling out the type system across the whole spec is a much larger enterprise than this PR and would probably benefit from being worked on separately (regardless of whether or not we adopt this PR).

One hesitation I have about rolling out the type system across the whole spec is that the spec might then seem more prescriptive and we'd have to be careful to point out that the types are _abstract_ and need not feature in implementations.

cabo:
On 1. Mar 2023, at 09:35, Glyn Normington ***@***.***> wrote:
> 
> We don't need to define exist() and value() now because the pressing use cases for these functions have been dispensed with by this PR.

The problem is that not having these defined upfront makes any design of a function extension likely to contort around not having them.

Grüße, Carsten

draft-ietf-jsonpath-base.md
glyn: Well, given subtyping concerns both of you, I've removed it in 6d968a4f.

draft-ietf-jsonpath-base.md
glyn: My view is that the table is unhelpful. The column headings are misleading. I don't see that the spec needs to talk about _evaluation_ of a test expression. The current language of the spec is that test expressions _yield_ true or false. What this was trying to get across is that test expressions are predicates. Evaluation (literally "turning into a value") feels much more like implementation language.

Similarly, I don't think the various kinds of node lists need to be _evaluated_ in order to specify the outcome of a comparison involving one or more node lists.

draft-ietf-jsonpath-base.md
glyn: > I think the context of line 1276 allows us to simplify this
> 
> > When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
> > 
> > * a comparison using the operator `==` yields true if and only if both operands, whether path or function expression, result in an empty nodelist.
> 
> We already have from the 1276 the condition that _one_ side is an empty nodelist. We just need to say that the other one is as well.

Nice. Adopted.

> 
> Nit: I don't really care for "function _expression_." I much prefer just "function." The whole thing is an expression that can contain a function.

We have to be careful to distinguish between a function and where it is used. Also I wouldn't want `length(@.a) == 17`, for example, to be regarded as a function expression. (I considered "function call" instead of "function expression", but that has  operational connotations which aren't helpful.)

glyn: > > Essentially, it says that comparisons of the form `<empty nodelist> == ...` or `... == <empty nodelist>` yield true if and only if both sides are `<empty nodelist>`.
> 
> You never get to compare nodelists. You always compare comparables, which are ValueType. These have "Nothing", not "empty nodelist".

Yes, I was being a bit sloppy there. But I'd stop short of saying that comparables are ValueType at this point. See my other comment about rolling out the type system across the spec.

glyn:
> > You never get to compare nodelists. You always compare comparables, which are ValueType. These have "Nothing", not "empty nodelist".
> 
> The simple solution is to make the empty nodelist and the single nodelist `comparable`...

I think that confuses syntax and semantics. We don't have nodelist syntax.

draft-ietf-jsonpath-base.md
glyn: > On 1. Mar 2023, at 09:35, Glyn Normington ***@***.***> wrote: We don't need to define exist() and value() now because the pressing use cases for these functions have been dispensed with by this PR.
> The problem is that not having these defined upfront makes any design of a function extension likely to contort around not having them. Grüße, Carsten

Ok, I can see the value of adding some such functions, but I'd want to consider them carefully and make them the subject of a separate issue/PR. I don't think we need them in this PR.

gregsdennis:
> > The simple solution is to make the empty nodelist and the single nodelist comparable...
> 
> I think that confuses syntax and semantics. We don't have nodelist syntax.

Okay.  Said a different way, we make the singular path `comparable`.

draft-ietf-jsonpath-base.md
```diff
@@ -1521,13 +1520,15 @@ A function expression is well-typed if all of the following are true:
...
   its declared result type.
-* Each argument of the function can be used for the declared type of the corresponding declared
+* Each argument of the function conforms to the declared type of the corresponding declared
   parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as,
-     or a subtype of, the declared type of the parameter.
+   * The argument is a function expression with declared result type that is the same as
+     the declared type of the parameter.
+   * The argument is a function expression with declared result type `SingleNodeType` and
+     the declared type of the parameter is `NodesType`.
```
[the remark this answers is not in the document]
glyn: > `SingleNodeType` should be valid for both.

@gregsdennis That sounds plausible, but wouldn't we also need to modify the note below Table 13:

> An instance of `SingleNodeType` is well-typed wherever an instance of `NodesType` is well-typed.

to read:

> An instance of `SingleNodeType` is well-typed wherever an instance of `NodesType` or `ValueType` is well-typed.

?

draft-ietf-jsonpath-base.md
```diff
@@ -1485,87 +1494,45 @@ Notes:
 * `LogicalType` is an abstraction of the result of a `logical-expr`.
   Its two instances, `LogicalTrue` and `LogicalFalse`, are not related to
   the JSON literals `true` and `false` and have no syntactical representation in JSONPath.
+* `SingleNodeType` is an abstraction of a `singular-path` (which appears in a comparison,
+  in a test expression, or as a function argument).
```
glyn, line 1498: Note to self: it may be necessary to add: 
```suggestion
  in a test expression, or as a function argument).
  Members of `SingleNodeType` have no syntactical representation in JSONPath.
```
but I'm puzzling over the similar statements for `LogicalType` (queried [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#discussion_r1122546568)) and `NodesType` (queried [here](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#discussion_r1122548030)).

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: On reflection, it may be better to avoid this implicit type conversion in the function side of the spec and use a conversion function instead (as part of adding various such functions, see https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/412#discussion_r1121407915).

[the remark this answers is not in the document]
gregsdennis: > wouldn't we also need to modify the note below Table 13

Yes, that would need to update as well.

draft-ietf-jsonpath-base.md
gregsdennis: I agree the statement should be there.  My only concern is the possible redundancy of saying it three times.  Maybe there's a way to combine the statements, but I don't mind the repetition.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: > On reflection, it may be better to avoid this implicit type conversion in the function side of the spec and use a conversion function instead (as part of adding various such functions, see https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/412#discussion_r1121407915).

What implicit type conversion do you still see?  You've explicitly specified away all of the conversions.

draft-ietf-jsonpath-base.md
gregsdennis: I think the table can be helpful to readers to clarify things, but we can remove it for now and see how it reads.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: `SingleNodeType` to `ValueType` (in the case where a `SingleNodeType` function expression is used as an argument for a `ValueType` parameter). This isn't covered by the non-function side of the spec.

draft-ietf-jsonpath-base.md
glyn: If we decide this PR is the way to go, then we can apply some editorial polish.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
gregsdennis: Maybe that just needs an explicit statement then.

> Singular Paths and functions which return `SingleNodeType` are valid as arguments for function parameters expressed as `ValueType`.  In these cases, the function parameter receives the value of the single node.

or something.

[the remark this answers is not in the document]
glyn: That would certainly work, but I'm not sure the it makes the spec particularly clear. I'm currently favouring explicit type conversions in the function world and implicit conversions only at the boundary between the function world and the rest of the spec.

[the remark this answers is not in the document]
gregsdennis: > I'm currently favouring explicit type conversions in the function world and implicit conversions only at the boundary between the function world and the rest of the spec.

This is an inconsistency, though.  There's no reason to treat function syntax any differently than the rest of the expression.  This is what I'm hung up on that both of you keep pushing.

Functions are part of the overall expression grammar.  The expression as a whole needs certain rules, and every part of the expression should follow those rules, including inside and outside of functions.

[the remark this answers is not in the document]
glyn: > > I'm currently favouring explicit type conversions in the function world and implicit conversions only at the boundary between the function world and the rest of the spec.
> 
> This is an inconsistency, though.

> There's no reason to treat function syntax any differently than the rest of the expression.

I haven't explored this option because I suspect it would be less understandable than the alternatives we are currently exploring. Since you have a different intuition, please feel free to submit a PR to explore that part of the design space and then we can compare it with the alternatives.

>This is what I'm hung up on that both of you keep pushing.
> 
> Functions are part of the overall expression grammar. The expression as a whole needs certain rules, and every part of the expression should follow those rules, including inside and outside of functions.

Please correct me if I've misunderstood, but I think you are talking about referential transparency (the ability to use expressions with the same value interchangeably) and I agree that would be a valuable property. I'm just not sure how to achieve it in general.

> 
> I think those rules are fine being expressed as conversion; you think they need to be expressed more concretely. I'm fine with either so long as they apply everywhere.

That suggests two possible options in the design space. One is to roll out implicit conversions consistently everywhere which I suspect will not be easy for readers of the spec to understand (see my comment above). The other is to roll out the use of conversion functions everywhere, but I think that would break a lot of existing queries.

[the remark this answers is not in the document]
cabo: I don't think there is any problem with the function expression space to be different (which Greg calls "inconsistent") from the grammar-based space.  You can do more with function expressions, so it is natural that you have to be more specific when using them.  Everything else being equal, you would want to avoid exposing these differences, but everything else is not equal.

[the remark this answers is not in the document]
gregsdennis: > I don't think there is any problem with the function expression space to be different

We're defining an expression grammar.  Rules that we apply in one part need to apply to the whole grammar.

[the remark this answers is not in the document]
gregsdennis: To drive this home, we're describing `NodesType` as an abstraction of `filter-path`, which is a Path.  This implies that Paths have a type, and that type is `NodesType`.

draft-ietf-jsonpath-base.md
```diff
@@ -1162,22 +1161,27 @@ logical-not-op      = "!"               ; logical NOT operator
...
-
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType or
-                                      ; NodesType
+                       logical-func /
+                       nodes-func /
+                       single-node-func)
+
+logical-func        = function-expr ; LogicalType
```
cabo, line 1181: Since you can't syntactically distinguish the declared result types of functions, this doesn't help a lot.

gregsdennis:
ugh... I didn't want to push to this branch!

draft-ietf-jsonpath-base.md
gregsdennis: I'm hoping that there might be a way, but I haven't gotten there, yet.

glyn:
@gregsdennis Now that PR #418 is merged here, would you like to approve this PR?

gregsdennis:
I'll have a look in the morning.

draft-ietf-jsonpath-base.md
glyn: @cabo with the merging of PR #418, hopefully the various combinations are now clear. PTAL.

glyn:
> I'll have a look in the morning, but I think it'd be a good idea for @cabo to do the same and ensure his questions have been addressed.

I think @cabo may prefer https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/420 to this PR, so please could you review that one instead. Meanwhile, I'm going to set this PR to draft until we've decided the outcome of #420.

draft-ietf-jsonpath-base.md
```diff
@@ -1164,22 +1161,22 @@ logical-not-op      = "!"               ; logical NOT operator
...
+its use in a test expression is not well-typed.
 
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType or
-                                      ; NodesType
+                       function-expr) ; SingleNodeType/NodesType
+                                      ; existence or LogicalType
```
gregsdennis, line 1179:
```suggestion
                       function-expr) ; SingleNodeType/NodesType existence or
                                      ; LogicalType
```

It may exceed line wrapping but it's clearer this way.

draft-ietf-jsonpath-base.md
```diff
@@ -1257,13 +1255,16 @@ Children of an array appear in array order in the resultant nodelist.
...
+which yields true if the function returns a non-empty nodelist and yields false if the function returns an empty node-list.
 
 Existence tests differ from comparisons in that:
 
 * they work with arbitrary relative or absolute paths (not just Singular Paths).
 * they work with paths that select structured values.
+* they work with function expressions with result type `SingleNodeType` or `NodesType` (not just `SingleNodeType`).
 
-To test the value of a node selected by a path, an explicit comparison is necessary.
+To test the value of a node selected by a path or returned by a function expression, an explicit comparison is necessary.
```
gregsdennis, line 1267: I think we're changing context from "test expression" to "comparison" here, but that context shift isn't clear.

gregsdennis (approved):
Some editorial nits, but I really like where this has landed.

draft-ietf-jsonpath-base.md
glyn: I agree that would be clearer, but we need to stay within 72 characters. Let's do an editorial change if this PR is merged, noting that the previous line talks about "path existence/non-existence" so is inconsistent with saying just "existence" here.

draft-ietf-jsonpath-base.md
gregsdennis: The word "comparison" makes me think it is a comparison context.

How does sentence this apply to a test expression?  It's talking about testing values, which isn't a thing we do.  To be honest, this whole sentence is confusing, and I think it can probably just be removed.

cabo: s/test/examine/

glyn: I've added that word substitution to PR #420. Please can we continue the rewording there as it's more relevant to the base than the subject of the current PR?

cabo (changes requested):
I think it is now clear that adding value() is superior to adding SingleNodeType.
(Value is not a "conversion function", but an one extraction function out of several that I could come up with to make

```
@..color == "red"
```

useful.  So it solved that longstanding problem, and it provides at least one way to complete the graph of the type system.)

glyn:
> I think it is now clear that adding value() is superior to adding SingleNodeType. (Value is not a "conversion function", but one extraction function out of several that I could come up with to make
> 
> ```
> @..color == "red"
> ```
> 
> useful. So it solved that longstanding problem, and it provides at least one way to complete the graph of the type system.)

I agree. The extra complexity introduced by SingleNodeType isn't justified since we can address issue #404 straightforwardly using the value() function.

gregsdennis:
> I think it is now clear that adding value() is superior to adding SingleNodeType.

It is not clear.  `@..color == "red"` is not useful or meaningful.  The left produces multiple values, and the right is one value.  They are not comparable, and it should be invalid.

If you want to separately carry on and define `value()` to give meaning to `value(@..color) == "red"`, sure, but it should be unrelated to the issue that this is trying to solve.

gregsdennis:
> I've also realised that there isn't a usecase for obtaining a single node where we don't immediately need the value of that node, so SingleNodeType isn't particularly useful on its own. - @glyn 

`SingleNodeType` is the only way to allow a function to be used both in a test expression and a comparison.  You were correct to introduce it.  If we make `match()` and `search()` return `SingleNodeType`, then both of these can be valid and work as expected:

```
$[?match(@.timezone, 'Europe/.*')==@.a]
```

If the match is found, it returns a single nodelist containing the value; otherwise it returns and empty nodelist.

- In a test expression, a found match evaluates to true, and no match evaluates to false.
- In a comparison, a found match evaluates to the matched value, and no match evaluates to `Nothing`.

I highly favor this, and this is the thing we were originally trying to solve anyway.

--

`value()` only requires the user to be explicit about the conversion.  The problems with having multiple-nodelists in comparisons still exist, and I shouldn't have put that in the text in the first place.  Doing so doesn't make sense.

If you still want to have a `value()` function, that's fine, but it needs to be separate from and not a replacement for `SingleNodeType`.

gregsdennis:
> Once we decided that these should return LogicalType there was no need for these functions to be used in comparisons.

I argue that `$[?match(@.timezone, 'Europe/.*')==@.a]` is a legitimate use case that we had to leave behind with my other changes.  Adding `SingleNodeType` re-enables this.  It's a good change.

> SingularNodeType isn't needed and the value() function nicely encapsulates, and makes explicit the usage of, the potentially surprising behaviour that is the subject of issue https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404.

Without `SingleNodeType`, one would have to wrap any nodelist function, even those which will only ever return a single node in `value()`.

I'm arguing that we _still_ need `SingleNodeType`, even if we adopt `value()`.

gregsdennis:
> Allowing a function to be used in both a test expression and a comparison isn't a requirement.

Having functions at all isn't a requirement.  What's your point?

draft-ietf-jsonpath-base.md
gregsdennis: It's only 55 chars.

glyn:
Closing. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404#issuecomment-1465175176 for the rationale behind this.

PR #413, opened 2023-03-03
Sharpen up the absence of syntactical representations

glyn:
There are some syntactical representations, e.g.:

* LogicalTrue can be represented by (1==1).
* LogicalFalse can be represented by (1==0).
* Nothing can be represented by a query that always returns the empty nodelist, such as $[?1==0].

But these are somewhat circuitous rather than direct.

See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/403#discussion_r1122546568

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1484,10 +1484,10 @@ Notes:
 * `ValueType` is an abstraction of a JSON value or `Nothing`.
 * `LogicalType` is an abstraction of the result of a `logical-expr`.
   Its two instances, `LogicalTrue` and `LogicalFalse`, are not related to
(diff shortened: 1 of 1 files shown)


glyn:
@cabo I also considered deleting the statements about there being no (direct) syntactical representation of the values of certain types because I'm not sure they add much. The text in this PR adds an air of mystery - what are "direct syntactical representations"? WDYT?

cabo (approved):
This is obviously an improvement.
We also could say "literals", but then we would assume knowledge about this programming language construct.
So I think this should simply be merged.

gregsdennis:
I'm not sure how "direct" clarifies "syntactical representation."  Either something can be represented in the syntax or it can't.

Merge if you like.

PR #414, opened 2023-03-03
Explicit conversions with single node type (variant of #410)

glyn:
This is a variant of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/410

PR #415, opened 2023-03-03
Add SingleNodeType (variant of #410, equivalent to #414)

glyn:
This avoids the surprise factor when the value() function is passed a multiple-node nodelist.

This is a PR on the branch for [PR ](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/410). See also https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/414 which rolls the two together as a single PR.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -733,15 +733,17 @@ HEXDIG              = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 
 Note: `double-quoted` strings follow the JSON string syntax ({{Section 7 of RFC8259}});
 `single-quoted` strings follow an analogous pattern ({{syntax-index}}).
-No attempt was made to improve on this syntax, so characters with
-scalar values above 0x10000, such as <u format="num-lit-name">🤔</u>, need to be represented
+No attempt was made to improve on this syntax, so if it is desired to
+escape characters with
+scalar values above 0x10000, such as <u format="num-lit-name">🤔</u>,
+they need to be represented
 by a pair of surrogate escapes (`"\uD83E\uDD14"` in this case).
 
 #### Semantics
 {: unnumbered}
 
 A `name-selector` string MUST be converted to a
-member name M by removing the surrounding quotes and
+member name `M` by removing the surrounding quotes and
 replacing each escape sequence with its equivalent Unicode character, as
 in the table below:
 
@@ -760,15 +762,15 @@ in the table below:
 {: title="Escape Sequence Replacements" cols="c c"}
 
 Applying the `name-selector` to an object node
-selects a member value whose name equals the member name M,
+selects a member value whose name equals the member name `M`,
 or selects nothing if there is no such member value.
 Nothing is selected from a value that is not an object.
 
-Note that processing the name selector requires comparing the member name string M
+Note that processing the name selector requires comparing the member name string `M`
 with member name strings in the JSON to which the selector is being applied.
 Two strings MUST be considered equal if and only if they are identical
 sequences of Unicode scalar values. In other words, normalization operations
-MUST NOT be applied to either the member name string M from the JSONPath or to
+MUST NOT be applied to either the member name string `M` from the JSONPath or to
 the member name strings in the JSON prior to comparison.
 
 #### Examples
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1521,29 +1527,43 @@ A function expression is well-typed if all of the following are true:
   parameter according to one of the following rules:
    * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
    * The argument is a literal primitive value and the declared type of the parameter is `ValueType`.
+   * The argument is a Singular Path and the declared type of the parameter is `SingleNodeType`.
    * The argument is a Singular Path or `filter-path` (which includes
      Singular Paths) and the declared type of the parameter is `NodesType`.
 
-{{tab-exist-value}} summarizes how the two conversion functions can be
-used to make use of functions returning nodelists, using a
+{{tab-exist-value}} summarizes how the two conversion functions behave
```
glyn, line 1534: I'm not convinced of the value of the following two tables, but they may be helpful here to emphasise the effect of this PR.

draft-ietf-jsonpath-base.md
```diff
@@ -1568,11 +1588,31 @@ If the nodelist contains one or more nodes, the result is
...
+
+Result:
+: `SingleNodeType`
+
+The "single" function extension provides a way to obtain a
+node in a nodelist and make that available for further
+processing in the filter expression:
+
+~~~ JSONPath
+$[?value(single(fn(@.*.price))) >= 7.99]
```
gregsdennis, line 1604: No one is going to want to type this when `$[?fn(@.*.price) >= 7.99]` is just as understandable.  I fear we're defining this grammar into non-adoption.

draft-ietf-jsonpath-base.md
```diff
@@ -1490,18 +1493,21 @@ Notes:
 * `ValueType` is an abstraction of a JSON value or `Nothing`.
 * `LogicalType` is an abstraction of the result of a `logical-expr`.
   Its two instances, `LogicalTrue` and `LogicalFalse`, are not related to
-  the JSON literals `true` and `false` and have no syntactical representation in JSONPath.
+  the JSON literals `true` and `false` and have no direct syntactical representation in JSONPath.
```
cabo, line 1496: The "direct" stuff should be done directly to #410

draft-ietf-jsonpath-base.md
```diff
@@ -1490,18 +1493,21 @@ Notes:
...
+* `SingleNodeType` is an abstractions of a `singular-path` (which appears in a comparison,
```
cabo, line 1497:
```suggestion
* `SingleNodeType` is an abstraction of a `singular-path` (which appears in a comparison,
```

draft-ietf-jsonpath-base.md
```diff
@@ -1568,11 +1588,31 @@ If the nodelist contains one or more nodes, the result is
...
+The "single" function extension provides a way to obtain a
+node in a nodelist and make that available for further
+processing in the filter expression:
+
+~~~ JSONPath
+$[?value(single(fn(@.*.price))) >= 7.99]
+~~~
+
+Its only argument is a nodelist.
+If the nodelist contains a single node, the result is the value of the node.
```
cabo, line 1608: That's what "value" is supposed to do.
"single" is useful for finding out whether a nodelist is singular

Use `single($.a.*.author)` to see whether there is only one author

draft-ietf-jsonpath-base.md
glyn: The point of this PR is to explore what it looks like to mirror (on the function side) the fact that expressions like `$[?@.*.price >= 7.99]` are not well-formed and to avoid what, in my opinion, is a surprise in #410 of `value` converting a multi-node nodelist to `Nothing`.

**If** (big if), we chose to go in this direction, we could abbreviate `value(single(...))` by providing another function (e.g. `uniqueValue`) which composes `value` with `single`.

If `$[?fn(@.*.price) >= 7.99]` is understandable (remember that "fn is a hypothetical function extension that takes and results in `NodesType`"), then presumably so is `$[?@.*.price >= 7.99]`, but the WG has failed to reach a consensus on what that might mean, so we have ruled it out syntactially.

draft-ietf-jsonpath-base.md
glyn: Agreed, but it should also be done directly to the main branch, independently of #410. Please see/approve/merge https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/413.

draft-ietf-jsonpath-base.md
glyn: The point of this PR is to explore what it looks like to mirror (on the function side) the fact that expressions like `$[?@.*.price >= 7.99]` are not well-formed and to avoid what, in my opinion, is a surprise in #410 of `value` converting a multi-node nodelist to `Nothing`.

Feel free to add a function like `single` (presumably returning `LogicalType`) in #410, but in this PR, that's not the purpose of `single()`, as defined.

glyn:
Closing this, pending the outcome of https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/420.

PR #416, opened 2023-03-05
Gregsdennis/improve well typedness rules

gregsdennis:
Builds on #412

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1172,11 +1172,16 @@ If a declared function result type is any other than `SingleNodeType`, `NodesTyp
 its use in a test expression is not well-typed.
 
 ~~~ abnf
-
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; SingleNodeType/NodesType
-                                      ; existence or LogicalType
+                       logical-func /
+                       nodes-func /
+                       single-node-func)
+
+logical-func        = function-expr ; LogicalType
+nodes-func          = function-expr ; NodesType
+single-node-func    = function-expr ; SingleNodeType
+
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1197,7 +1202,9 @@ literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
                       singular-path /   ; Singular Path -> value
-                      function-expr  ; ValueType/SingleNodeType
+                      value-func /
+                      single-node-func
+value-func          = function-expr  ; ValueType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
                       "<"  / ">"
@@ -1485,7 +1492,7 @@ a nodelist.
 | `ValueType`          | `Value(v)`, `Nothing`                    |
 | `LogicalType`        | `LogicalTrue`, `LogicalFalse`            |
 | `SingleNodeType`     | `Nodes(nl1)`                             |
-| `NodesType`       | `Nodes(nl)`                              |
+| `NodesType`          | `Nodes(nl)`                              |
 {: #tbl-types title="Function extension type system"}
 
 Notes:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1276,7 +1276,7 @@ The comparison operators `==` and `<` are defined first and then these are used 
 When a path or function expression resulting in an empty nodelist appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two expressions, each of which is a path or a function expression, and each of which result in an empty nodelist.
+is between two `comparable`s, each of which is a path or a function expression, and each of which result in an empty nodelist.
 
 * a comparison using the operator `<` yields false.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1522,8 +1522,8 @@ A function expression is well-typed if all of the following are true:
   to have a result type of `LogicalType`, `SingleNodeType`, or `NodesType`.
 * If it occurs directly as a `comparable` in a comparison, the
   function is declared to have a result type of `ValueType` or `SingleNodeType`.
-* If it occurs as an argument to a further function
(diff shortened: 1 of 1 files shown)


cabo:
Oh.  You pushed these changes to #412.

gregsdennis:
Yeah, I didn't want to.  I'm going to revert over on that branch.

draft-ietf-jsonpath-base.md
```diff
@@ -1162,22 +1161,27 @@ logical-not-op      = "!"               ; logical NOT operator
...
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType or
-                                      ; NodesType
+                       logical-func /
+                       nodes-func /
+                       single-node-func)
+
+logical-func        = function-expr ; LogicalType
+nodes-func          = function-expr ; NodesType
+single-node-func    = function-expr ; SingleNodeType
```
gregsdennis, line 1183: I realize this doesn't do much for automated ABNF parsers, but it helps readability, and I hope we might be able to find a way to distinguish these functions syntactically.

gregsdennis: Is there a way to have extensible ABNF?  Maybe we can do something like

```abnf
function-name         = value-func-name /
                        logical-func-name /
                        single-node-func-name /
                        nodes-func-name
value-func-name       = "length" /
                        "count" /
                        {???} ; a template that can be completed by an extension
logical-func-name     = "match"
                        "search" /
                        {???}
single-node-func-name = {???}
nodes-func-name       = {???}
```

I don't know if the `{???}` is possible in ABNF.

cabo: The intention with the extensibility approach was to have a stable grammar (ABNF) plus semantic rules that checks validity of the constructed parse tree, where the latter make use of information from the registry.
I don't think we have technology to import registry information into ABNF (we are looking at that in CDDL, but even that won't be able to extend the grammar to express the semantic constraints, which include constraints on arguments).  Your additions look like they need a painful hand-written parser with a lexer that assigns different syntactical categories based on a symbol table lookup, like in a C compiler.  I'd like to stick with ABNF.

gregsdennis: Okay... I can revert this bit.  I'm not married to it.  It was just an idea.

glyn (changes requested):
This PR seems to have the same effective content as #412.

gregsdennis:
Yes, no functional change, but the language around where functions are valid is more complete.

glyn:
It's hard to see the wood for the trees in this PR, but this diff shows the changes relative to PR #412:
```
diff --git a/draft-ietf-jsonpath-base.md b/draft-ietf-jsonpath-base.md
index 194edfe..6928dcd 100644
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1172,7 +1172,6 @@ If a declared function result type is any other than `SingleNodeType`, `NodesTyp
 its use in a test expression is not well-typed.
 
 ~~~ abnf
-
                       (filter-path /  ; path existence/non-existence
                        function-expr) ; SingleNodeType/NodesType
@@ -1485,7 +1484,7 @@ a nodelist.
 | `ValueType`          | `Value(v)`, `Nothing`                    |
 | `LogicalType`        | `LogicalTrue`, `LogicalFalse`            |
 | `SingleNodeType`     | `Nodes(nl1)`                             |
-| `NodesType`       | `Nodes(nl)`                              |
+| `NodesType`          | `Nodes(nl)`                              |
 
 Notes:
@@ -1499,8 +1498,7 @@ Notes:
   Members of `SingleNodeType` have no direct syntactical representation in JSONPath.
 * `NodesType` is an abstraction of a `filter-path` (which appears
-  Members of `NodesType` have no syntactical representation in JSONPath.
-* An instance of `SingleNodeType` is well-typed wherever an instance of `NodesType` is well-typed.
+  Members of `NodesType` have no direct syntactical representation in JSONPath.
 
 The abstract instances above can be obtained from the concrete representations in {{tbl-typerep}}.
 
@@ -1521,19 +1519,21 @@ A function expression is well-typed if all of the following are true:
   to have a result type of `LogicalType`, `SingleNodeType`, or `NodesType`.
 * If it occurs directly as a `comparable` in a comparison, the
   function is declared to have a result type of `ValueType` or `SingleNodeType`.
-* Otherwise, it occurs as an argument to a further function
-  expression and the following rules for function arguments apply to
-  its declared result type.
 * Each argument of the function conforms to the declared type of the corresponding declared
-  parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as
-     the declared type of the parameter.
-   * The argument is a function expression with declared result type `SingleNodeType` and
-     the declared type of the parameter is `NodesType`.
-   * The argument is a literal primitive value and the declared type of the parameter is `ValueType`.
-   * The argument is a Singular Path and the declared type of the parameter is `SingleNodeType`.
-   * The argument is a Singular Path or `filter-path` (which includes
-     Singular Paths) and the declared type of the parameter is `NodesType`.
+  parameter according to one of the following scenarios:
+   * The declared type of the parameter is `ValueType` and the argument is one of:
+     * a function with declared type of `ValueType` or `SingleNodeType`
+     * a `singular-path`
+     * a `literal`
+   * The declared type of the parameter is `LogicalType` and the argument is one of:
+     * a function with declared type of `LogicalType`, `SingleNodeType`, or `NodesType`
+     * a `filter-path` (which includes `singular-path`)
+   * The declared type of the parameter is `SingleNodeType` and the argument is one of:
+     * a function with declared type of `SingleNodeType`
+     * a `singular-path`
+   * The declared type of the parameter is `NodesType` and the argument is one of:
+     * a function with declared type of `SingleNodeType` or `NodesType`
+     * a `filter-path`
 
 ### `length` Function Extension {#length}
 

```

gregsdennis:
I can change the target branch to yours if you think that'll help.

glyn:
Yes,  or also provide another PR on that base.

glyn:
@gregsdennis General comment: the implicit conversions of function arguments to the corresponding parameter type can't borrow from the text in the non-function side of the spec, so will need some wording. I'm happy to wait and make detailed comments about this when you have changed the target branch, but feel free to address them sooner if my general comment makes sense to you.

gregsdennis:
Looks like changing the base didn't work out well.  I'll create a new PR.

PR #417, opened 2023-03-06
Extract some changes in #410 that are unrelated to conversions

cabo:
Wouldn't mind if we could get these editorials merged quickly so the bulk of the other PRs shrinks a bit.


glyn (approved):
LGTM, thanks.

PR #418, opened 2023-03-06
Improve well typedness rules

gregsdennis:
Builds on #412.  Attempts to address [this comment](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/412#discussion_r1120934952) from @cabo:

> There are 16 combinations. 5 of them are explained. You need functions like exist() and value() to make a few more work.

Only valid/well-formed scenarios are explained; others are considered invalid/poorly-formed.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1521,19 +1521,21 @@ A function expression is well-typed if all of the following are true:
   to have a result type of `LogicalType`, `SingleNodeType`, or `NodesType`.
 * If it occurs directly as a `comparable` in a comparison, the
   function is declared to have a result type of `ValueType` or `SingleNodeType`.
-* Otherwise, it occurs as an argument to a further function
-  expression and the following rules for function arguments apply to
-  its declared result type.
 * Each argument of the function conforms to the declared type of the corresponding declared
-  parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as
-     the declared type of the parameter.
-   * The argument is a function expression with declared result type `SingleNodeType` and
-     the declared type of the parameter is `NodesType`.
-   * The argument is a literal primitive value and the declared type of the parameter is `ValueType`.
-   * The argument is a Singular Path and the declared type of the parameter is `SingleNodeType`.
-   * The argument is a Singular Path or `filter-path` (which includes
-     Singular Paths) and the declared type of the parameter is `NodesType`.
+  parameter according to one of the following scenarios:
+   * The declared type of the parameter is `ValueType` and the argument is one of:
+     * a function with declared type of `ValueType` or `SingleNodeType`
+     * a `singular-path`
+     * a `literal`
+   * The declared type of the parameter is `LogicalType` and the argument is one of:
+     * a function with declared type of `LogicalType`, `SingleNodeType`, or `NodesType`
+     * a `filter-path` (which includes `singular-path`)
+   * The declared type of the parameter is `SingleNodeType` and the argument is one of:
+     * a function with declared type of `SingleNodeType`
+     * a `singular-path`
+   * The declared type of the parameter is `NodesType` and the argument is one of:
+     * a function with declared type of `SingleNodeType` or `NodesType`
+     * a `filter-path`
 
 ### `length` Function Extension {#length}
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1113,11 +1113,6 @@ When a function is defined, it is given a unique name, and its return value and 
 The type system is limited in scope; its purpose is to express
 restrictions that, without functions, are implicit in the grammar of
 filter expressions.
-The type system is supported by the two conversion functions `exist`
-and `value` (see Sections {{<exist-fn}} and {{<value-fn}}), which enable
-explicitly performing transitions between different types, mimicking
-the way different kinds of expressions are handled in the grammar when
-function expressions are not in use.
 
 #### Syntax
 {: unnumbered}
@@ -1166,20 +1161,22 @@ logical-not-op      = "!"               ; logical NOT operator
 ~~~~
 
 A test expression
-either tests the existence of a node
-designated by an embedded query (see {{extest}}) or tests the
+either tests the existence of nodes
+designated by an embedded query or by a function expression of declared type
+`SingleNodeType` or `NodesType` (see {{extest}}) or tests the
 result of a function expression of declared type `LogicalType` (see
 {{fnex}}).
 In the latter case, the test expression tests whether the function
 result is `LogicalTrue`.
-If the declared function result type is any other than `LogicalType`,
+If a declared function result type is any other than `SingleNodeType`, `NodesType`, or `LogicalType`,
 its use in a test expression is not well-typed.
 
 ~~~ abnf
 
 test-expr           = [logical-not-op S]
                       (filter-path /  ; path existence/non-existence
-                       function-expr) ; LogicalType
+                       function-expr) ; SingleNodeType/NodesType
+                                      ; existence or LogicalType
 filter-path         = rel-path / json-path
 rel-path            = current-node-identifier segments
 current-node-identifier = "@"
@@ -1190,8 +1187,8 @@ Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Paths, each of
 which selects at most one node the value of which is then used; and
-function expressions (see {{fnex}}) of type `ValueType`.
-If the declared function result type is any other than `ValueType`,
+function expressions (see {{fnex}}) of type `ValueType` or `SingleNodeType`.
+If the declared function result type is any other than `ValueType` or `SingleNodeType`,
 its use in a comparison expression is not well-typed.
 
 ~~~~ abnf
@@ -1200,7 +1197,7 @@ literal             = number / string-literal /
                       true / false / null
 comparable          = literal /
                       singular-path /   ; Singular Path -> value
-                      function-expr  ; ValueType
+                      function-expr  ; ValueType/SingleNodeType
 comparison-op       = "==" / "!=" /
                       "<=" / ">=" /
                       "<"  / ">"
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1521,19 +1521,21 @@ A function expression is well-typed if all of the following are true:
...
-  parameter according to one of the following rules:
-   * The argument is a function expression with declared result type that is the same as
-     the declared type of the parameter.
-   * The argument is a function expression with declared result type `SingleNodeType` and
-     the declared type of the parameter is `NodesType`.
-   * The argument is a literal primitive value and the declared type of the parameter is `ValueType`.
-   * The argument is a Singular Path and the declared type of the parameter is `SingleNodeType`.
-   * The argument is a Singular Path or `filter-path` (which includes
-     Singular Paths) and the declared type of the parameter is `NodesType`.
+  parameter according to one of the following scenarios:
```
glyn, line 1525: "scenarios" is a bit vague. What's wrong with "rules"?

draft-ietf-jsonpath-base.md
```diff
@@ -1521,19 +1521,21 @@ A function expression is well-typed if all of the following are true:
...
+   * The declared type of the parameter is `ValueType` and the argument is one of:
+     * a function with declared type of `ValueType` or `SingleNodeType`
+     * a `singular-path`
```
glyn, line 1528: The PR needs to explain how to get from `SingleNodeType` and `singular-path` to `ValueType`. C.f. the following text from the "Comparisons" section:
> When any path on either side of a comparison results in a nodelist consisting of a single node, each such path is replaced by the value of its node

draft-ietf-jsonpath-base.md
```diff
@@ -1521,19 +1521,21 @@ A function expression is well-typed if all of the following are true:
...
+     * a `literal`
+   * The declared type of the parameter is `LogicalType` and the argument is one of:
+     * a function with declared type of `LogicalType`, `SingleNodeType`, or `NodesType`
+     * a `filter-path` (which includes `singular-path`)
```
glyn, line 1532: The PR needs to explain how to get from `SingleNodeType`, `NodesType`, and `filter-path` to `LogicalType`. C.f. the following text from "Existence Tests":
> A path by itself in a Logical context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.

glyn (changes requested):
The PR needs to generalise the rules which apply to existence tests and comparisons. This will necessarily reintroduce certain conversions, which isn't necessarily a deal-breaker if it can be done cleanly.

draft-ietf-jsonpath-base.md
gregsdennis: "Rules" contain a scenario and an outcome.  "If X, then Y" is a rule.  "X" is a scenario, and "Y" is the outcome that applies when that scenario exists.

> The declared type of the parameter is `ValueType` and the argument is one of...

is a scenario.  The outcome is indicated above:

> Each argument of the function conforms to the declared type of the corresponding declared
  parameter

Those _together_ are a rule.

draft-ietf-jsonpath-base.md
gregsdennis: The text on this branch actually reads:

> When any path **or function expression** on either side of a comparison results in a nodelist consisting of a single node, each such path **or function expression** is replaced by the value of its node and then...

so it's already there.

draft-ietf-jsonpath-base.md
gregsdennis: The text on this branch actually reads:

> A path by itself in a Logical context is an existence test which yields true if the path selects at least one node and yields false if the path does not select any nodes.
> A function expression with result type `SingleNodeType` or `NodesType` in a Logical context is an existence test which yields true if the function returns a non-empty nodelist and yields false if the function returns an empty node-list.

so it's already there.

gregsdennis:
> The PR needs to generalise the rules which apply to existence tests and comparisons.

I don't follow.  How does the current text not do this?

> This will necessarily reintroduce certain conversions...

I'm not sure it does.  We've been careful to explicitly declare the behavior of everything where it's valid/well-formed.  (Personally I think the conversions are clearer, but that's not what this PR is about.)

draft-ietf-jsonpath-base.md
gregsdennis: We _could_ reword this slightly to be a rule if you prefer:

>  * If the declared type of the parameter is `ValueType`, then the argument must be one of:
>    * a function with declared type of `ValueType` or `SingleNodeType`
>    * a `singular-path`
>    * a `literal`

... etc.

draft-ietf-jsonpath-base.md
gregsdennis: > supplying an argument to a parameter with type `LogicalType` has any connection with a "logical context"

I don't understand.  How can anywhere `LogicalType` is used _not_ be a "logical context?"

draft-ietf-jsonpath-base.md
glyn: Certainly that text is the kind of thing that is needed. But that text limits its application to either side of a comparison and this excludes arguments passed as parameters of type `ValueType`.

draft-ietf-jsonpath-base.md
glyn: Because we haven't actually defined "logical context", so it's somewhat open to interpretation. My interpretation in the context of the existence test section is that "logical context" refers to "logical-expr":
```
logical-expr        = logical-or-expr
logical-or-expr     = logical-and-expr *(S "||" S logical-and-expr)
                        ; disjunction
                        ; binds less tightly than conjunction
logical-and-expr    = basic-expr *(S "&&" S basic-expr)
                        ; conjunction

basic-expr          = paren-expr /
                      comparison-expr /
                      test-expr

```

draft-ietf-jsonpath-base.md
gregsdennis: Addressed in [ba8c16b](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/418/commits/ba8c16bf51f0aa495c605ca16e46006a1dbb3340)

draft-ietf-jsonpath-base.md
gregsdennis: Addressed in [ba8c16b](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/418/commits/ba8c16bf51f0aa495c605ca16e46006a1dbb3340)

draft-ietf-jsonpath-base.md
glyn: I get it now, thanks. Actually I suggest:
```suggestion
  parameter according to one of the following:
```

draft-ietf-jsonpath-base.md
glyn: Yes, that does the job nicely.

draft-ietf-jsonpath-base.md
glyn: Yes, that does the job nicely.

draft-ietf-jsonpath-base.md
```diff
@@ -1521,19 +1521,27 @@ A function expression is well-typed if all of the following are true:
...
+   * The declared type of the parameter is `SingleNodeType` and the argument is one of:
+     * a function with declared type of `SingleNodeType`
+     * a `singular-path`
+   * The declared type of the parameter is `NodesType` and the argument is one of:
+     * a function with declared type of `SingleNodeType` or `NodesType`
```
glyn, line 1537: The other scenarios treat the parameter type first, so:
```suggestion
     * a function with declared type of `NodesType` or `SingleNodeType`
```

draft-ietf-jsonpath-base.md
```diff
@@ -1521,19 +1521,27 @@ A function expression is well-typed if all of the following are true:
...
+     * a `filter-path`
+
+### Processing Nodelists as Arguments
+
+When any `singular-path` or function expression with declared type of `SingleNodeType` appears as an argument for a parameter of type `ValueType`, each such path or function expression is replaced by the value of its node or `Nothing` if it is empty.
```
glyn, line 1542: There's an editorial wrinkle in that `Nothing` isn't really empty. How about the following wordier, but hopefully clearer possibility? If you don't like this, I'd be happy to apply it as an editorial tweak if this PR is merged.
```suggestion
When any `singular-path` appears as an argument for a parameter of type `ValueType`, the path is replaced by the value of the node in the resulting nodelist or `Nothing` if the resulting nodelist is empty.

When any function expression with declared type of `SingleNodeType` appears as an argument for a parameter of type `ValueType`, the function expression is replaced by the value of the node in the resulting nodelist or `Nothing` if the function expression returns `Nothing`.
```

glyn (commented):
This looks good now, thanks. I've suggested some editorial improvements, but I don't mind if these are deferred.

I have now set my review type to "comment" rather than "request changes". (I don't want to approve this PR over the alternatives as I'm still mulling over the best option.)

glyn (commented):
Looks like a fine option now, thanks.

draft-ietf-jsonpath-base.md
gregsdennis: I was actually listing them in the order they're in the table.

glyn:
Argh! I dismissed my review but this ended up approving the PR rather than changing it to a comment. (The [github docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) claim is would change to a comment.) @cabo & @goessner: please note.

draft-ietf-jsonpath-base.md
glyn: I'm afraid that subtlety was lost on me.

draft-ietf-jsonpath-base.md
gregsdennis: Maybe this tongue-in-cheek kebaberizing will help:

> When any `singular-path` or function-expression-with-declared-type-of-`SingleNodeType` appears as an argument for a parameter of type `ValueType`, each such path or function expression is replaced by the value of its node or `Nothing` if it is empty.

I agree that it's a poor result of English grammar.  Maybe if I reverse them:

> When any function expression with declared type of `SingleNodeType` or `singular-path` appears as an argument for a parameter of type `ValueType`, each such path or function expression is replaced by the value of its node or `Nothing` if it is empty.

It still has the grammatical issue, but since a function can't result in a `singular-path`, maybe it makes more sense.

gregsdennis:
I'd also like to revisit the text quoted in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/418#discussion_r1127034954 to add the empty nodelist <-> `Nothing` case that is notably missing.  A separate PR, though, and maybe only if this goes through.

draft-ietf-jsonpath-base.md
gregsdennis: I'm not sure it really matters.

draft-ietf-jsonpath-base.md
glyn: Let's not waste more time on editorial niceties then.

draft-ietf-jsonpath-base.md
glyn: Let's not waste more time on editorial niceties then.

glyn (approved):
This is a good addition to PR #412, thanks.

PR #420, opened 2023-03-08
Avoid implicitly converting NodesType to ValueType

glyn:
This implicit conversion can be surprising in some situations, so it is made explicit via the value() function.

Calling the value() function is a conscious choice, so should not be a surprise to the user.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1509,7 +1509,7 @@ The following implicit type conversion may occur:
     is `LogicalTrue`.
   * If the nodelist is empty, the conversion result is `LogicalFalse`.
 
-An expression with a declared type of `NodesType` can be converted to a `ValueType` using the explicit type conversion function `value()` (see {{value}}).
+Given an expression with a declared type of `NodesType`, a `ValueType` can be obtained using a function such as `value()` (see {{value}}).
 
 The well-typedness of function expressions can now be defined in terms of this type system.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1675,7 +1675,9 @@ instance of `ValueType`.
 | `$[?count(1) == 1]` | not well-typed since `1` is not a path  |
 | `$[?count(foo(@.*)) == 1]` | well-typed, where `foo` is a function extension with a parameter of type `NodesType` and result type `NodesType` |
 | `$[?match(@.timezone, 'Europe/.*')]`         | well-typed |
-| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as JSONPath logicals may not be used in comparisons |
+| `$[?match(@.timezone, 'Europe/.*') == true]` | not well-typed as `LogicalType` may not be used in comparisons |
+| `$[?value(@..color) == "red"]` | well-typed |
+| `$[?value(@..color)]` | not well-typed as `ValueType` may not be used in a test expression |
 {: title="Function expression examples"}
 
 ## Segments  {#segments-details}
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1500,22 +1500,18 @@ The abstract instances above can be obtained from the concrete representations i
...
-    the value of the node.
-  * If the nodelist is empty or contains multiple nodes, the
-    conversion result is `Nothing`.
 * Where a member of `NodesType` needs to be converted to a
   `LogicalType`, the conversion proceeds as follows:
   * If the nodelist contains one or more nodes, the conversion result
     is `LogicalTrue`.
   * If the nodelist is empty, the conversion result is `LogicalFalse`.
 
+An expression with a declared type of `NodesType` can be converted to a `ValueType` using the explicit type conversion function `value()` (see {{value}}).
```
cabo, line 1513: I'm not even sure this is a "conversion" function (there may be several semantics for such a conversion function, and the merit of the present PR is that it makes that function explicit so others can be added later).

cabo (approved):

draft-ietf-jsonpath-base.md
cabo, line 1513:
```suggestion
Given an expression with a declared type of `NodesType`, a `ValueType` can be obtained using a function such as `value()` (see {{value}}).
```

glyn: Yes, let's avoid that terminology. Adopted your suggestion.

gregsdennis:
I specifically do not like conversion functions.  They are unnecessary cruft.

For the conversion from `NodesType` to `ValueType`, I'm happy forbidding it, as we have in #412.  I see no reason why a multiple nodelist should ever be converted to a value, even if it _happens_ to have zero or one node.

gregsdennis (changes requested):
I prefer #412

draft-ietf-jsonpath-base.md
```diff
@@ -1671,6 +1639,31 @@ that is the second argument; the result is `LogicalTrue` if such a
...
+~~~
+
+Its only argument is an instance of `NodesType` (possibly taken from a
+`filter-path` as in the example above).  The result is an
+instance of `ValueType`.
+
+* If the argument contains a single node, the result is
+  the value of the node.
+* If the argument is `Nothing` or contains multiple nodes, the
+  result is `Nothing`.
```
gregsdennis, line 1664: This is a gotcha.  It produces potentially unexpected behavior in expressions.

(Yes, I understand that its behavior is specified, but users won't read the spec.)

This is no different than the implicit conversion that exists in the current `main` document.  The problem is the surprise.  This is why I included the table of behaviors that has been deleted in this PR.

The behavior we define shouldn't contain surprises for users.  That just generates StackOverflow questions.  It should be preferred to reduce the number of questions by devising a system that is intuitive.  This isn't it.  (Arguably, what's in `main` also isn't it.)

glyn: The point of putting this behaviour inside a function is to make opting in to the behaviour a conscious choice.

draft-ietf-jsonpath-base.md
```diff
@@ -1671,6 +1639,31 @@ that is the second argument; the result is `LogicalTrue` if such a
...
+
+
 ### Examples
```
gregsdennis, line 1667: We should include examples.

glyn: Yes: added.

goessner (approved):
After clarification of some points everything is ok.

glyn:
Merged. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/404#issuecomment-1465175176 for the rationale behind this.

PR #421, opened 2023-03-09
Editorial extracts from PR 412

glyn:
These editorial changes from PR #412 are needed in the base.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1532,9 +1532,6 @@ A function expression is well-typed if all of the following are true:
 * If it occurs directly as a `comparable` in a comparison, the
   function is declared to have a result type of `ValueType`, or
   (conversion applies) `NodesType`.
-* Otherwise, it is occurring as an argument to a further function
-  expression, and the following rules for function arguments apply to
-  its declared result type.
 * Each argument of the function can be used for the declared type of the corresponding declared
   parameter according to one of the following rules:
    * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1272,14 +1272,18 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
 
 The comparison operators `==` and `<` are defined first and then these are used to define `!=`, `<=`, `>`, and `>=`.
 
-When a path resulting in an empty nodelist appears on either side of a comparison:
+When a path resulting in an empty nodelist or a function expression returning `Nothing` appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two paths each of which result in an empty nodelist.
+is between:
+    * two paths each of which result in an empty nodelist,
+    * two functions each of which returns `Nothing`, or
+    * a path which results in an empty nodelist and a function which returns `Nothing`.
 
 * a comparison using the operator `<` yields false.
 
-When any path on either side of a comparison results in a nodelist consisting of a single node, each such path is
+When any path or function expression on either side of a comparison results in a nodelist consisting of a single node, each such path
+or function expression is
 replaced by the value of its node and then:
 
 * a comparison using the operator `==` yields true if and only if the comparison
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1528,9 +1532,6 @@ A function expression is well-typed if all of the following are true:
 * If it occurs directly as a `comparable` in a comparison, the
   function is declared to have a result type of `ValueType`, or
   (conversion applies) `NodesType`.
-* Otherwise, it is occurring as an argument to a further function
```
cabo, line 1531: I still like this text, as it rounds out the cases in the outer list.

draft-ietf-jsonpath-base.md
```diff
@@ -1272,14 +1272,18 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
...
-is between two paths each of which result in an empty nodelist.
+is between:
+    * two paths each of which result in an empty nodelist,
+    * two functions each of which returns `Nothing`, or
+    * a path which results in an empty nodelist and a function which returns `Nothing`.
 
 * a comparison using the operator `<` yields false.
 
-When any path on either side of a comparison results in a nodelist consisting of a single node, each such path is
+When any path or function expression on either side of a comparison results in a nodelist consisting of a single node, each such path
```
cabo, line 1285: "this side"?

draft-ietf-jsonpath-base.md
```diff
@@ -1272,14 +1272,18 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
 
 The comparison operators `==` and `<` are defined first and then these are used to define `!=`, `<=`, `>`, and `>=`.
 
-When a path resulting in an empty nodelist appears on either side of a comparison:
+When a path resulting in an empty nodelist or a function expression returning `Nothing` appears on either side of a comparison:
 
 * a comparison using the operator `==` yields true if and only if the comparison
-is between two paths each of which result in an empty nodelist.
+is between:
```
cabo, line 1278: maybe just say "the other side also results in an empty nodelist or Nothing"?

cabo (commented):
These changes are probably needed, but a bit on the tedious side.  Please see detailed comments.

draft-ietf-jsonpath-base.md
```diff
@@ -1272,14 +1272,18 @@ rather than the negated existence test `!@.foo` (which yields false if `@.foo` s
...
+    * two paths each of which result in an empty nodelist,
+    * two functions each of which returns `Nothing`, or
+    * a path which results in an empty nodelist and a function which returns `Nothing`.
```
gregsdennis, line 1281: If we're listing out cases, we need to add "empty nodelist function == nothing function"

draft-ietf-jsonpath-base.md
gregsdennis: I had removed it because the changes I had made below it made it moot.  Without those other changes, I think this is still necessary.

draft-ietf-jsonpath-base.md
glyn: Yes. I think I've found a more compact way of writing all this.

draft-ietf-jsonpath-base.md
glyn: Very good. I'm including that in a rewording.

draft-ietf-jsonpath-base.md
glyn: Yes, but I think "that side" is better.

draft-ietf-jsonpath-base.md
glyn: Ok, I'll put the bullet back with a minor rewording.

cabo (approved):
Looks good now!
Should merge this quickly so we can rebase the others.

draft-ietf-jsonpath-base.md
glyn: Screenshot 2023-03-09 at 11 23 26

gregsdennis: Yes, but the thread view still shows the commit as it was when the comment was made.

gregsdennis:
Why was this merged so quickly?!

glyn:
> Why was this merged so quickly?!

We merged this editorial change quickly so we could rebase the other substantive PRs on it and put us in a better position to compare their pro's and con's.

PR #424, opened 2023-03-12
Close #411: Use "path" only for paths, and "query" for queries

### LINKED ISSUE #411: -path should mostly be -query
The grammar (and some of the text) uses `-path` as a qualifier for embedded queries.
This is confusing, as a query Is not a path.
(`normalized-path` also is a query, but is *used* as a path and therefore is not addressed by this issue.)

cabo:


glyn (changes requested):
A couple of occurrences of "path" remain in section A.1, at least one of which needs to be changed to "query":

Screenshot 2023-03-12 at 17 42 30

I would have made the change, but I'm no expert on XPath terminology.

cabo:
XPath terminology is a bit weird, because XPath has a somewhat competing standard XQuery...
XPath is based on DSSSL, which had a complex expression language or building paths in the trees that was based on the Scheme language.
(Which is the experience that makes me hesitant just jumping headlong into another expression language...)

glyn (approved):
LGTM with that change. Thanks!

PR #426, opened 2023-03-14
Motivate the term "JSONPath"

glyn:
PR #424 changed the spec to use "path" only for paths, and "query" for queries. The term "JSONPath" is then left in need of some motivation. This is provided in the History section so as not to interfere with the normative text.


gregsdennis (approved):

cabo (approved):

goessner (approved):
... fine ...

Issue #427, opened 2023-03-14
The terms "instance" and "abstract"


timbray:
I am working my way through 2.6.1 and my progress is being hampered by the use of the undefined terms "instance" and "abstract".   
 
I'm trying to understand this: "A type is a set of instances."  The difficulty is increased when I run into undefined notations such as `Value(v)`.

I *think* you're trying to say something like "A type is construct that constrains what can appear as an argument to a function and the result of evaluating a function."

I'm not sure Table 13 is helpful at all, the bullet list seems to contain all the useful information without relying on mysterious notations.   And I'm not sure you need to use the words "abstract" or "abstraction" at all.  For example, is the following clearer?

"ValueType" means either a JSON value or Nothing.

"LogicalType" means the result of a logical-expr. Its possible values, LogicalTrue and LogicalFalse, are not related to the JSON literals true and false and have no direct syntactical representation in JSONPath.

"NodesType" means the results of evaluating a filter-query (which appears in a test expression or as a function argument). Members of NodesType have no direct syntactical representation in JSONPath.[¶](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-11.html#section-2.6.1-6.3)

glyn:
As you point out, this is tricky language for readers unfamiliar with abstract data types. It would be good to find wording which presumes less background knowledge.

I have three concerns over the rewording you propose: the loss of abstraction, whether the resultant types are well defined, and the impact on the spec.

## Loss of abstraction

The types are intended to be _abstract_: each JSONPath implementation is free to choose its own concrete representation of each type.

Let's take `ValueType` as an example. As you observe, it's meant to cover two possibilities. Either it corresponds to a JSON value or to the abstract value Nothing. The value `Nothing` doesn't correspond to any JSON value.

So an implementation in C could represent a `ValueType` as a pointer to a JSON value (possibly a struct) with the null pointer representing `Nothing`. On the other hand, an implementation in Haskell could use a data declaration:
```
data ValueType = Value JSONValue | Nothing
```

Note that in both these implementations, `ValueType` is distinct from JSON values: a JSON value is not valid as a `ValueType`.

If we removed the "abstract" language, we'd need to ensure that these kinds of implementations, and others, are still acceptable.

## Well definedness of types

[...]

PR #428, opened 2023-03-14
lightweight editorial suggestions and lose "canonical"

timbray:
The editorial suggestions are trying to be a bit more idiomatic and improve flow; my feelings will not be hurt by any rejections.

I'm not sure the spec quite covers index selectors of the form [5:] so I tried adding that. If I got it wrong, that may be a symptom of an editorial problem.

I think it's bad form to use the word "canonical" without saying what you mean, so I removed it from section 2.9.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -396,7 +396,7 @@ of bracket notation. Examples and descriptions use shorthands where convenient.
 
 ### Selectors
 
-A name selector, e.g. `'name'`, selects a named child in an object.
+A name selector, e.g. `'name'`, selects a named child of an object.
 
 An index selector, e.g. `3`, selects an indexed child in an array.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -259,7 +259,7 @@ Parameter:
 : Formal parameter that can take Arguments (actual parameters) in a function-expression.
 
 Normalized Path:
-: A simple form of JSONPath expression that identifies a node in a value by
+: A minimal form of JSONPath expression that identifies a node in a value by
   providing a query that results in exactly that node.  Similar
   to, but syntactically different from, a JSON Pointer {{-pointer}}.
 
@@ -300,7 +300,7 @@ member names are not.
 
 This section is informative.
 
-This document picks up {{{Stefan Gössner}}}'s popular JSONPath proposal
+This document is based on {{{Stefan Gössner}}}'s popular JSONPath proposal
 dated 2007-02-21 {{JSONPath-orig}}, builds on the experience from the widespread
 deployment of its implementations, and provides a normative specification for it.
 
@@ -371,7 +371,7 @@ The root node identifier `$` refers to the root node of the argument,
 i.e., to the argument as a whole.
 
 The current node identifier `@` refers to the current node in the context
-of the evaluation of a filter expression (described later).
+of the evaluation of a filter expression (described below).
 
 ### Segments
 
@@ -396,9 +396,9 @@ of bracket notation. Examples and descriptions use shorthands where convenient.
 
 ### Selectors
 
-A name selector, e.g. `'name'`, selects a named child of an object.
+A name selector, e.g. `'name'`, selects a named child in an object.
 
-An index selector, e.g. `3`, selects an indexed child of an array.
+An index selector, e.g. `3`, selects an indexed child in an array.
 
 A wildcard `*` ({{wildcard-selector}}) in the expression `[*]` selects all children of a
 node and in the expression `..[*]` selects all descendants of a node.
@@ -525,7 +525,7 @@ as described in {{fnex}}.
 A JSONPath implementation MUST raise an error for any query which is not
 well-formed and valid.
 The well-formedness and the validity of JSONPath queries are independent of
-the JSON value the query is applied to; no further errors relating to the
+the JSON value the query is applied to. No further errors relating to the
 well-formedness and the validity of a JSONPath query can be
 raised during application of the query to a value.
 
@@ -568,9 +568,7 @@ the *argument*, and produces a nodelist (i.e., a list of zero or more nodes of t
 The query is a root identifier followed by a sequence of zero or more segments, each of
 which is applied to the result of the previous root identifier or segment and provides
 input to the next segment.
-These results and inputs take the form of a nodelist.
-
-Segments can be added to a query to drill further into the structure of the input value.
+These results and inputs take the form of nodelists.
 
 The nodelist resulting from the root identifier contains a single node,
 the argument.
@@ -933,6 +931,7 @@ S                   = *B        ; optional blank space
 ~~~~
 
 The slice selector consists of three optional decimal integers separated by colons.
+The second colon is omitted when the third integer is.
 
 To be valid, the integers provided MUST be in the I-JSON
 range of exact values, see {{synsem-overview}}.
@@ -1074,6 +1073,7 @@ The following examples show the array slice selector in use by a child segment.
 | Query | Result | Result Paths | Comment |
 | :---: | ------ | :----------: | ------- |
 | `$[1:3]` | `"b"` <br> `"c"` | `$[1]` <br> `$[2]` | Slice with default step |
+| `$[5:]` | `"f"` <br> `"g"` | `$[5]` <br> `$[6]` | Slice with no end index |
 | `$[1:5:2]` | `"b"` <br> `"d"` | `$[1]` <br> `$[3]` | Slice with step 2 |
 | `$[5:1:-2]` | `"f"` <br> `"d"` | `$[5]` <br> `$[3]` | Slice with negative step |
 | `$[::-1]` | `"g"` <br> `"f"` <br> `"e"` <br> `"d"` <br> `"c"` <br> `"b"` <br> `"a"` | `$[6]` <br> `$[5]` <br> `$[4]` <br> `$[3]` <br> `$[2]` <br> `$[1]` <br> `$[0]` | Slice in reverse order |
@@ -1191,7 +1191,7 @@ current-node-identifier = "@"
 Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Queries, each of
-which selects at most one node the value of which is then used; and
+which selects at most one node the value of which is then used; or
 function expressions (see {{fnex}}) of type `ValueType` or
 `NodesType` (see {{type-conv}}).
 
@@ -1253,7 +1253,7 @@ Applied to primitive values, it selects nothing.
 
 The order in which the children of an object appear in the resultant nodelist is not stipulated,
 since JSON objects are unordered.
-Children of an array appear in array order in the resultant nodelist.
+Children of an array are ordered by their array index in the resultant nodelist.
 
 ##### Existence Tests {#extest}
 {: unnumbered}
@@ -1912,7 +1912,7 @@ Queries:
 
 ## Normalized Paths
 
-A Normalized Path is a canonical representation of the location of a node in a value and
+A Normalized Path is a minimal representation of the location of a node in a value which
 uniquely identifies the node in the value.
 Specifically, a Normalized Path is a JSONPath query with restricted syntax (defined below),
 e.g., `$['book'][3]`, which when applied to the value results in a nodelist consisting
@@ -1920,7 +1920,7 @@ of just the node identified by the Normalized Path.
 Note that a Normalized Path represents the identity of a node _in a specific value_.
 There is precisely one Normalized Path identifying any particular node in a value.
 
-A canonical representation of a nodelist is as a JSON arrays of strings, where the strings are
+A nodelist may be represented compactly in JSON as an array of strings, where the strings are
 Normalized Paths.
 
 Normalized Paths provide a predictable format that simplifies testing and post-processing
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -396,9 +396,9 @@ of bracket notation. Examples and descriptions use shorthands where convenient.
 
 ### Selectors
 
-A name selector, e.g. `'name'`, selects a named child of an object.
+A name selector, e.g. `'name'`, selects a named child in an object.
```
glyn, line 399: We can talk about a named value in an object, but when it comes to children, I would say that an object's children are children _of_ the object, not children _in_ the object. Similarly for arrays.

draft-ietf-jsonpath-base.md
```diff
@@ -933,6 +931,7 @@ S                   = *B        ; optional blank space
 ~~~~
 
 The slice selector consists of three optional decimal integers separated by colons.
+The second colon is omitted when the third integer is.
```
glyn, line 934: Not so. `::` is a valid slice for example.

draft-ietf-jsonpath-base.md
```diff
@@ -1929,10 +1929,12 @@ Normalized Paths are used in this document as result paths in examples.
...
 Normalized Paths use the canonical bracket notation, rather than dot notation.
 
-Single quotes are used to delimit string member names. This reduces the number of characters that
-need escaping when Normalized Paths appear in double quote delimited strings, e.g., in JSON texts.
+Single quotes are used in Normalized Paths to delimit string member names. This reduces the
+number of characters that need escaping when Normalized Paths appear in double quote-delimited
+strings, e.g., in JSON texts.
 
-Certain characters are escaped, in one and only one way; all other characters are unescaped.
+Certain characters are escaped in Noramlized Paths, in one and only one way; all other
```
glyn, line 1936:
```suggestion
Certain characters are escaped in Normalized Paths, in one and only one way; all other
```

glyn (changes requested):
All looks good except where I've commented.

draft-ietf-jsonpath-base.md
```diff
@@ -259,7 +259,7 @@ Parameter:
 : Formal parameter that can take Arguments (actual parameters) in a function-expression.
 
 Normalized Path:
-: A simple form of JSONPath expression that identifies a node in a value by
+: A minimal form of JSONPath expression that identifies a node in a value by
```
gregsdennis, line 262: I'm not sure this is right either.  If anything, the "normalized" syntax is more verbose for many cases (e.g. `$.a.b` vs `$['a']['b']`.

Maybe "formal"?

draft-ietf-jsonpath-base.md
```diff
@@ -371,7 +371,7 @@ The root node identifier `$` refers to the root node of the argument,
 i.e., to the argument as a whole.
 
 The current node identifier `@` refers to the current node in the context
-of the evaluation of a filter expression (described later).
+of the evaluation of a filter expression (described below).
```
gregsdennis, line 374: or just link to the section?

draft-ietf-jsonpath-base.md
```diff
@@ -1191,7 +1191,7 @@ current-node-identifier = "@"
 Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Queries, each of
-which selects at most one node the value of which is then used; and
+which selects at most one node the value of which is then used; or
 function expressions (see {{fnex}}) of type `ValueType` or
 `NodesType` (see {{type-conv}}).
```
gregsdennis, line 1196: I think with the addition of `value()`, `NodesType` is no longer valid in a comparison.  Is this an oversight?

draft-ietf-jsonpath-base.md
```diff
@@ -1253,7 +1253,7 @@ Applied to primitive values, it selects nothing.
 
 The order in which the children of an object appear in the resultant nodelist is not stipulated,
 since JSON objects are unordered.
-Children of an array appear in array order in the resultant nodelist.
+Children of an array are ordered by their array index in the resultant nodelist.
```
gregsdennis, line 1256: We may want to avoid "index" outside of the index selector.

draft-ietf-jsonpath-base.md
```diff
@@ -1912,15 +1912,15 @@ Queries:
 
 ## Normalized Paths
 
-A Normalized Path is a canonical representation of the location of a node in a value and
+A Normalized Path is a minimal representation of the location of a node in a value which
```
gregsdennis, line 1915: "formal" here too, maybe?

draft-ietf-jsonpath-base.md
glyn: Good catch. I think this oversight is repeated in 2.6.3: "If it occurs directly as a comparable in a comparison, the function is declared to have a result type of ValueType, or (conversion applies) NodesType." doesn't need to talk about `NodesType`.

draft-ietf-jsonpath-base.md
glyn: Not sure "formal" is a great deal better. Perhaps just skip the adjective altogether?
```suggestion
: A form of JSONPath expression that identifies a node in a value by
```

draft-ietf-jsonpath-base.md
glyn: Again:
```suggestion
A Normalized Path is a representation of the location of a node in a value which
```

draft-ietf-jsonpath-base.md
gregsdennis: Addressed in #429.

draft-ietf-jsonpath-base.md
```diff
@@ -1912,15 +1912,15 @@ Queries:
...
 uniquely identifies the node in the value.
 Specifically, a Normalized Path is a JSONPath query with restricted syntax (defined below),
 e.g., `$['book'][3]`, which when applied to the value results in a nodelist consisting
 of just the node identified by the Normalized Path.
 Note that a Normalized Path represents the identity of a node _in a specific value_.
 There is precisely one Normalized Path identifying any particular node in a value.
 
-A canonical representation of a nodelist is as a JSON arrays of strings, where the strings are
+A nodelist may be represented compactly in JSON as an array of strings, where the strings are
```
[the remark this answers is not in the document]
gregsdennis: Actually, looking at this again, I think the adjective is fine.  It's describing a nodelist, which is more compact if you just have the Normalized Paths without the values.

Rescinding my comment here.

draft-ietf-jsonpath-base.md
cabo: > Not sure "formal" is a great deal better. Perhaps just skip the adjective altogether?

Uniquely?

draft-ietf-jsonpath-base.md
cabo, line 374:
```suggestion
of the evaluation of a filter expression ({{filter-selector}}).
```

draft-ietf-jsonpath-base.md
cabo: Unborn child?  (Just kidding.)
+1

draft-ietf-jsonpath-base.md
cabo, line 934:
```suggestion
The second colon can be omitted when the third integer is.
```

draft-ietf-jsonpath-base.md
cabo, line 1256:
```suggestion
In the resultant nodelist, children of an array are ordered by their position in the array.
```

draft-ietf-jsonpath-base.md
cabo: Yep.  One more, later.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: I think uniqueness of the identified node is implicit. Or did you mean uniqueness of the query??

draft-ietf-jsonpath-base.md
glyn: Ok:
```suggestion
A name selector, e.g. `'name'`, selects a named child of an object.
```

draft-ietf-jsonpath-base.md
```diff
@@ -396,9 +396,9 @@ of bracket notation. Examples and descriptions use shorthands where convenient.
 
 ### Selectors
 
-A name selector, e.g. `'name'`, selects a named child of an object.
+A name selector, e.g. `'name'`, selects a named child in an object.
 
-An index selector, e.g. `3`, selects an indexed child of an array.
+An index selector, e.g. `3`, selects an indexed child in an array.
```
glyn, line 401: Similarly:
```suggestion
An index selector, e.g. `3`, selects an indexed child of an array.
```

glyn (changes requested):
The only issues with this PR that remain are covered by my suggestions, which I believe to be necessary before merging. @cabo PTAL.

draft-ietf-jsonpath-base.md
cabo: For each node, there is exactly one ("unique") normalized path.
That is the main point of them to me.

draft-ietf-jsonpath-base.md
cabo, line 1915:
```suggestion
A Normalized Path is a unique representation of the location of a node in a value which
```

This is definitely a place where "unique" fits.  (Not so great in the other place further up.)

draft-ietf-jsonpath-base.md
```diff
@@ -259,7 +259,7 @@ Parameter:
 : Formal parameter that can take Arguments (actual parameters) in a function-expression.
 
 Normalized Path:
-: A simple form of JSONPath expression that identifies a node in a value by
+: A minimal form of JSONPath expression that identifies a node in a value by
   providing a query that results in exactly that node.  Similar
```
cabo, line 263: So let's do the "unique" here:

```suggestion
  providing a unique query that results in exactly that node.  Similar
```

cabo (approved):

draft-ietf-jsonpath-base.md
gregsdennis: Here, "unique" doesn't work.  Identifying `foo` in `{"foo": "bar"}` can be achieved using `$['foo']` and `$.foo`.  The key here is that no shorthand syntaxes are used.

Secondly, are Normalized Paths necessarily Singular Paths?  If so, we should state that.

I think these two requirements (Singular Path and expressed without shorthand) are sufficient to define Normalized Path.

glyn: > Here, "unique" doesn't work. Identifying `foo` in `{"foo": "bar"}` can be achieved using `$['foo']` and `$.foo`. The key here is that no shorthand syntaxes are used.

The way I read the spec is that the terminology section gives a loose definition of Normalized Path which is tightened up in section 2.9. With that reading, it would be ok to add uniqueness to the term definition because that falls out of the tighter definition in section 2.9. Maybe we need to add a forward reference from the term definition to section 2.9 to make it clear this is the way to read the spec? (I wouldn't want to add the full definition of Normalized Path to the terminology section. )

> 
> Secondly, are Normalized Paths necessarily Singular Paths? If so, we should state that.

Yes, they are. Section 2.9 already says:

>> Note: Normalized Paths are Singular Queries, but not all Singular Queries are Normalized Paths.

> 
> I think these two requirements (Singular Path and expressed without shorthand) are sufficient to define Normalized Path.

Unfortunately, those conditions are not sufficient. As section 2.9 says:

>> For example, $[-3] is a Singular Query, but is not a Normalized Path.

`$[-3]` is a Singular Query and is expressed without shorthand. (NB. the spec doesn't describe such array indices as shorthand.)

(PS. Pardon the above markup. I used `>>` for quotes from the draft to distinguish them from quotes of the earlier comment.)

gregsdennis: All of that is fair, I suppose.

The thing with `$[-3]` is that it's not unique _for all JSON instances_.  Processing that path on JSON arrays of different lengths will select nodes at different positions, but processing `$[3]` on those arrays will always select the fourth node (if it exists).

Maybe

> providing a unique query that results in exactly that node for all JSON instances.

(or "data" or whatever instead of "instances" [ref](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/427))

glyn: Ok, how about `$[1:2]` (as a synonym for `$[1]`)? It's a Singular Query, without shorthand, and results in the same node for all JSON instances.

gregsdennis: Yeah, I don't know.  This is hard to nail down, and it's late for me. 🙂

cabo: There is nothing that is fundamentally better about always using an unsigned integer to represent normalized paths than using a negative integer.
We just decided always using the unsigned integer integer is more appealing (probably because we feel that this is a more fundamental way to access array elements, but that is not fundamentally so...).
So normalized paths are first and foremost a convention that lead to them being unique for each node.

cabo: We could be more explicit and add a sentence:
Each node in an argument is identified by exactly one normalize path (we say, the normalized path is "unique" for that node), and, to be a normalized path for a specific argument, the normalize path needs to identify exactly one node.

draft-ietf-jsonpath-base.md
cabo: Since "of" includes "in" (the keyboard *of* my laptop), I think we can accept this suggestion.

draft-ietf-jsonpath-base.md
glyn:
```suggestion
  providing a query that results in exactly that node.  Each node in an argument is identified by exactly one Normalized Path (we say, the Normalized Path is "unique" for that node), and, to be a Normalized Path for a specific argument, the Normalized Path needs to identify exactly one node. Similar
```

draft-ietf-jsonpath-base.md
glyn: Get rid of minimal now we have expanded the definition:
```suggestion
: A form of JSONPath expression that identifies a node in a value by
```

glyn (approved):
LGTM now, thanks.

PR #429, opened 2023-03-14
NodesType functions are not directly comparable

glyn:
Thanks to @gregsdennis for spotting this.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1192,8 +1192,7 @@ Comparison expressions are available for comparisons between primitive
 values (that is, numbers, strings, `true`, `false`, and `null`).
 These can be obtained via literal values; Singular Queries, each of
 which selects at most one node the value of which is then used; and
-function expressions (see {{fnex}}) of type `ValueType` or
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1520,8 +1519,7 @@ The well-typedness of function expressions can now be defined in terms of this t
 A function expression is well-typed if all of the following are true:
 
 * If it occurs directly in a test expression, the function is declared
-  to have a result type of `LogicalType`, or (conversion applies)
-  `NodesType`.
```
gregsdennis, line 1524: I think this change was meant for below.  `NodesType` is still valid for test expressions.  The line below is about comparisons.

gregsdennis (changes requested):

gregsdennis (approved):

glyn:
@cabo I presume you are ok with this change, so I'm going to merge it. Any concern, let me know and we can adjust the result.

cabo:
> @cabo I presume you are ok with this change, so I'm going to merge it. 

Yes. Just didn't have chance to check this in context, but if I remember this properly, this change is good (and necessary).

PR #441, opened 2023-03-18
Rewrite Function Well-typedness (Section 2.6.3) to be less clunky

### LINKED ISSUE #439: Function Well-typedness section feels clunky
The section in question (and subject to some change soon) is

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/7d016f14ecbde412dfdc2fd4476f0b20a27867bb/draft-ietf-jsonpath-base.md?plain=1#L1512-L1521

The part that feels clunky to me is the first line:

> A function expression is well-typed if all of the following are true:

While not _wrong_, when you get into the bullet points, it feels odd that we're saying "all of the following are true" when only one things ever applies due to the conditions each item defines.

We're saying we have three options, A, B, and C and then we define rules:

- If A, then ...
- If B, then ...
- Otherwise C, then ...

(I know I made a [big stink](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/418#discussion_r1127031473) about changing "rules" to "scenarios" in a previous PR, but each of these statements conforms to the definition I gave for a "rule.")

The thing that feels weird is that we're saying they _all_ must be true, which... is "correct" because the requirements are conditionally applied, but also it seems that (because of the conditions) only one ever actually applies, meaning that the other two are more "not applicable" than they are "true."

Maybe if we instead said:

> A function expression is well-typed under the following rules:

Previous to the current text (before the type system was introduced), we had:

> A function expression is syntactically valid if:

cabo:
Close #439

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1512,39 +1512,61 @@ The well-typedness of function expressions can now be defined in terms of this t
 
 ### Well-Typedness of Function Expressions
 
-A function expression is well-typed if all of the following are true:
+For a function expression to be well typed:
 
-* If the function expression occurs as a `test-expr` in a logical expression, the function is declared
-  to have a result type of `LogicalType`, or (by conversion as
-  per {{type-conv}})
-  `NodesType`.
-* If the function expression occurs directly as a `comparable` in a comparison, the
-  function is declared to have a result type of `ValueType`.
-* Otherwise (the function expression occurs as an argument in another function
-  expression), the following rules for function arguments apply to
-  its declared result type.
-* The arguments of the function expression are well-typed, as follows.
+1. its declared type must be well typed in the context it occurs, and
+2. its arguments must be well typed for the declared type of the corresponding parameters.
 
-Each argument of the function can be used for the declared type of the corresponding declared
-parameter according to one of the following rules:
+(1) As per the grammar, a function expression can occur in three different
+immediate contexts, which lead to the following conditions for well-typedness:
 
-   * The argument is a function expression with declared result type that is the same as the declared type of the parameter.
-   * The argument is a function expression with declared result type `NodesType` and the declared type of the parameter is
-     `LogicalType`. In this case the argument is converted to
-     `LogicalType` as per {{type-conv}}.
-   * The argument is a value expressed as a literal and the declared type of the parameter is `ValueType`.
-   * The argument is a singular query and the declared type of the parameter is `ValueType`.
-     In this case:
-      * If the query results in a nodelist consisting of a single  node, the argument is the value of the node.
-      * If the query results in an empty nodelist, the argument is `Nothing`.
-   * The argument is a query (including singular query) and the declared type of the parameter is `NodesType`.
-   * The argument is a `logical-expr` and the declared type of the parameter is `LogicalType`.
+{:vspace}
+As a `test-expr` in a logical expression:
+: The function is declared to have a result type of `LogicalType`, or
+  (giving rise to conversion as per {{type-conv}}) `NodesType`.
 
-Note that the last bullet item includes the case that the argument is
-a query (including singular query) and the declared type of the
-parameter is `LogicalType`. In this case the nodelist resulting
-from the query is interpreted as a logical-expr in the same way
-({{extest}}) it would be converted to `LogicalType` as per {{type-conv}}.
+As a `comparable` in a comparison:
+: The function is declared to have a result type of `ValueType`.
+
+As a `function-argument` in another function expression:
+: The function's declared result type fulfills the following rules for
+  the corresponding parameter of the enclosing function.
+
+(2) The arguments of the function expression are well-typed when
+each argument of the function can be used for the declared type of the
+corresponding parameter, each according to one of the following
+conditions depending on the way the argument is supplied:
+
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1512,39 +1512,61 @@ The well-typedness of function expressions can now be defined in terms of this t
...
-   * The argument is a `logical-expr` and the declared type of the parameter is `LogicalType`.
-
-Note that the last bullet item includes the case that the argument is
-a query (including singular query) and the declared type of the
-parameter is `LogicalType`. In this case the nodelist resulting
-from the query is interpreted as a logical-expr in the same way
-({{extest}}) it would be converted to `LogicalType` as per {{type-conv}}.
+For a function expression to be well typed:
+
+1. its declared type must be well typed in the context it occurs, and
```
glyn, line 1517: Grammar:
```suggestion
1. its declared type must be well typed in the context in which it occurs, and
```

draft-ietf-jsonpath-base.md
```diff
@@ -1512,39 +1512,61 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+: the declared type of the parameter is `ValueType`.
+  In this case:
+
+  * If the query results in a nodelist consisting of a single node, the argument is the value of the node.
+  * If the query results in an empty nodelist, the argument is `Nothing`.
+
+As a query (including singular query):
+: the declared type of the parameter is `NodesType`.
+
+As a `logical-expr` (including a query):
```
glyn, line 1554: It is _possible_ to read this to mean "As a `logical-expr` that includes a query". Perhaps:
```suggestion
As a `logical-expr` (including `filter-query`):
```

draft-ietf-jsonpath-base.md
```diff
@@ -1512,39 +1512,61 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+
+1. its declared type must be well typed in the context it occurs, and
+2. its arguments must be well typed for the declared type of the corresponding parameters.
+
+(1) As per the grammar, a function expression can occur in three different
+immediate contexts, which lead to the following conditions for well-typedness:
+
+{:vspace}
+As a `test-expr` in a logical expression:
+: The function is declared to have a result type of `LogicalType`, or
```
glyn, line 1525: "declared to have a result type of" is a bit wordy. How about one of the following options?
```suggestion
: The function result type is `LogicalType`, or
```
```suggestion
: The function result is a `LogicalType`, or
```

If we adopt some shorter phrase, it needs to be applied consistently elsewhere, of course.

glyn (approved):
Very good. I suggested some small corrections/improvements for your consideration.

PS. Have you gone through all the combinations to check we have all the bases covered?

draft-ietf-jsonpath-base.md
cabo: Solved a bit differently in 8df54a4

draft-ietf-jsonpath-base.md
cabo: Solved more extensively in 1a8109c

cabo:
> PS. Have you gone through all the combinations to check we have all the bases covered?

I actually ordered the two definition lists in the order the production appear in the grammar, so this is now easy to check (and I have checked it).

draft-ietf-jsonpath-base.md
```diff
@@ -1512,39 +1512,61 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+  that a query in a logical context is interpreted as an existence test
+  ({{extest}}), effecting the conversion as per {{type-conv}}.
+
+As a function expression:
+: 1. the declared result type of the argument is the same as the declared
+     type of the parameter, or
+  2. the declared result type of the argument is `NodesType`, and the
+     declared type of the parameter is `LogicalType`.
+     In this case the argument is converted to `LogicalType` as per
+     {{type-conv}}.
```
gregsdennis, line 1569: Currently this is organized by argument type then specifying what parameter type allow that argument to be valid.  The language is a bit circular.

I think it may make more sense if it's organized by parameter type then specifying what argument types are permitted.

I had restructured this section that way in a previous PR, and I felt it flowed better.

gregsdennis (dismissed):

draft-ietf-jsonpath-base.md
cabo: Yes, I thought about which dimension to use as the outer and which as the inner one.
It turns out you want to use the easier to identify one (declared type) as the inner one, so that is what I chose.

draft-ietf-jsonpath-base.md
glyn: Yes, that works, thanks.

draft-ietf-jsonpath-base.md
glyn: I re-read the PR with @gregsdennis's comment in mind. I didn't notice any circularity and the flow seemed good to me. Probably boils down to a matter of taste.

cabo: Merging this now so we can get the other PRs going -- we might revisit this in another PR.

glyn: I was just about to do the same thing, with the same rationale.

PR #443, opened 2023-03-18
Don't capitalize Singular Query (except in Terminology Section)

### LINKED ISSUE #440: Inconsistent capitalisation
Sometimes the spec uses "Singular Query" and sometimes "singular query". We should be consistent. (If we decide to use lower case, it's probably best to address this issue after PR #438 has been merged, since that PR adds some usages of "Singular Query".)

Also, there is one instance of "Logical context", which I think would be better uncapitalised ("logical context").

cabo:
Close #440


glyn (approved):
After merging in my changes from main, it appears some Singular Queries still need fixing.

PR #444, opened 2023-03-18
Editorial consistency improvements

glyn:
Also, delete a term which is used only in the definition of the next term.

Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/440

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -273,12 +273,9 @@ Unicode Scalar Value:
   In other words, integers in either of the inclusive base 16 ranges 0 to D7FF and
   E000 to 10FFFF. JSON string values are sequences of Unicode scalar values.
 
-Singular Nodelist:
-: A nodelist containing at most one node.
-
 Singular Query:
 : A JSONPath expression built from segments each of which, regardless of the input value,
-  produces a Singular Nodelist.
+  produces a nodelist containing at most one node.
 
 Selector:
 : A single item within a segment that takes the input value and produces a nodelist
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -514,22 +511,22 @@ Unicode code points as described
 in {{RFC3629}}; implementation approaches that lead to an equivalent
 result are possible.
 
-A string to be used as a JSONPath query needs to be *well-formed* and
+A string to be used as a JSONPath query needs to be *well formed* and
 *valid*.
-A string is a well-formed JSONPath query if it conforms to the ABNF syntax in this document.
-A well-formed JSONPath query is valid if it also fulfills all semantic
+A string is a well formed JSONPath query if it conforms to the ABNF syntax in this document.
```
cabo, line 516: English grammar requires the hyphen here.

cabo (approved):
I'll approve that, but if you use a multi-word adjective in an attribute position, it must be hyphenated, so a few of the "well typed" are regressions.

draft-ietf-jsonpath-base.md
glyn: Oops! Fixed.

cabo (approved):
Thanks.

glyn:
> I'll approve that, but if you use a multi-word adjective in an attribute position, it must be hyphenated, so a few of the "well typed" are regressions.

I only found the two occurrences you suggested changing. If there are others, please let me know!

PR #445, opened 2023-03-18
442 unnumbered fixes

cabo:
This can be done in many different ways.
This PR values noise reduction over consistency.
Comments welcome.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -854,7 +854,6 @@ members (but not when it is applied to object nodes with fewer than two members 
 ### Index Selector {#index-selector}
 
 #### Syntax {#syntax-index}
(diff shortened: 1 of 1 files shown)


glyn (approved):
This all seems reasonable to me.

Note: now that the Filter Selector/Syntax section is visible in the TOC, it's a little
disturbing that it contains quite a lot of semantics. I tried separating these out once, but it became unreadable because the semantics section repeated a lot of syntax so the two could be associated properly. I think we have to live with this.

cabo:
That is indeed a problem in several sections, there are enough subtopics common to both that you would need to go back and forth.  Note how I dissolved the "logical operators" subsection from the semantics by adding the one sentence to the syntax subsection.  I also believe we have to live with this.

Issue #446, opened 2023-03-18
"Well-Typedness of Function Expressions" function parameter organization


gregsdennis:
The organization of the expansion of point 2 in "Well-Typedness of Function Expressions" can be improved.

Currently, it's grouped by the kind of argument, but it's much easier to consume if you group by function parameter type.  (Also, I think we're missing some cases.)

I'm going to put it in a table so it's a bit easier to see.

| Argument Kind | Parameter Types | Unlisted parameter types where this argument is valid |
|:-|:-|:-|
|literal|`ValueType`||
|singular query|`ValueType`|`LogicalType` (kinda included via `logical-expr` below)<br>`NodesType` (_maybe_ included in `logical-expr`?)|
|general query|`NodesType`|`LogicalType` (kinda included via `logical-expr` below)|
|`logical-expr` (feels odd to reference ABNF here and not in the others)|`LogicalType`||
|function|type match or<br>func:`NodesType` in param:`LogicalType`||

The above answers the question

> If I have an argument _X_, where can I use it?

I don't think that's going to be a very common question JSON Path authors will have.  I think the more common question will be

> What kinds of things can I use for arguments to this function?

To that end, we need to invert the grouping by using parameter type as the key.  Doing this, we get

|Parameter Type|Availble arguments|
|:-|:-|
|`ValueType`| literal, singular query, `ValueType` function |
|`LogicalType`| any `logical-expr` (including singular/general query, `LogicalType` function, `NodesType` function) |
|`NodesType`| singular/general query, `NodesType` function |

This is a lot easier to consume and mentally digest.  It also better serves as a reference for JSON Path authors.

(We lose the "function type matches parameter type" thing, but I don't think we need to explicitly say that.)

danielaparker:
Questions:

(1) A `nodelist` is defined in the draft as a list of location-value pairs. Is there a distinction between `nodelist` and `NodesType`?

(2) If a function has a parameter that accepts arguments of `NodesType`, does it receive a list of  location-value pairs?

gregsdennis:
> (1) A `nodelist` is defined in the draft as a list of location-value pairs. Is there a distinction between `nodelist` and `NodesType`?

Abstractly, yes. A nodelist is an instance (or member) of `NodesType`.

> (2) If a function has a parameter that accepts arguments of `NodesType`, does it receive a list of  location-value pairs?

Yes.

danielaparker:
[...]

PR #451, opened 2023-03-22
Transpose well-typedness argument/parameter conditions

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/446


draft-ietf-jsonpath-base.md
```diff
@@ -1515,38 +1515,29 @@ As a `function-argument` in another function expression:
...
   ({{extest}}), effecting the conversion as per {{type-conv}}.
 
-As a function expression:
-: 1. the declared result type of the argument is the same as the declared
-     type of the parameter, or
-  2. the declared result type of the argument is `NodesType`, and the
-     declared type of the parameter is `LogicalType`.
-     In this case the argument is converted to `LogicalType` as per
-     {{type-conv}}.
+* When the declared type of the parameter is `NodesType` and the argument is one of the following:
```
gregsdennis, line 1539: There's only one case here.  Can we remove "one of the following?"  If you want to keep it for symmetry, I'm fine with that.

draft-ietf-jsonpath-base.md
```diff
@@ -1515,38 +1515,29 @@ As a `function-argument` in another function expression:
...
-: the declared type of the parameter is `NodesType`.
-
-As a `logical-expr` (which includes `filter-query`):
-: the declared type of the parameter is `LogicalType`.
-  This includes the case that the argument is
-  a query (which includes singular query).
+conditions:
+
+* When the argument is a function expression with declared result type the same as the
+  declared type of the parameter.
```
gregsdennis, line 1521: Having this is fine since you don't repeat these cases below.  My table included these cases, so this statement would have been redundant.

gregsdennis (commented):
Thanks for taking a crack at this. I think it reads a lot better and now serves as a reference for users.

gregsdennis (approved):

draft-ietf-jsonpath-base.md
glyn: Yes, I will remove that, thanks. It was an oversight.

draft-ietf-jsonpath-base.md
glyn: I was in two minds whether to factor out that statement. There is a wrinkle either way, which was one of the reasons I made this a _draft_ PR. I'd appreciate @cabo's opinion (he'll be tied up with the IETF meeting for some time).

The parameter type major order has the advantage that the parameter types _nearly_ (!) partition the possibilities.

There is one overlap, AFAICT, when a `LogicalType` parameter takes a `LogicalType` function expression argument.  This case is covered both by the first and third conditions (i.e. major bullets). It's not very important because the conclusion is the same from either of those conditions. But it seems inaccurate to introduce the condition list with the words "according to _one of_ the following conditions" (emphasis added), since two of the conditions can apply at the same time in this case.

draft-ietf-jsonpath-base.md

glyn:
> Thanks for taking a crack at this.

You're welcome. The issue needed a bit of "show, don't tell".

> I think it reads a lot better and now serves as a reference for users.

I currently agree, but I'd like to let this PR go "cold" for a while and then compare how this section reads relative to the main branch. Meanwhile, others may like to give their opinions.

draft-ietf-jsonpath-base.md
```diff
@@ -1515,38 +1515,30 @@ As a `function-argument` in another function expression:
...
+  This includes the following cases:
+    * The argument is a query (which includes singular query).
+      In this case the nodelist resulting
+      from the query is interpreted as a `LogicalType` in the same way
+      that a query in a logical context is interpreted as an existence test
+      ({{extest}}), effecting the conversion as per {{type-conv}}.
+    * The argument is a function expression with declared result type `NodesType`.
+      In this case the argument is converted to LogicalType as per {{type-conv}}.
+
+* When the declared type of the parameter is `NodesType` and the argument is a general query
```
cabo, line 1540: Why do we say "general" here and not above?  Is there a difference?

glyn: No good reason. Fixed. (Also got rid of the extraneous ABNF reference just following this.)

draft-ietf-jsonpath-base.md
```diff
@@ -1515,38 +1515,30 @@ As a `function-argument` in another function expression:
...
+  declared type of the parameter.
+
+* When the declared type of the parameter is `ValueType` and the argument is one of the following:
+    * A value expressed as a literal.
+    * A singular query. In this case:
+        * If the query results in a nodelist consisting of a single node, the
+          argument is the value of the node.
+        * If the query results in an empty nodelist, the argument is Nothing.
+
+* When the declared type of the parameter is `LogicalType` and the argument is a `logical-expr`.
```
cabo, line 1530: This is the last remnant of trying to use an ABNF name to explain a condition.  Since this syntactically includes all function expressions, the "This includes" below is not strong enough (unless you are a lawyer and believe "expressio unius est exclusio alterius").  The first bullet point below does not exclude anything and just supplies additional information, which is fine.  The second bullet point essentially is trying to exclude Valuetype (as it addresses NodesType and the LogicalType case is already covered by the above first bullet point); it probably should say so.

glyn: I believe I've addressed this by the simplification in 0083810df7b04b401afd571633cf98f6dcba0ed1. (See the commit log for my reasoning.)

draft-ietf-jsonpath-base.md
glyn: The overlap was avoided by the simplification in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/commit/0083810df7b04b401afd571633cf98f6dcba0ed1.

cabo (approved):
I reordered (2) of 2.4.3 so related items are closed to each other (and the declared types also happen to be in alphabetical order).

cabo:
I now ran the tests (and updated a few details with changing external tests)

Issue #449, opened 2023-03-21
Types for Implementators vs. Consumers of JSON Path


hiltontj:
The latest version of the spec simplifies the type system for functions down to `NodesType`, `ValueType`, and `LogicalType`. I like it in this state, as it is digestible and should provide a good foundation to implement function extensions. I see these three types as natural given that each have analogs in the rest of the spec:

* `NodesType` is analogous to a query result
* `ValueType` is analogous to a comparable
* `LogicalType` is analogous to that which is used in existence tests

However, while implementing it, I have found this not complete enough. In order to satisfy the implicit conversions defined in the spec, I found that it helped to break the `ValueType` into three separate types, representing:

1. A JSON Value _(from a node)_
2. A JSON Value _(from a literal)_
3. Nothing

It is important in the implementation to distinguish between (1.) and (2.). Although a node, e.g., from a singular query like `@.a`, results in either a JSON Value, or Nothing, it still requires special treatment when handling implicit conversion. In other words, it makes sense to implicitly convert (1.) to `NodesType`, but not (2.).

I don't think that this necessitates a `SingleNode` type at the top level in the type system, because I think that it is more of an implementation detail that would confuse _consumers_ of the spec. However, I think that this deserves mention for _implementors_ of the spec.

cabo:
I don't think you can convert a ValueType to a NodesType, for exactly the reason you give -- a value need to be anchored in the query argument (beyond literals, e.g., length()).

gregsdennis:
> I think my confusion arises from the fact that, e.g., `@.a` (a singular query) needs to be interpreted as a `ValueType` or a `NodesType`, depending upon the context within which it is used.

Singular queries _always_ produce nodelists; specifically, they produce singular nodelists.  For comparisons, this case is explicitly called out ([current working doc](https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/blob/268ac37030739271af5b431f527c77ac62e8c8ea/draft-ietf-jsonpath-base.md?plain=1#L1269))

> When any query or function expression on either side of a comparison results in a nodelist consisting of a single node, that side is replaced by the value of its node and then...

There's not really a "conversion" as such; it just says that you extract the value.  But you _always_ get a nodelist from the query.

(Note that we list functions in that text, but `NodesType` functions are specifically disallowed in comparisons.  Probably an oversight that needs to be addressed.)

[...]

PR #452, opened 2023-03-26
Prototype the addition of SingleNodeType

glyn:
The purpose of this PR is to demonstrate the downside of introducing a `SingleNodeType` in terms of its impact on the spec and the additional complexity.

Ref https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/449


glyn:
#449 was closed.

PR #454, opened 2023-04-01
Improve note on comparisons using the operator `<`

glyn:


draft-ietf-jsonpath-base.md
```diff
@@ -1300,8 +1300,9 @@ the comparison is between values which are both numbers or both strings and whic
       lower Unicode scalar value than the second string or if both strings start with the same Unicode scalar value and
       the remainder of the first string compares less than the remainder of the second string.
 
-Note that comparisons using the operator `<` yield false if either value being
-compared is an object, array, boolean, or `null`.
+Note that comparisons using the operator `<` yield false if either side of the
+comparison results in an empty nodelist or `Nothing` or if either side is a value that
+is an object, array, boolean, or `null`.
```
timbray, line 1305: Has the term "Nothing" (especially since it's presented in code font) been introduced at this point?  If not, I think you need at least "(see below)".

glyn: No, good catch! It was actually used near the start of 2.3.5.2.2, so I added a cross-reference at that point (in 2874edff6356d009917b426d45f73beb033f4613).

draft-ietf-jsonpath-base.md
```diff
@@ -1300,8 +1300,9 @@ the comparison is between values which are both numbers or both strings and whic
       lower Unicode scalar value than the second string or if both strings start with the same Unicode scalar value and
       the remainder of the first string compares less than the remainder of the second string.
 
-Note that comparisons using the operator `<` yield false if either value being
-compared is an object, array, boolean, or `null`.
+Note that comparisons using the operator `<` yield false if either side of the
```
cabo, line 1303: Independent of your change:  I don't think this is a note.

```suggestion
Comparisons using the operator `<` yield false if either side of the
```

glyn: @cabo The reason I thought it should be a note is that the behaviour of `<` is already fully specified and this is just a summary. With that in mind, is it still better to remove the note?

cabo: That's what I get for looking at a change in isolation.  You are right, and with the additional complexity necessary, it is indeed questionable whether this note is needed.  I have a mild preference for it to stay, but wouldn't mind at all if it were deleted.

draft-ietf-jsonpath-base.md
cabo, line 1305: If this stays, I'd:

```suggestion
is an object, array, `true`, `false`, or `null`.
```

Improves symmetry (container or special value).

draft-ietf-jsonpath-base.md
glyn: I deleted it. We don't have a note summarising the semantics of `==` and I don't think we need to summarise the semantics of `<` either.

glyn:
@timbray Deleting the text should have addressed your original editorial issue. This PR also adds the missing cross-reference. If you're happy, please approve.

cabo (approved):

PR #456, opened 2023-04-01
Delete surprising behaviour sentence

glyn:
Based on a comment from @timbray and the attempt to delete this sentence in https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/454.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1483,12 +1483,7 @@ parameter of declared type `LogicalType`, with the equivalent conversion rule:
   * If the nodelist contains one or more nodes, the conversion result is `LogicalTrue`.
   * If the nodelist is empty, the conversion result is `LogicalFalse`.
 
-Notes:
-
-* Extraction of a value from a nodelist can be performed in several
-ways, so an implicit conversion from `NodesType` to `ValueType`
-may be surprising and has therefore not been defined.
-* A function expression with a declared type of `NodesType` can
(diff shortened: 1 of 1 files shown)


timbray (changes requested):
I don't think the text adds any value and, in most readers, is going to lead to puzzlement and fear that they're missing something important.  

The fact that you're going to all this work to motivate a feature that isn't actually used makes me wonder why it's even in the spec.

glyn:
> I don't think the text adds any value and, in most readers, is going to lead to puzzlement and fear that they're missing something important.

You're probably right.

> 
> The fact that you're going to all this work to motivate a feature that isn't actually used makes me wonder why it's even in the spec.

I've deleted it. I doubt we'll miss it once we've gotten over the deletion.

cabo:
Well, the sentence points out that `value()` is not special in any way (we have discussed variants of this function).  It is just the one that fits naturally into what we have been doing elsewhere.  The sentence is true. It also provides information that may not be obvious.  So how does it not add value?  It is true that it is a note and has no normative effect, so deleting it only hurts those people who now independently have to discover this.

glyn:
@timbray WDYT to @cabo's rationale for keeping the sentence? I can see both advantages either way.

glyn:
> @timbray WDYT to @cabo's rationale for keeping the sentence? I can see advantages either way.

@timbray Closing. Please re-open if you feel sufficiently strongly.

gregsdennis:
I think stating the rationale behind not having that conversion is reasonable.

PR #457, opened 2023-04-01
Improve function extension language

glyn:
Use parentheses as an indication that something is a function.

Also, don't confuse the reader by alluding to potential future alternatives to value().

Reviewers: is this an improvement? It's a draft PR for now until we decide the change is worthwhile.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1485,8 +1485,8 @@ ways, so an implicit conversion from `NodesType` to `ValueType`
 may be surprising and has therefore not been defined.
 A function expression with a declared type of `NodesType` can
 indirectly be used as an argument for a parameter of declared type
-`ValueType` by wrapping the expression in a call to a function
-extension such as "value" (see {{value}}).
+`ValueType` by wrapping the expression in a call to the
+`value()` function extension (see {{value}}).
 
 The well-typedness of function expressions can now be defined in terms of this type system.
 
@@ -1535,7 +1535,7 @@ conditions:
           argument is the value of the node.
         * If the query results in an empty nodelist, the argument is Nothing.
 
-### `length` Function Extension {#length}
+### `length()` Function Extension {#length}
 
 Parameters:
 : 1. `ValueType`
@@ -1543,7 +1543,7 @@ Parameters:
 Result:
 : `ValueType` (unsigned integer or `Nothing`)
 
-The "length" function extension provides a way to compute the length
+The `length()` function extension provides a way to compute the length
 of a value and make that available for further processing in the
 filter expression:
 
@@ -1564,7 +1564,7 @@ instance of `ValueType`: an unsigned integer or `Nothing`.
 * For any other argument value, the result is `Nothing`.
 
 
-### `count` Function Extension {#count}
+### `count()` Function Extension {#count}
 
 Parameters:
 : 1. `NodesType`
@@ -1572,7 +1572,7 @@ Parameters:
 Result:
 : `ValueType` (unsigned integer)
 
-The "count" function extension provides a way to obtain the number of
+The `count()` function extension provides a way to obtain the number of
 nodes in a nodelist and make that available for further processing in
 the filter expression:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1484,8 +1484,8 @@ ways, so an implicit conversion from `NodesType` to `ValueType`
 may be surprising and has therefore not been defined.
 A function expression with a declared type of `NodesType` can
 indirectly be used as an argument for a parameter of declared type
-`ValueType` by wrapping the expression in a call to a function
-extension such as "value" (see {{value}}).
+`ValueType` by wrapping the expression in a call to the
+`value()` function extension (see {{value}}).
```
glyn, line 1488: I don't think it's worth alluding to other potential extensions here. Let's just keep it simple and just talk about `value()`. This doesn't rule out other function extensions doing something similar in the future.

timbray: I am beginning to think that simply removing the "value()" function is the best path forward. YAGNI. It's a hypothetical solution to a not-yet-encountered problem.

glyn: I could live with deleting value() until such time as we actually need it. We can already pass a singular query as an argument for a parameter of type ValueType, which is the most common related usecase.

@cabo: would you be comfortable with us deleting "value()". (If not, what use it it?)

cabo: I don't understand this discussion.

value() is clearly needed to avoid a one-way dead-end in the type system.
If we only needed it for functions that return NodesType, we still would want to define it now, because leaving the definition to the first person who happens to define one is not so smart.

But there is a striking gap that value() fills even before there is such a function: `$[?value(@..color) == 'red']`.

With respect to the sentence under discussion: value() is not special here, so "such as" is exactly right.

glyn: > I don't understand this discussion.
> 
> value() is clearly needed to avoid a one-way dead-end in the type system. If we only needed it for functions that return NodesType, we still would want to define it now, because leaving the definition to the first person who happens to define one is not so smart.

Agreed.

> 
> But there is a striking gap that value() fills even before there is such a function: `$[?value(@..color) == 'red']`.

I don't find this usecase particularly compelling, but I know others do, so fair enough.

> 
> With respect to the sentence under discussion: value() is not special here, so "such as" is exactly right.

The wording might be right, but it needs expanding a little to be easier to understand. See fe4e662947aded320121cd2f74e95157281bd830.

cabo: Well, maybe this sentence *is* a note.

I didn't think that qualifying the potential alternatives to `value()` in that level of detail was necessary -- "such as" was good enough for me.

glyn: > Well, maybe this sentence _is_ a note.

I agree the sentence could be a note. The previous sentence ("Extraction of a value from a nodelist [...]") in the same paragraph is also a note. So I've introduced a note list.

> 
> I didn't think that qualifying the potential alternatives to `value()` in that level of detail was necessary -- "such as" was good enough for me.

I've reworded the sentence again to try to improve it.

cabo:
I think we'll need to explain the () notation, which should be familiar to C programmers but not necessarily to people who grew up on other languages.

I also think we have a  few more examples that point to hypothetical functions, such as `blt`, `bar`, `bnl`.

cabo (approved):
We could do the fun() explanation and completeness search in a separate PR.

glyn:
> I think we'll need to explain the () notation, which should be familiar to C programmers but not necessarily to people who grew up on other languages.

I don't think there's a great point at which to explain that notation and, based on your observation, I'm going off the idea.

> 
> I also think we have a few more examples that point to hypothetical functions, such as `blt`, `bar`, `bnl`.

glyn:
@cabo Let's see how #456 pans out before taking the current PR any further. Once I remove the function parentheses, the two PRs are changing the same text.

cabo:
> I don't think there's a great point at which to explain that notation

1.1 would be the place.

glyn:
@cabo:

> I think we'll need to explain the () notation, which should be familiar to C programmers but not necessarily to people who grew up on other languages.

As you suggested, I introduced this in 1.1, but near the start since it's notation rather than terminology.

> 
> I also think we have a few more examples that point to hypothetical functions, such as `blt`, `bar`, `bnl`.

I did a regexp search to find and fix the remaining examples (your list was complete). In doing so, I found the strange markup `wildcard` and removed the backticks since there was no good reason for them.

PR #464, opened 2023-04-15
≥, not >

cabo:
Turn wrong > 0x10000 into > 0xFFFF instead, as this makes more obvious why this is required (more than 4 hex digits).


glyn (approved):
Nice, thanks.

PR #467, opened 2023-04-20
Require function names to be registered

glyn:
Without this change, it is not very clear what happens when an unregistered function name is used.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1502,8 +1502,9 @@ The well-typedness of function expressions can now be defined in terms of this t
 
 For a function expression to be well typed:
 
-1. its declared type must be well typed in the context in which it occurs, and
-2. its arguments must be well typed for the declared type of the corresponding parameters.
+1. its function name must be registered (see {{iana-fnex}}),
+2. its declared type must be well typed in the context in which it occurs, and
+3. its arguments must be well typed for the declared type of the corresponding parameters.
 
 (1) As per the grammar, a function expression can occur in three different
 immediate contexts, which lead to the following conditions for well-typedness:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1510,7 +1510,8 @@ For a function expression to be well typed:
 parameter types of a function extension.
 
 Although a function expression being treated as well typed implies its function name is registered,
-the reverse is not necessarily true. Implementations must support the function extensions defined
+an implementation need not treat all registered functions as well typed.
+Implementations must support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
 If an implementation does not support a function extension used in a query, it should treat the
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1454,17 +1454,16 @@ Implementations MUST support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
 Implementations MUST NOT support function extensions that are not registered.
-Note: if an implementation supports a function extension that is not registered,
-the implementation will not conform to this specification until the function extension is
+
+Notes:
+
+* If an implementation supports a function extension that is not registered,
+the implementation will not conform to this specification unless or until the function extension is
 registered.
-To support a new function extension in an implementation, it is recommended that
-the implementation behaves by default in a "strict" mode that conforms to this specification and
-provides some way of behaving in a "non-strict" mode that supports the new function extension.
+* An implementation may behave by default in a "strict" mode that conforms to this specification
+and offer some way of behaving in a "non-strict" mode that supports unregistered function extensions.
 This ensures that the default behaviour of the implementation does not mislead users into
-thinking that the function extension is interoperable.
-To use the function extension successfully, users have to consciously choose to use non-strict mode.
-If the function extension is later registered, the implementation can then support the
-function extension in the default, strict mode.
+thinking that uregistered function extensions can be used interoperably.
 
 To define which function expressions are well typed,
 a type system is first introduced.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1509,13 +1509,15 @@ For a function expression to be well typed:
 (1) The registered set of function extensions is the source of knowledge regarding the declared return and
 parameter types of a function extension.
 
-Although a function expression being treated as well typed implies its function name is registered,
-an implementation need not treat all registered functions as well typed.
+If a function expression is treated as well typed, the function extensions involved in the expression belong
+to the registered set.
+However, an implementation may treat a function expression as not well typed even if all the function extensions
+involved in the expression are registered.
 Implementations MUST support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
-If an implementation does not support a function extension used in a query, it MUST treat the
-function extension as not well typed.
+If an implementation does not support a function extension used in a query, it MUST treat
+any function expression involving the function extension as not well typed.
 
 (2) As per the grammar, a function expression can occur in three different
 immediate contexts, which lead to the following conditions for well-typedness:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1502,7 +1502,7 @@ The well-typedness of function expressions can now be defined in terms of this t
 
 For a function expression to be well typed:
 
-1. its function name must be registered (see {{iana-fnex}}),
+1. its function name MUST be registered (see {{iana-fnex}}),
 2. its declared type must be well typed in the context in which it occurs, and
 3. its arguments must be well typed for the declared type of the corresponding parameters.
 
@@ -1511,10 +1511,10 @@ parameter types of a function extension.
 
 Although a function expression being treated as well typed implies its function name is registered,
 an implementation need not treat all registered functions as well typed.
-Implementations must support the function extensions defined
+Implementations MUST support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
-If an implementation does not support a function extension used in a query, it should treat the
+If an implementation does not support a function extension used in a query, it MUST treat the
 function extension as not well typed.
 
 (2) As per the grammar, a function expression can occur in three different
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1506,7 +1506,17 @@ For a function expression to be well typed:
 2. its declared type must be well typed in the context in which it occurs, and
 3. its arguments must be well typed for the declared type of the corresponding parameters.
 
-(1) As per the grammar, a function expression can occur in three different
+(1) The registered set of function extensions is the source of knowledge regarding the return and
+parameter types of a function extension.
+
+Although a function expression being treated as well typed implies its function name is registered,
+the reverse is not necessarily true. Implementations must support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+If an implementation does not support a function extension used in a query, it should treat the
+function extension as not well typed.
+
+(2) As per the grammar, a function expression can occur in three different
 immediate contexts, which lead to the following conditions for well-typedness:
 
 {:vspace}
@@ -1521,7 +1531,7 @@ As a `function-argument` in another function expression:
 : The function's declared result type fulfills the following rules for
   the corresponding parameter of the enclosing function.
 
-(2) The arguments of the function expression are well typed when
+(3) The arguments of the function expression are well typed when
 each argument of the function can be used for the declared type of the
 corresponding parameter, according to one of the following
 conditions:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1449,11 +1449,11 @@ otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
 
 For a function expression to be supported, each function name in the expression
-MUST be registered (see {{iana-fnex}}).
+must be supported.
 Implementations MUST support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
-To define which supported function expressions are well typed,
+To define which function expressions are well typed,
 a type system is first introduced.
 
 ### Type System for Function Expressions {#typesys}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1451,7 +1451,7 @@ otherwise the JSONPath implementation MUST raise an error
 For a function expression to be supported, each function extension involved in the expression
 must be supported.
 Implementations MUST support the function extensions defined
-in this document and SHOULD support all other registered function extensions.
+in this document, but are free to support or not support other registered function extensions.
 
 To define which function expressions are well typed,
 a type system is first introduced.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -525,7 +525,7 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
-2. Uses of function extensions must be *well typed*,
+2. Uses of function extensions must be *supported* and *well typed*,
 as described in {{fnex}}.
 
 A JSONPath implementation MUST raise an error for any query which is not
@@ -1443,16 +1443,25 @@ function-argument   = literal /
                       function-expr
 ~~~
 
-Any function expressions in a query must be well formed (by conforming to the above ABNF)
-and well typed,
+Any function expressions in a query must be well formed (by conforming to the above ABNF),
+supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
-To define which function expressions are well typed,
+
+For a function expression to be supported, each function name in the expression
+MUST be registered (see {{iana-fnex}}).
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+To define which supported function expressions are well typed,
 a type system is first introduced.
 
 ### Type System for Function Expressions {#typesys}
 
 Each parameter as well as the result of a function extension must have a declared type.
+
+The registered set of function extensions (see {{iana-fnex}}) is the source of knowledge
+regarding the declared return and parameter types of a function extension.
 
 Declared types enable checking a JSONPath query for well-typedness
 independent of any query argument the JSONPath query is applied to.
@@ -1502,24 +1511,10 @@ The well-typedness of function expressions can now be defined in terms of this t
 
 For a function expression to be well typed:
 
-1. its function name MUST be registered (see {{iana-fnex}}),
-2. its declared type must be well typed in the context in which it occurs, and
-3. its arguments must be well typed for the declared type of the corresponding parameters.
+1. its declared type must be well typed in the context in which it occurs, and
+2. its arguments must be well typed for the declared type of the corresponding parameters.
 
-(1) The registered set of function extensions is the source of knowledge regarding the declared return and
-parameter types of a function extension.
-
-If a function expression is treated as well typed, the function extensions involved in the expression belong
-to the registered set.
-However, an implementation may treat a function expression as not well typed even if all the function extensions
-involved in the expression are registered.
-Implementations MUST support the function extensions defined
-in this document, but are free to support or not support other registered function extensions.
-
-If an implementation does not support a function extension used in a query, it MUST treat
-any function expression involving the function extension as not well typed.
-
-(2) As per the grammar, a function expression can occur in three different
+(1) As per the grammar, a function expression can occur in three different
 immediate contexts, which lead to the following conditions for well-typedness:
 
 {:vspace}
@@ -1534,7 +1529,7 @@ As a `function-argument` in another function expression:
 : The function's declared result type fulfills the following rules for
   the corresponding parameter of the enclosing function.
 
-(3) The arguments of the function expression are well typed when
+(2) The arguments of the function expression are well typed when
 each argument of the function can be used for the declared type of the
 corresponding parameter, according to one of the following
 conditions:
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1506,7 +1506,7 @@ For a function expression to be well typed:
 2. its declared type must be well typed in the context in which it occurs, and
 3. its arguments must be well typed for the declared type of the corresponding parameters.
 
-(1) The registered set of function extensions is the source of knowledge regarding the return and
+(1) The registered set of function extensions is the source of knowledge regarding the declared return and
 parameter types of a function extension.
 
 Although a function expression being treated as well typed implies its function name is registered,
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1448,10 +1448,10 @@ supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
 
-For a function expression to be supported, each function name in the expression
+For a function expression to be supported, each function extension involved in the expression
 must be supported.
 Implementations MUST support the function extensions defined
-in this document, but are free to support or not support other registered function extensions.
+in this document and SHOULD support all other registered function extensions.
 
 To define which function expressions are well typed,
 a type system is first introduced.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1455,11 +1455,6 @@ in this document, but are free to support or not support other registered functi
 
 Implementations MUST NOT support function extensions that are not registered.
 
-Note: an implementation may behave by default in a "strict" mode that conforms to this specification
-and offer some way of behaving in a "non-strict" mode that supports unregistered function extensions.
-This ensures that the default behaviour of the implementation does not mislead users into
-thinking that uregistered function extensions can be used interoperably.
-
 To define which function expressions are well typed,
 a type system is first introduced.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1455,12 +1455,7 @@ in this document, but are free to support or not support other registered functi
 
 Implementations MUST NOT support function extensions that are not registered.
 
-Notes:
-
-* If an implementation supports a function extension that is not registered,
-the implementation will not conform to this specification unless or until the function extension is
-registered.
-* An implementation may behave by default in a "strict" mode that conforms to this specification
+Note: an implementation may behave by default in a "strict" mode that conforms to this specification
 and offer some way of behaving in a "non-strict" mode that supports unregistered function extensions.
 This ensures that the default behaviour of the implementation does not mislead users into
 thinking that uregistered function extensions can be used interoperably.
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1453,6 +1453,19 @@ must be supported.
 Implementations MUST support the function extensions defined
 in this document, but are free to support or not support other registered function extensions.
 
+Implementations MUST NOT support function extensions that are not registered.
+Note: if an implementation supports a function extension that is not registered,
+the implementation will not conform to this specification until the function extension is
+registered.
+To support a new function extension in an implementation, it is recommended that
+the implementation behaves by default in a "strict" mode that conforms to this specification and
+provides some way of behaving in a "non-strict" mode that supports the new function extension.
+This ensures that the default behaviour of the implementation does not mislead users into
+thinking that the function extension is interoperable.
+To use the function extension successfully, users have to consciously choose to use non-strict mode.
+If the function extension is later registered, the implementation can then support the
+function extension in the default, strict mode.
+
 To define which function expressions are well typed,
 a type system is first introduced.
 


gregsdennis:
Registration is orthogonal to well-typedness.

With this change implementations are disallowed from supporting custom functions.  This may be a crucial feature for users who don't care about interoperability.

glyn:
> Registration is orthogonal to well-typedness.
> 
> With this change implementations are disallowed from supporting custom functions. This may be a crucial feature for users who don't care about interoperability.

If the spec allows implementations to support custom functions, then it's the spec that doesn't care about interoperability.

I think it's better if the spec is strict/interoperable. Implementers who want custom functions and don't care about interoperability are free to drop the requirement that function names be registered (but they mustn't claim spec compliance in doing so).

cabo:
I'm not sure we are fully understanding our evolution concept here.  Whatever we say, there will be a period when an implementation for a new name is out there, while the name is still being registered.  Reserving a space for such non-registered names has its problems (RFC 6648).  So how is an implementer supposed to do this?

glyn:
I'm still trying to understand what we mean by interoperation and I'm happy
to be corrected, but let me set out my current thinking.

An implementer wishing to register a function extension can use it before
registration. While doing so, they are not conforming to the spec and their
implementation is not interoperable. If and when the function extension is
registered, the implementation can conform to the spec and be interoperable
with other implementations that implement the function extension.
(Admittedly, this requires implementers to take a risk when attempting to
introduce a new function extension.)

Without this, wouldn't we be allowing multiple non-interoperable
implementations to arise? Implementers might not bother getting round to
registering their custom function extensions. This could also encourage the
use of unregistered, but commonly implemented function extensions.

danielaparker:
> Without this, wouldn't we be allowing multiple non-interoperable implementations to arise? 

The only part of JSONPath that could _ever_ be interoperable is what's defined in a particular version of the core spec. And hopefully there aren't many versions. It takes years for a new version to be widely supported, even for popular specs. JSONSchema still has spotty support for 2019-09 and 2020-12, most users are stuck on version 7. 

> the implementation can ... be interoperable with other implementations that implement the function extension

And not interoperable with all other implementations that do not implement the function extension, which presumably would be the vast majority of them. CBOR has extensions, but there is fallback behaviour if an extension tag is not recognized. It's hard to imagine fallback behaviour if a function name is not recognized.

Daniel

glyn:
> > Without this, wouldn't we be allowing multiple non-interoperable implementations to arise?
> 
> The only part of JSONPath that could _ever_ be interoperable is what's defined in a particular version of the core spec. And hopefully there aren't many versions. It takes years for a new version to be widely supported, even for popular specs. JSONSchema still has spotty support for 2019-09 and 2020-12, most users are stuck on version 7.
> 
> > the implementation can ... be interoperable with other implementations that implement the function extension
> 
> And not interoperable with all other implementations that do not implement the function extension, which presumably would be the vast majority of them. CBOR has extensions, but there is fallback behaviour if an extension tag is not recognized. It's hard to imagine fallback behaviour if a function name is not recognized.
> 
> Daniel

I don't follow the logic. Suppose there are N (>=2) implementations of a given version of a JSONPath RFC, that an extra function extension (not in the RFC) is defined in the subregistry, and all N implementations happen to implement that function. (If you find this hard to believe, set N=2 and suppose the same person wrote both implementations.) Why would the RFC plus that function extension then not be interoperable?

cabo:
I think we need to have a slightly sharper terminology to discuss this.

First of all, we have to accept that standards evolve.
Extension points are an approach to make this evolution less painful, but they cannot take the pain away completely.

Note that we don’t have implementations that are intended to interoperate among each other, but we have implementations that operate in an interoperable way on some data.

One objective of the extension point is to achieve backward compatibility, i.e., making sure that evolved systems can interoperably operate on data that was designed at an earlier point of the evolution.
I think we have achieved that.

The other objective of an extension point is forward compatibility: older systems providing some form of successful, interoperable operation when confronted with new data.
Forward compatibility will always be limited: We wouldn’t evolve if we could achieve the new objectives with the older systems as well.

With evolution, there is no way that implementations that are on different levels of evolution can be fully interoperable with all instances.

So what is the limitation here?

What we still could have:

(1) Older systems can successfully detect the presence of an extension they do not support.
So we fully prevent false interoperability.
(2) Extensions have names (the function name), so we can have accurate diagnostics.
(3) In several architectures, will have a way to be added dynamically to existing systems, based on the function name.

These statements require collision-free naming of extensions.
A registry provides the service of collision-free naming.
Other such services can be defined, usually based on hierarchical naming (ASN.1 Object IDs, various forms of URIs, etc.).  Function names are not designed to interface with such naming systems.
RFC 6648 tells us why it is a bad idea to embed a temporary component into what is likely to become a permanent name as deployment of the temporary component spreads.

So what can we improve?

Anyone performing a limited deployment of an extension in order to validate it and complete its detailed definition undergoes a gamble with the name.  
Registering the name early reduces the stakes, but also discloses information (usually, function names should be descriptive).
Registering the name late means there is greater potential for collisions, a risk similar to others that developers already need to manage.

I believe the best contribution we can make is having a good registry policy in place.
Section 3.2 goes a long way, but we probably should have another look.
Obviously, the choice of the designated experts who implement these policies will also influence the success; I believe it is a strength of the IETF registry model that it recognizes that not all decisions can be predetermined in a static ruleset.

Grüße, Carsten

glyn:
Thanks for those helpful clarifications @danielaparker and @cabo. So, in this PR, I am focussing on consistency of named extensions across implementations. The objective I would like to achieve is that any two implementations should provide consistent behaviour for a function extension with a given name in the sense that:

1. If an implementation does not (yet) support the function extension, any query using the function extension will be rejected with an error.

2. If an implementation supports the function extension, any query using the function extension will behave according to the registered definition of the function extension.

Implementations should also be free to provide a (non-default) mode of operation where unregistered function extensions (perhaps on the way to being registered) are also supported. Users choosing to use such a non-default mode are then running a conscious risk of inconsistent behaviour between implementations.

What should the spec say about such matters? It seems to me that the current spec can be interpreted such that unregistered function extensions can be offered freely to users who could then encounter inconsistent behaviour between implementations. I think the current text of this PR is a step in the direction of making the behaviour more consistent, at least for implementations running in "spec compliant mode".

timbray:
I generally agree with Guy's points 1. and 2.

gregsdennis:
> Implementations should also be free to provide a (non-default) mode of operation where unregistered function extensions (perhaps on the way to being registered) are also supported. Users choosing to use such a non-default mode are then running a conscious risk of inconsistent behaviour between implementations.

This is my primary concern.  Thank you for stating it.

However my concern is that this change says that such unregistered functions result in a not-well-typed expression.  That's simply not the case.  My user can create a custom function for my implementation, and the expression will remain well-typed because they have to choose a type for their custom function when they create it.

I suggest an alternate change that states (separately from any well-typedness language):

> Implementations MAY support custom functions, but such support MUST be behind a configuration option which is defaulted off.  Users who make use of such custom functions acknowledge that the resulting behavior is expected to be unsupported by other implementations.

glyn:
> > Implementations should also be free to provide a (non-default) mode of operation where unregistered function extensions (perhaps on the way to being registered) are also supported. Users choosing to use such a non-default mode are then running a conscious risk of inconsistent behaviour between implementations.
> 
> This is my primary concern. Thank you for stating it.
> 
> However I feel that this change says that such unregistered functions result in a not-well-typed expression. That's simply not the case. My user can create a custom function for my implementation, and the expression will remain well-typed because they have to choose a type for their custom function when they create it.
> 
> I suggest an alternate change that states (separately from any well-typedness language):
> 
> > Implementations MAY support custom functions, but such support MUST be behind a configuration option which is defaulted off.  Users who make use of such custom functions acknowledge that the resulting behavior is expected to be unsupported by other implementations.

Some language along those lines may turn out to be necessary, but talking about configuration options feels a bit too prescriptive. Implementers have total freedom once they diverge from the spec, so we can simply leave non-default behaviour unspecified. Therefore, I'd like to focus on the default behaviour in this PR.

I'd like the use of custom functions to fail with an error due to an invalid query. The rationale for this PR is that since a custom function is by default undefined, its result type is unknown and any function expression involving the custom function will then not be well typed.

gregsdennis:
> Implementers have total freedom once they diverge from the spec, so we can simply leave non-default behaviour unspecified.

This is my point. I'm not saying that we specify non-spec behavior. I'm saying that implementations have to hide such behavior behind an option, default off.

cabo:
Our definition of well-typed requires a source of knowledge as to the return and parameter types of a function extension.

I think it is within our remit to specify that we only accept the registry as a source of that knowledge.
(I think we could make this more explicit in this PR.)

I don't think we can require implementations to instantaneously update with registry updates, and much less to implement each newly registered function, so, more generally, we will already have to deal with interoperability failures due to function extensions that are not implemented (the predefined function extensions are privileged here).

How much can we put implementations on a leash with respect to accepting their own extensions (inside or outside of defined extension points)?  Generally, we can't stop innovation.  But we have (limited) authority to stop people from calling a query that is not interoperable due to lack of JSONPath compliance a (standard) JSONPath query.  And we can strongly recommend implementing and defaulting to a strict mode (as a quality of implementation point), preventing interoperability failures by a query author believing their query is interoperable when it is not.

draft-ietf-jsonpath-base.md
```diff
@@ -1502,8 +1502,9 @@ The well-typedness of function expressions can now be defined in terms of this t
 
 For a function expression to be well typed:
 
-1. its declared type must be well typed in the context in which it occurs, and
-2. its arguments must be well typed for the declared type of the corresponding parameters.
+1. its function name must be registered (see {{iana-fnex}}),
```
cabo, line 1505:
```suggestion
1. its function name must be registered (see {{iana-fnex}}), which provides its declared type and the declared types of its parameters;
```

this would make it clear why that is a requirement.

(Do we want to spend MUST interoperability requirements here?)

glyn: @cabo I made a similar change in 5f210fed5a2987e586f197d782cc1c4cf8edd824. Note that the bullets (1), (2), and (3) correspond to the summary points 1, 2, and 3.

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,21 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+2. its declared type must be well typed in the context in which it occurs, and
+3. its arguments must be well typed for the declared type of the corresponding parameters.
 
-(1) As per the grammar, a function expression can occur in three different
+(1) The registered set of function extensions is the source of knowledge regarding the return and
```
cabo, line 1509:
```suggestion
(1) The registered set of function extensions is the source of knowledge regarding the declared return and
```

I like saying that these are declared (as opposed to actual, as in a dynamic type system).

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,21 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+parameter types of a function extension.
+
+Although a function expression being treated as well typed implies its function name is registered,
+the reverse is not necessarily true. Implementations must support the function extensions defined
```
cabo, line 1513: I don't know what the reverse is: registered ➔ well-typed?  Of course not.
Or is it the inverse?
Better be explicit?

draft-ietf-jsonpath-base.md
glyn: Section 2.1 already states:
> A JSONPath implementation MUST raise an error for any query which is not
well formed and valid.

So I don't think we need to use MIST here also.

cabo: We don't need to, but we could.  I think it would help.

draft-ietf-jsonpath-base.md
glyn: Fixed in e25a509952eeae5c66facd5fd98a073c844b1b20.

draft-ietf-jsonpath-base.md
glyn: Ok, see what you make of 3819d94bc7e821cb529003293dee54ec680cd43b.

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,22 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+(1) The registered set of function extensions is the source of knowledge regarding the declared return and
+parameter types of a function extension.
+
+Although a function expression being treated as well typed implies its function name is registered,
+an implementation need not treat all registered functions as well typed.
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+If an implementation does not support a function extension used in a query, it MUST treat the
+function extension as not well typed.
```
cabo, line 1518: See my previous comment: The error message is allowed to say the function extension is not supported, so "as not well typed" is overstating it.

draft-ietf-jsonpath-base.md
cabo: Thanks, that is what I meant.

draft-ietf-jsonpath-base.md
glyn: Sorry - I've lost track of your previous comment.

The reason I say  it MUST treat the function extension as not well typed is so that it's clear that section 2.1 applies (a query that is not well typed MUST raise an error).

A helpful implementation will provide diagnostics as part of the error - in this case, an error message saying the function extension is not supported. Do we have to spell that out or can we leave that as a quality of implementation consideration?

glyn: Ah, found your previous comment https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/commit/e25a509952eeae5c66facd5fd98a073c844b1b20#r109974852. Let me try to iron that out.

glyn: See bdf27be40ee2f5d5995dcdd6a13d7944952b3869.

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,24 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+1. its function name MUST be registered (see {{iana-fnex}}),
+2. its declared type must be well typed in the context in which it occurs, and
+3. its arguments must be well typed for the declared type of the corresponding parameters.
 
-(1) As per the grammar, a function expression can occur in three different
+(1) The registered set of function extensions is the source of knowledge regarding the declared return and
+parameter types of a function extension.
+
+If a function expression is treated as well typed, the function extensions involved in the expression belong
+to the registered set.
```
gregsdennis, line 1513: The way this logic is stated means that well-typedness automatically implies membership in the registry.  That's not precisely the case.

I think what you're meaning is that membership in the registry is a prerequisite to being considered well-typed.

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,24 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+However, an implementation may treat a function expression as not well typed even if all the function extensions
+involved in the expression are registered.
```
gregsdennis, line 1515: Is this a roundabout way of saying that an implementation is allowed to not support a registered function?

draft-ietf-jsonpath-base.md
```diff
@@ -1502,10 +1502,24 @@ The well-typedness of function expressions can now be defined in terms of this t
...
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+If an implementation does not support a function extension used in a query, it MUST treat
+any function expression involving the function extension as not well typed.
```
gregsdennis, line 1520: I still don't understand this association between registration and well-typedness.

If a function isn't recognized, then _that_ is the error.  The implementation cannot determine the well-typedness of a function it doesn't know.  It may be that the expression _is_ well-typed, but the implementation just doesn't support the function.  In this case, raising an error saying that it's not well-typed is incorrect.

glyn: Section 2.1 describes the high level error processing model: if a query is not well formed (conforms to the ABNF) and valid, an error MUST be raised (before the argument is considered). Validity (of a well formed query) means fulfilling all the semantic requirements of the spec: integers are in range and uses of function extensions are well typed.

We _could_ introduce a third category of semantic requirement to cover registration, but that's not necessary and it would introduce more function extension detail into section 2.1 which I think would be a bit of a distraction at that stage in the document.

As you point out, if an implementation can't determine the result type of a function extension, it can't know that an expression involving the function extension is well typed. I am proposing that in such a situation, the implementation should presume that the expression is not well typed. (Any error message can certainly point out that the function extension isn't recognised, and I'd expect helpful implementations to do that.)

There is a parallel in distributed two phase commit which may be helpful. When the global outcome of a transaction is (permanently) in doubt, there are three policies for handling the situation. "Presumed abort" presumes the transaction failed. "Presumed commit" presumes the transaction succeeded. "Presumed nothing" makes no presumption: it essentially waits until the global outcome can be determined (or at least asserted, e.g. by a human operator). The parallel is that I am proposing that when the well typedness of a function expression is not known, we should presume it is not well typed. The policy of presuming the function expression is well typed is non-sensical. I _think_ you are suggesting that it would be better not to make a presumption, _but to fail with an error anyway_, which presumes invalidity (of a different kind).

draft-ietf-jsonpath-base.md
glyn: That's certainly an implication of that sentence, but there is another implication: the query must fail with an error.

draft-ietf-jsonpath-base.md
glyn: > The way this logic is stated means that well-typedness automatically implies membership in the registry. That's not precisely the case.

Notice the phrase "treated as well typed". What's incorrect about the statement?

> 
> I think what you're meaning is that membership in the registry is a prerequisite to being considered well-typed.

That's certainly true, but I also had in mind the effect on the user. If a query succeeds, the user can safely assume that any function extensions in the query are registered. Perhaps I need to spell that out.

draft-ietf-jsonpath-base.md
gregsdennis: > Any error message can certainly point out that the function extension isn't recognised, and I'd expect helpful implementations to do that.

If this is your expectation, then why require that the error is "not well-typed"?  Either the spec is going to dictate errors or it's not.

gregsdennis: > when the well typedness of a function expression is not known, we should presume it is not well typed.

We don't have enough information.  Making this presumption is wrong, and requiring implementations to do so, is doubly wrong.

glyn: > > Any error message can certainly point out that the function extension isn't recognised, and I'd expect helpful implementations to do that.
> 
> If this is your expectation, then why require that the error is "not well-typed"? Either the spec is going to dictate errors or it's not.

There is no requirement that the error is "not well-typed". The only requirement is that the query fails with an error. This is consistent with the policy in section 2.1.

glyn: > > when the well typedness of a function expression is not known, we should presume it is not well typed.
> 
> We don't have enough information. Making this presumption is wrong, and requiring implementations to do so, is doubly wrong.
> 
> Implementations should error, yes. And they should do so at the parsing stage. But an unrecognized function is just that: unrecognized; and the spec has no business saying the error should suggest anything else.

This PR does not assume there will be an error message along the lines of "not well-typed". An implementation is free to produce an error message along the lines "error: function extension foo() is not supported".

Remember my fundamental reason for adopting this policy is to avoid raising too many details about function extensions to the level of section 2.1 (and perhaps elsewhere in the spec).

cabo: > This PR does not assume there will be an error message along the lines of "not well-typed".

I think Greg and I are reading "treat as not well-typed" as saying just that.  I think we are looking for some other way of saying that without that implication, e.g., "cannot treat as well-typed".

glyn: Ah, since I may be in a minority of one, I'll try adjusting the approach.

glyn: @gregsdennis and @cabo: please see the update. Section 2.1 is impacted slightly, but perhaps this is worth it to cleanly separate supportedness from well-typedness.

cabo: It may be worth keeping in mind that is it much easier to update an implementation with new registry information than to actually implement the functions (in particular if they rely on platform capabilities you may not have).
So implementations may be able to provide accurate information about welltypedness of a query without being able to actually execute it -- we need to keep these two separate.

draft-ietf-jsonpath-base.md
```diff
@@ -525,7 +525,7 @@ to the JSONPath processing (e.g., index values and steps) MUST be
 within the range of exact values defined in I-JSON {{-i-json}}, namely
 within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
 
-2. Uses of function extensions must be *well typed*,
+2. Uses of function extensions must be *supported* and *well typed*,
```
cabo, line 528: This section is about validity of a query.  Validity is independent of what you have implemented.  A query may be

* valid
* not valid
* of unknown validity (e.g., because you don't know a registration that might have occurred).

cabo: Whether you actually support a valid query is a separate question.  (You can't really "support" invalid or unknown-valid-or-not queries.)

draft-ietf-jsonpath-base.md
cabo: This leads to the observation that there is an objective and a subjective view of validity.  A query may be valid objectively (in a god-view, who always has the most recent registration updates), it may also be valid subjectively (which should only be true if it is also valid objectively), or it may be of an unknown validity subjectively (cannot happen from an objective point of view).

glyn: @cabo & @gregsdennis: I made another change which makes supportedness and (subjective) registration of a function extension othogonal.

Let's consider the four combinations for a function extension `foo()`:

* supported and registered - no error
* unsupported and registered - error "foo() unsupported" (see §2.4)
* supported and unregistered - error A (see below)
* unsupported and unregistered - either error "foo() unsupported" (see §2.4) or error A (see below), at the choice of the implementation

What should error A be for an unregistered function extension? Since well-typedness of the containing function expression cannot be known, we could raise an error "function expression involving foo() not necessarily well typed"). Alternatively, we could separate unregistration from well-typedness and raise an error "unregistered foo()" without saying anything about well-typedness.

If we separated unregistration from well-typeness, we could either make unregistration its own kind of invalidity (i.e. a peer of unsupported and not well typed) or we could collapse unregistration and unsupported into a single kind of invalidity (essentially saying that an implementation cannot support a function extension that is unregistered, at least not in strict/spec-compliant mode).

I didn't separate subjective and objective registration in the above. Perhaps that needs doing. Subjective registration seems more interesting to implementers while objective registration may be more interesting to users.

Any thoughts on the best way forward?

gregsdennis: > essentially saying that an implementation cannot support a function extension that is unregistered, at least not in strict/spec-compliant mode

I think this is the correct approach.  An implementation would need to explicitly "unsupport" any custom functions when operating in strict mode.  Further, that strict mode should be the default.

Regarding subjective vs objective support, the implementation supports what it supports.  A user can't realistically expect an implementation to support a function that was added to the registry after the library was released, for example.  The library should document what functions it supports (just like any other functionality), and the user should be expected to read such documentation.  (Strangely, I'm having this same conversation in JSON Schema Land.)

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,26 @@ function-argument   = literal /
...
-Any function expressions in a query must be well formed (by conforming to the above ABNF)
-and well typed,
+Any function expressions in a query must be well formed (by conforming to the above ABNF),
+supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
-To define which function expressions are well typed,
+
+For a function expression to be supported, each function name in the expression
+MUST be registered (see {{iana-fnex}}).
```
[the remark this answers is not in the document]
gregsdennis: (@glyn apparently I had left this as a review comment, then didn't submit the review.  My concern still applies to the current text.)

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,26 @@ function-argument   = literal /
...
-and well typed,
+Any function expressions in a query must be well formed (by conforming to the above ABNF),
+supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
+
+For a function expression to be supported, each function name in the expression
+must be supported.
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
```
gregsdennis, line 1454:
```suggestion
Implementations MUST support the function extensions defined
in this document, and SHOULD support other registered function extensions.
```

I think this gives a little more pressure to implement other registered functions but still an escape if they don't.

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: > A function can be in the registry and still be unsupported by an implementation. This seems to say that the sole requirement for support is its registration. But that's not the case: _implementation_ support is also required.

What I meant was that a function extension being supported implies the function name is registered. (But there may be multiple functions in a function expression.) Needs rewording.

> 
> Secondly, I think we're conflating "function expression" and "function name." Each function expression should only have a single name, but the way this is worded seems to imply that a function expression can have multiple names.

Take the function expression `foo(bar())`. There are two function names in the expression.

draft-ietf-jsonpath-base.md
glyn: I was trying to avoid being prescriptive, but I'd be comfortable with a statement like the above saying "[...] and SHOULD support all other registered function extensions".

draft-ietf-jsonpath-base.md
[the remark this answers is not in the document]
glyn: I've adjusted the wording.

draft-ietf-jsonpath-base.md
glyn: I've added the SHOULD requirement.

cabo: I don't think we should say this, as there probably will be registrations for very platform- or application-specific functions.

Some IETF registries have a "recommended" column, which is kind of the inverse: if it is not recommended, you need a good reason to implement it at all.

I think at some point we'll roll up a set of registered functions and say that there is an expectation these will be available, kind of a JSONPath 1.1.  But remember that the largest pressure for extensions is in the projection/transformation space, and we don't even have an extension point for this.

glyn: Ok, reverted the SHOULD change. At least I'm not now in a minority of one.

gregsdennis: > there probably will be registrations for very platform- or application-specific functions.

God, I hope not.  The registry should be a list of functions that should and _can_ be globally supported.  Platform- and application-specific function are by definition _not_ interoperable.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,26 @@ function-argument   = literal /
...
 
-Any function expressions in a query must be well formed (by conforming to the above ABNF)
-and well typed,
+Any function expressions in a query must be well formed (by conforming to the above ABNF),
+supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
+
+For a function expression to be supported, each function extension involved in the expression
+must be supported.
```
gregsdennis, line 1452: This still feels weird to me.  We have

- function expression
- function extension
- expression

```suggestion
For a filter expression to be supported, each function extension involved in the expression
must be supported.
```

I think this is what we mean.  We call the entire expression a "filter expression."  We don't define a "function expression" anywhere that I'm aware (although the function syntax itself is `function-expr` which may be a misnomer).

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,39 @@ function-argument   = literal /
...
+the implementation will not conform to this specification until the function extension is
+registered.
+To support a new function extension in an implementation, it is recommended that
+the implementation behaves by default in a "strict" mode that conforms to this specification and
+provides some way of behaving in a "non-strict" mode that supports the new function extension.
+This ensures that the default behaviour of the implementation does not mislead users into
+thinking that the function extension is interoperable.
+To use the function extension successfully, users have to consciously choose to use non-strict mode.
+If the function extension is later registered, the implementation can then support the
+function extension in the default, strict mode.
```
glyn, line 1467: @cabo & @gregsdennis: I added the above paragraph to clarify the situation when an implementation supports an unregistered function extension. It may need trimming down, but perhaps not. WDYT?

draft-ietf-jsonpath-base.md
glyn: `function-expr` is defined in the ABNF as you observe (and includes expressions involving nested function calls). That ABNF section is immediately followed by "Any function expressions in a query [...]". §2.4.1 is entitled "Type System for Function Expressions" and §2.4.3 "Well-typedness of Function Expressions". So I think the term is well defined and can and should be used here in preference to "filter expression".

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,39 @@ function-argument   = literal /
...
+
+For a function expression to be supported, each function extension involved in the expression
+must be supported.
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+Implementations MUST NOT support function extensions that are not registered.
+Note: if an implementation supports a function extension that is not registered,
+the implementation will not conform to this specification until the function extension is
+registered.
```
gregsdennis, line 1459: I think this is unnecessary.  It also implies that any custom functions will be proposed for registration, which is unlikely.  Mostly, I expect users will extend an implementation with some custom function for their use only and be satisfied with that without considering registration.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,39 @@ function-argument   = literal /
...
+To support a new function extension in an implementation, it is recommended that
+the implementation behaves by default in a "strict" mode that conforms to this specification and
+provides some way of behaving in a "non-strict" mode that supports the new function extension.
```
gregsdennis, line 1462: How about:

> Custom functions MUST be supported only by a configuration option. Support of such functions MUST NOT be the default behavior.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,39 @@ function-argument   = literal /
...
+This ensures that the default behaviour of the implementation does not mislead users into
+thinking that the function extension is interoperable.
+To use the function extension successfully, users have to consciously choose to use non-strict mode.
```
gregsdennis, line 1465: This should be a note or comment.

draft-ietf-jsonpath-base.md
gregsdennis, line 1467: I don't believe this is necessary.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
gregsdennis: What you have is weird and feels redundant and a bit confusing.

glyn: I don't follow. This sentence defines function expression support (required for validity by the previous paragraph) with function extension support. Why is this redundant? What's confusing about it?

draft-ietf-jsonpath-base.md
glyn: Does this help?
```suggestion
Note: if an implementation supports a function extension that is not registered,
the implementation will not conform to this specification unless or until the function extension is
registered.
```

draft-ietf-jsonpath-base.md
glyn: I'm trying to avoid constraining non-spec behaviour such as support for unregistered functions. Perhaps it isn't clear that the non-normative "Note:" is meant to extend to the end of the paragraph?

draft-ietf-jsonpath-base.md
glyn: It is intended to be. Maybe I need to make that clearer.

draft-ietf-jsonpath-base.md
glyn: Perhaps not, but I was trying to explain the recommended route to introducing a new function extension. If we omit this sentence, then it gives the impression users will always have to use non-strict mode to use the function extension.

draft-ietf-jsonpath-base.md
gregsdennis: I think the whole note is unnecessary.  I think it's generally implied that custom functionality outside of the scope of a spec.

draft-ietf-jsonpath-base.md
gregsdennis: > I'm trying to avoid constraining non-spec behaviour

But that's exactly what you're doing by saying "strict mode" and "non-strict mode."

draft-ietf-jsonpath-base.md
glyn: Ok, I deleted that note.

draft-ietf-jsonpath-base.md
glyn: I've clarified that the strict vs non-strict stuff is a non-normative note. I think it's worth having even though it doesn't constrain non-spec behaviour at all.

draft-ietf-jsonpath-base.md
glyn: The note should be clearer now as it has been slimmed down a lot.

draft-ietf-jsonpath-base.md
glyn: Ok, I've deleted that sentence.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,33 @@ function-argument   = literal /
...
+
+For a function expression to be supported, each function extension involved in the expression
+must be supported.
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+Implementations MUST NOT support function extensions that are not registered.
+
+Note: an implementation may behave by default in a "strict" mode that conforms to this specification
+and offer some way of behaving in a "non-strict" mode that supports unregistered function extensions.
```
gregsdennis, line 1459:
```suggestion
Note: an implementation MAY offer some way of switching between a "strict" mode that conforms to this specification, which must be the default,
and  a "non-strict" mode that supports unregistered function extensions.
```

Maybe?

glyn: That's reasonable except using "MAY" in non-normative text feels a little over the top and "which must be the default" also seems slightly too prescriptive. How about this instead?
```suggestion
Note: an implementation may offer a default "strict" mode that conforms to this specification
and a "non-strict" mode that supports unregistered function extensions.
```

glyn: I don't want to waste any more time discussing the unspec'd behaviour of implementations, such as what that should be like and whether it should appear in normative text, so I deleted the note.

draft-ietf-jsonpath-base.md
cabo: So let me make this point clearer: This section is about queries and when they are valid.  It is now about implementations.

draft-ietf-jsonpath-base.md
cabo: Again, support is not a function of the query, so we cannot make this statement.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,28 @@ function-argument   = literal /
...
+supported, and well typed,
 otherwise the JSONPath implementation MUST raise an error
 (see {{synsem-overview}}).
+
+For a function expression to be supported, each function extension involved in the expression
+must be supported.
+Implementations MUST support the function extensions defined
+in this document, but are free to support or not support other registered function extensions.
+
+Implementations MUST NOT support function extensions that are not registered.
```
cabo, line 1456: This is not a very realistic requirement.
In particular during the time from developing a function extension to stabilizing it and making the specification widely available.

draft-ietf-jsonpath-base.md
```diff
@@ -1443,17 +1443,28 @@ function-argument   = literal /
...
+
 To define which function expressions are well typed,
 a type system is first introduced.
 
 ### Type System for Function Expressions {#typesys}
 
 Each parameter as well as the result of a function extension must have a declared type.
 
+The registered set of function extensions (see {{iana-fnex}}) is the source of knowledge
```
cabo, line 1465: This is maybe the place where we concede that an implementation may support further function extensions, which then are obviously not interoperable (and their use makes the queries using them non-interoperable).

draft-ietf-jsonpath-base.md
glyn: My thinking was that during that period of time, the implementation would no longer conform to the spec. But it then becomes tricky to talk about that situation in the spec, as the back-and-forth between Greg and I has shown.

So perhaps this PR should be abandoned and I/we have to accept that there will be a free for all of custom function extensions, some of which may be registered and others of which will never get that far.

draft-ietf-jsonpath-base.md
glyn: Thanks for this important point. Another reason to abandon the PR.

glyn:
Closing this PR. There are too many problems with it and I'm not even sure that the original goal -- to define what happens when an implementation encounters an unregistered function extension in a query -- is achievable without getting into implementation detail. Clearly implementations must fail if they encounter function extensions they don't support.

Issue #468, opened 2023-05-08
Editorial feedback from Erik Auerswald


glyn:
@auerswal wrote:

Hello JSONPath Working Group,

thanks for your work on standardizing query expressions for JSON!

I have not yet used JSONPath and looked at the current I-D to learn
about it.  While reading draft-ietf-jsonpath-base-13, I wondered for
quite some time how "selectors" fit into a JSONPath expression.

I also wondered about the meaning of early examples using selector
lists inside a segment, e.g., "$..book[0,1]", and if this only supported
index selectors.  Next came the "$.o[*,*]" example, that additionally
showed wildcard selectors.  But only section 2.5.1 "Child Segment", much
later in the document, introduced the generic "comma-separated sequence
of selectors" without restriction to one or two selector variants.

Perhaps the initial descriptions of "segments" could be expanded a bit
to already mention "selectors"?

I'd like to suggest the following small additions to the early parts of
the draft:

1. Fill the square brackets in the definition of "segment" in the
   "Terminology" section 1.1 with "<selectors>":

   OLD:

     Segment:  One of the constructs which select children ([]) or
        descendants (..[]) of an input value.

   NEW:

     Segment:  One of the constructs which select children ([\<selectors\>])
        or descendants (..[\<selectors\>]) of an input value.

2. Fill the square brackets in the description in section 1.4.2 "Segments"
   with "\<selectors\>":

   OLD:

      Segments select children ([]) or descendants (..[]) of an input
      value.

   NEW:

      Segments select children ([\<selectors\>]) or descendants
      (..[\<selectors\>]) of an input value.

3. Mention in section 1.4.2 "Segments" that bracket notation allows to use
   one or more selectors per segment (and supports all selector variants):

   OLD:

      A JSONPath expression may use a combination of bracket and dot
      notations.

   NEW:

      Bracket notation supports all selector variants.  It also allows to
      use more than one selector per segment (commas between selectors).

      A JSONPath expression may use a combination of bracket and dot
      notations.

Best regards,
Erik

P.S. Please keep me in Cc: of replies, since I am not subscribed to this
     mailing list.

PR #469, opened 2023-05-11
Editorial improvements

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/468

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -396,7 +396,7 @@ $.store.book[0].title
 ~~~~
 
 Bracket notation contains a comma separated list of one or more selectors of any kind.
-Selectors are detailed in the next section. 
+Selectors are detailed in the next section.
 
 A JSONPath expression may use a combination of bracket and dot notations.
 
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1,7 +1,7 @@
 ---
 v: 3
 
-docname: draft-ietf-jsonpath-base-14
+docname: draft-ietf-jsonpath-base-latest
 cat: std
 consensus: 'true'
 submissiontype: IETF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1,7 +1,7 @@
 ---
 v: 3
 
-docname: draft-ietf-jsonpath-base-latest
+docname: draft-ietf-jsonpath-base-14
 cat: std
 consensus: 'true'
 submissiontype: IETF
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -276,14 +272,18 @@ Unicode Scalar Value:
   In other words, integers in either of the inclusive base 16 ranges 0 to D7FF and
   E000 to 10FFFF.  Both JSON string values and JSONPath queries are sequences of Unicode scalar values.
 
-Singular Query:
-: A JSONPath expression built from segments each of which, regardless of the input value,
-  produces a nodelist containing at most one node.
+Segment:
+: One of the constructs which select children (`[<selectors>]`)
+  or descendants (`..[<selectors>]`) of an input value.
```
glyn, line 277: Note to reviewers: I moved the updated definition of Segment down so that it is just before the definition of Selector. If I had left it in the previous position, I would have felt the need to add a reference from the definition of Segment to the definition of Selector, which I didn't want to get into.

draft-ietf-jsonpath-base.md
```diff
@@ -1,7 +1,7 @@
 ---
 v: 3
 
-docname: draft-ietf-jsonpath-base-latest
+docname: draft-ietf-jsonpath-base-14
```
cabo, line 4: This is probably not really part of the PR?

draft-ietf-jsonpath-base.md
```diff
@@ -381,7 +381,7 @@ of the evaluation of a filter expression ({{filter-selector}}).
 
 ### Segments
 
-Segments select children (`[]`) or descendants (`..[]`) of an input value.
+Segments select children (`[<selectors>]`) or descendants (`..[<selectors>]`) of an input value.
```
cabo, line 384: The only problem I have is that this amplifies a bit that segments are always bracket notation, which they aren't.

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: Whoops! Reverted.

draft-ietf-jsonpath-base.md
gregsdennis: > segments are always bracket notation

I read this more that brackets are more of a formal notation.  Shorthand notation just works around that.

gregsdennis (approved):
This is fine with me.

draft-ietf-jsonpath-base.md
cabo: > I take your point, but actually the existing text does the same. We could add "e.g., " inside each of the parentheses, but that seems a bit fussy.

Indeed.  I just wanted to mention this observation, but it is not in the way of merging this PR.

cabo:
Oh, and

```
399:Selectors are detailed in the next section. 
*** Run 'make fix-lint' to automatically fix some errors
draft-ietf-jsonpath-base.md contains trailing whitespace
make: *** [lib/main.mk:205: lint-whitespace] Error 1
```

Fixed.

Issue #470, opened 2023-05-23
`@` as a parameter may not be immediately intuitive for `count()`


gregsdennis:
None of the examples use `count(@)`, but I wonder if we should add a note that this will always return 1.  The recommended usage should be `count(@.*)`.

I ran into this when writing some tests.  I expect others will, too.

glyn:
From the title of this issue, in what way does `count(@)` _not_ work as expected? It is simply confusion between `count()` and `length()`?

gregsdennis:
I expected `count(@)` to give me the number of children of the value at `@` the way that `length(@)` gives the length of the string at `@`.

So for example, with the document `[{"a": "ab", "b": "bc"}]`, I expected to get 2.  Thinking about it, I worked it out, but it was confusing at first.  Just thought path authors may benefit from a note.

glyn:
I'm not entirely convinced that a clarification is needed, but I've floated one in PR #471.

glyn:
We failed to reach consensus on a fix in PR #471:

* There is nothing special about `@`. When `count()` is applied to any singular query, the result is `1`. This can be unexpected if the semantics of `count()` have been misunderstood. So adding an example of `count(@)` seems odd unless some explanation is given.
* Attempts to clarify the semantics of `count()` resulted in redundant text.
* `count(@)` is of no practical use.
* The suggestion of renaming `count()` to remove the superficial similarity of `count()` and `length()` was not accepted.

Unless we can find another approach, this issue is going to be hard to fix.

@gregsdennis observed that someone might think that `count(@)` would return the number of children of `@` if they have confused nodelists with arrays. The definition of nodelist says that a nodelist can be represented as a JSON array which might cause someone to think that nodelists and JSON arrays are interchangeable. If that is the root cause of this issue, perhaps some clarification of the definition is needed.

gregsdennis:
> Attempts to clarify the semantics of count() resulted in redundant text.

Neither of my proposals had redundant text.

I'm beside myself that no one is taking this seriously, especially considering that I, one of the authors of this text, made this mistake.  If I can do it, others definitely will.

I simply wanted a clarifying note that `count(@)` always produces one and that subqueries on `@` were needed to get the desired result.

glyn:
Editorially, it's quite awkward to add an unmotivated example which is likely to raise further questions in the minds of readers, such as "What's so special about `@`?".

[...]

PR #471, opened 2023-05-25
Clarify count() behaviour

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/470

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1599,7 +1599,7 @@ Notes:
 
 * There is no deduplication of the nodelist.
 * There is one node in a nodelist consisting of a single node, even if that node is an array or an object
-with multiple children. For example, `count(@)` always equals 1, as does `count($)`. 
+with multiple children. For example, `count(@)` always equals 1, as does `count($)`.
 
 
 ### `match()` Function Extension {#match}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1599,7 +1599,7 @@ Notes:
 
 * There is no deduplication of the nodelist.
 * There is one node in a nodelist consisting of a single node, even if that node is an array or an object
-with multiple children. For example, `count(@)` always equals 1, as does `count($)`.
+with multiple children. For example, `count(@)` always equals 1.
 
 
 ### `match()` Function Extension {#match}
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1594,7 +1594,12 @@ $[?count(@.*.author) >= 5]
 Its only argument is a nodelist.
 The result is a value, an unsigned integer, that gives the number of
 nodes in the nodelist.
-Note: there is no deduplication of the nodelist.
+
+Notes:
+
+* There is no deduplication of the nodelist.
+* There is one node in a nodelist consisting of a single node, even if that node is an array or an object
+with multiple children. For example, `count(@)` always equals 1, as does `count($)`. 
 
 
 ### `match()` Function Extension {#match}


draft-ietf-jsonpath-base.md
```diff
@@ -1594,7 +1594,12 @@ $[?count(@.*.author) >= 5]
 Its only argument is a nodelist.
 The result is a value, an unsigned integer, that gives the number of
 nodes in the nodelist.
-Note: there is no deduplication of the nodelist.
+
+Notes:
+
+* There is no deduplication of the nodelist.
+* There is one node in a nodelist consisting of a single node, even if that node is an array or an object
+with multiple children. For example, `count(@)` always equals 1, as does `count($)`.
```
gregsdennis, line 1602: I think we need to focus more on that `@` itself returns a single-node nodelist.  `count(@)==1` is then a mere consequence of this, and it becomes apparent why this note is warranted.

```suggestion
* `@` always returns a nodelist of length 1.  Therefore `count(@)` always equals 1.  To obtain the number of children of the current node, the subquery requires additional segments, e.g. `count(@.*)`.
```

This removes the redundant language that @cabo was referring to.

Side note: I don't think it's necessary to include `count($)` here, but leave it if you think it helps.

gregsdennis (changes requested):

draft-ietf-jsonpath-base.md
glyn: It would seem strange to start explaining the semantics of `@` here. If that needs doing, it would be better to do it in the definition of `@`.

I would prefer to keep the text here general. We could talk about `count()` of a singular query and give `@` as an example. But of course, `@.foo`, `@[23]`, etc. are equally valid examples and I'm not sure what's so special about `@`.

gregsdennis: I think the distinction is in the difference between how `length(@)` and `count(@)` operate, and the note needs to focus on that.  It's less about `@` itself, and more about how it's processed by each function.

For `length(@)`, the singular node's value is extracted from the nodelist and the "length" of that value is returned.  If the value is an object or array, then this is the number of children of that object or array.

But for `count(@)` that value extraction doesn't occur, even though you still have a singular nodelist.  The "count" then is the number of nodes in the nodelist, which is just one.

This distinction is what originally caught me.  Once I realized the distinction, I knew how to fix the queries.

It may seem obvious to you because you've thought through it, likely many times, but I assure you that it's not going to be immediately obvious to most people.  The confusion is compounded by the fact that many people will regard the nodelist as an array (which I know isn't right, but that won't stop people from thinking that way).

glyn: The behaviours of `count()` and `length()` are clearly defined by the existing text, but the above difficulty could arise when readers:

(a) overlook the fact that `count()` takes a parameter of type `NodesType`, and/or
(b) forget (or fail to understand) what a nodelist is, and/or
(c) forget which function is which.

There is also the distinct possibility that people will use these functions without reading the spec at all, e.g. by cutting and pasting from an example query found on the internet.

There is nothing special about the singular query `@` and any singular query (such as `@.foo` or `$[2]`) is equally likely to cause exactly the same confusion.

One possible improvement would be to rename `count()` to something like `nodes()`, `num_nodes()`, or similar. This emphases the nodelist aspect and is perhaps more easily distinguished from `length()`.

WDYT @gregsdennis and @cabo?

cabo: I don't think renaming `count()` would be an improvement.

I don't see a need for action here at all, but a single example might still help.  
`count(@)` would be one such example (although practically useless).

glyn: Ok, thanks, I'm going to delete the `count($)` example and merge.

glyn: @gregsdennis I'd like your approval before merging. I think we've nailed down `count(@)` which was your initial issue.

gregsdennis (changes requested):

glyn:
> (I'd prefer not doing anything over this.)

I'm happy with that too.

gregsdennis:
@glyn I think you missed the point in my saying that.  My point wasn't that I'm happy doing nothing.  My point was that what you had didn't help the problem.

glyn:
> @glyn I think you missed the point in my saying that. My point wasn't that I'm happy doing nothing. My point was that what you had didn't help the problem.

Fair enough, but rather than go in circles with this particular attempt at a solution, I propose we try to clarify the root cause of the problem in the issue for posterity/visibility. See https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/470#issuecomment-1565842505.

PR #474, opened 2023-05-28
(editorial): Second attempt to close #470

### LINKED ISSUE #470: `@` as a parameter may not be immediately intuitive for `count()`

cabo:
...by adding a clarifying note.


glyn (approved):
Works for me, thanks.

gregsdennis (approved):
I'm good with this. Thanks.

PR #475, opened 2023-05-29
JSON strings are unfortunately not sequences of scalar values.

### LINKED ISSUE #473: Terminology: Unicode Scalar Values
The last sentence of this definition says: "Both JSON string values and JSONPath queries are sequences of Unicode scalar values."[¶](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#section-1.1-6.38.1)

Despite what we may wish, this is not true. JSON string values are only code points. ECMA-404 says: "JSON syntax describes a sequence of Unicode code points." RFC8259 also admits this can be the case, but it comes with a warning about interoperability. Section 1.3 handles it well enough, so I think the last sentence in the Unicode Scalar definition can be removed.

"Also, selecting a child by name ([Section 2.3.1](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#name-selector)) and comparing strings ([Section 2.3.5.2.2](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#comparisons) in [Section 2.3.5](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#filter-selector)) assume these strings are sequences of Unicode scalar values, becoming unpredictable if they are not ([Section 8.2](https://rfc-editor.org/rfc/rfc8259#section-8.2) of [[RFC8259](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#RFC8259)])."[¶](https://ietf-wg-jsonpath.github.io/draft-ietf-jsonpath-base/draft-ietf-jsonpath-base.html#section-1.3-3)

sayrer:
Section 1.3 covers this hazard. This should fix #473.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -270,7 +270,7 @@ Normalized Path:
 Unicode Scalar Value:
 : Any Unicode {{UNICODE}} code point except high-surrogate and low-surrogate code points.
   In other words, integers in either of the inclusive base 16 ranges 0 to D7FF and
-  E000 to 10FFFF.  Both JSON string values and JSONPath queries are sequences of Unicode scalar values.
+  E000 to 10FFFF. JSONPath queries are sequences of Unicode scalar values.
 
 Segment:
 : One of the constructs which select children (`[<selectors>]`)


cabo (approved):
This works for me, as it removes the impression of an issue and does not really lose information.

glyn (approved):
Module my almost total mental block when it comes to Unicode, this change is benign.

PR #478, opened 2023-06-09
Delete extraneous license file

glyn:
Ref: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/477

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,0 @@
-# License
-
-See the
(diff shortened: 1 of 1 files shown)


cabo:
The TLP provide text that is intended to be in the processed documents.
This repo contains the sources.
I think there is an implied link here that makes the TLP licenses apply to the source, but that is not a foregone conclusion.
So having a license file for the source files may be useful.
CONTRIBUTING.md is almost that, but doesn't make the license really explicit.
I-d-template has a one-line LICENSE.md that points to CONTRIBUTING.md.

glyn:
Closing. I don't see how this PR will add value to the spec.

PR #492, opened 2023-07-26
Explain the firewall between query errors and data mismatches

cabo:
(as per AD review by Murray Kucherawy)

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -542,8 +542,7 @@ the JSON value the query is applied to. No further errors relating to the
 well-formedness and the validity of a JSONPath query can be
 raised during application of the query to a value.
 This clearly separates well-formedness/validity errors in the query
-from mismatches that may actually stem from flaws in the data (see
-also {{data-attack}}).
+from mismatches that may actually stem from flaws in the data.
 
 Mismatches between the structure expected by a valid query
 and the structure found in the data can lead to empty query results,
@@ -614,7 +613,7 @@ operations that might be considered erroneous, such as using an index
 lying outside the range of an array,
 simply result in fewer nodes being selected.
 (Additional discussion of this property can be found in the
-introduction to {{synsem-overview}} and in {{data-attack}}.)
+introduction to {{synsem-overview}}.)
 
 As a consequence of this approach, if any of the segments produces an empty nodelist,
 then the whole query produces an empty nodelist.
@@ -2179,8 +2178,7 @@ and the entries in the column
 Security considerations for JSONPath can stem from
 
 * attack vectors on JSONPath implementations,
-* attack vectors on how JSONPath queries are formed,
-* attack vectors on the data examined by JSONPath queries, and
+* attack vectors on how JSONPath queries are formed, and
 * the way JSONPath is used in security-relevant mechanisms.
 
 ## Attack Vectors on JSONPath Implementations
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -541,6 +541,16 @@ The well-formedness and the validity of JSONPath queries are independent of
 the JSON value the query is applied to. No further errors relating to the
 well-formedness and the validity of a JSONPath query can be
 raised during application of the query to a value.
+This clearly separates well-formedness/validity errors in the query
+from mismatches that may actually stem from flaws in the data (see
+also {{data-attack}}).
+
+Often, mismatches between the structure expected by a valid query
```
timbray, line 548: Suggest: It is possible that mismatches … in the data lead to empty query results.

(Don't think "often" is called for, any path to replace it is OK by me)

draft-ietf-jsonpath-base.md
```diff
@@ -541,6 +541,16 @@ The well-formedness and the validity of JSONPath queries are independent of
...
+and the structure found in the data will lead to empty query results,
+which may be unexpected and indicate bugs in either.
+JSONPath implementations might therefore want to provide diagnostics
```
timbray, line 551: suggest MAY?

draft-ietf-jsonpath-base.md
```diff
@@ -2212,6 +2225,20 @@ The resulting class of attacks, *injections* (e.g., SQL injections),
 is consistently found among the top causes of application security
 vulnerabilities and requires particular attention.
 
+## Attack Vectors on the Data Examined by JSONPath Queries {#data-attack}
```
timbray, line 2228: Really?  I'm having trouble believing this.  I think you should lose this and stick with the good wording up at the top of the review.  This is really extraneous to Murray's comment.

draft-ietf-jsonpath-base.md
```diff
@@ -2167,7 +2179,8 @@ and the entries in the column
 Security considerations for JSONPath can stem from
 
 * attack vectors on JSONPath implementations,
-* attack vectors on how JSONPath queries are formed, and
+* attack vectors on how JSONPath queries are formed
```
timbray, line 2182: I don't think this change adds any value.

timbray (changes requested):

draft-ietf-jsonpath-base.md
cabo: This is not an interoperability requirement.  It doesn't affect the other side.  We use MAY to say party A MAY do something, so party B ~~needs to~~ MUST be prepared for that behavior.  That's why I believe this is a "might" (non-2119 may).

draft-ietf-jsonpath-base.md
cabo: I went for a simple "can".

timbray (approved):

draft-ietf-jsonpath-base.md
cabo: Well, it was the main reason *I* care for the clear separation between query and data errors.
So I don't think any explanation of this is complete without this.
But I agree we don't *have* to explain all that.

timbray: Murray asked for a little bit of explanation for a basic fact about JSONPath that frequently surprises people.  I think adding a new security consideration is overkill, and also it's a significant change that I think the WG should get to comment on.  So I still think we should just skip this.

timbray (approved):

draft-ietf-jsonpath-base.md
cabo: I split out the seccons part to #493, so #492 should now be clean of security

glyn (approved):
Happy to include this or not. There are lots of unspecified behaviours implementations could provide. I think this probably adds some value to the spec, though.

PR #493, opened 2023-07-26
seccons part of PR#492

cabo:


timbray (commented):
Co-chair hat off: I am really unconvinced that this deserves this security-considerations coverage, since we've already been clear (and are clearer now) that any query can potentially yield empty result. But not lie-down-in-the-road mode.

Co-chair hat on: Unhappy about introducing substantive changes in the security considerations section at this stage in the process without WG input.

Going to shut up and see if any other voices want to pitch in.

glyn (approved):

cabo:
We shipped -16 without this, as Tim's chair perspective on this definitely is convincing.
We might revive this later if we find we do need it.

timbray:
Thanks.

cabo:
We didn't need it.

Issue #495, opened 2023-08-03
Inconsistent TOC for syntax and semantics sections


glyn:
Some syntax and semantics sections appear in the table of contents and some do not. Do we want to make this consistent?

cabo:
We could make all sections numbered, then more of them will go to the TOC.  We could also change tocdepth, which is currently 4.  My personal view is that the current settings are good, even if I'd normally use fewer unnumbered sections.

PR #496, opened 2023-08-05
Number all headings

glyn:
Fixes https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/495


cabo (approved):
This is now consistent, but also tedious; the ToC has become less useful than before.
I'd prefer a more selective approach, but I don't have a proposal, and I can live with this, too.

PR #497, opened 2023-08-07
Define S near where it is first used

glyn:
Darrel Miller pointed out that the definition of S is not easy to find when reading the document sequentially.


gregsdennis (approved):

cabo (approved):
That kind of depth-first definition is not my usual style.
Since much of the grammar is structured this way, and the current position in the slice selector is much wronger, I'll concede.

PR #498, opened 2023-08-08
Add an appendix with collected ABNF grammar

cabo:
We don't have to merge this, but this PR shows how to do it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,4 +13,5 @@ endif
 
 
 sourcecode: draft-ietf-jsonpath-base.xml
+	rm -rf sourcecode.ba?
 	kramdown-rfc-extract-sourcecode -tfiles draft-ietf-jsonpath-base.xml
--- a/Makefile
+++ b/Makefile
@@ -10,3 +10,7 @@ else
 	git clone -q --depth 10 $(CLONE_ARGS) \
 	    -b main https://github.com/martinthomson/i-d-template $(LIBDIR)
 endif
+
+
+sourcecode: draft-ietf-jsonpath-base.xml
+	kramdown-rfc-extract-sourcecode -tfiles draft-ietf-jsonpath-base.xml
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -2284,7 +2284,7 @@ defines the syntax of a JSONPath Normalized Path, while also using the rules
 ~~~ abnf
 {::include sourcecode/abnf/normalized_path_collected_abnf}
 ~~~
-{: #normalized-path-abnf title="Collected ABNF of JSONPath normalized paths"}
+{: #normalized-path-abnf title="Collected ABNF of JSONPath Normalized Paths"}
 
 
 # Inspired by XPath
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -147,8 +147,8 @@ from a JSON value.
 {:unnumbered: numbered="false" toc="exclude"}
 <!-- use as {: unnumbered} -->
 
-{:abnf: sourcecode-name="jsonpath-collected-abnf"}
-{:abnfnp: sourcecode-name="normalized-path-collected-abnf"}
+{:abnf: sourcecode-name="jsonpath-collected.abnf"}
+{:abnfnp: sourcecode-name="normalized-path-collected.abnf"}
 
 <!-- editorial issue: lots of complicated nesting of quotes, as in -->
 <!-- `"13 == '13'"` or `$`.  We probably should find a simpler style -->
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -147,6 +147,9 @@ from a JSON value.
 {:unnumbered: numbered="false" toc="exclude"}
 <!-- use as {: unnumbered} -->
 
+{:abnf: sourcecode-name="jsonpath-collected-abnf"}
+{:abnfnp: sourcecode-name="normalized-path-collected-abnf"}
+
 <!-- editorial issue: lots of complicated nesting of quotes, as in -->
 <!-- `"13 == '13'"` or `$`.  We probably should find a simpler style -->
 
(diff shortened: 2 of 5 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -2239,6 +2259,34 @@ Other attacks can target the behavior of underlying technologies such as UTF-8 (
...
+{{jsonpath-abnf}} contains the collected ABNF grammar that defines the
+syntax of a JSONPath query.
+
+~~~ abnf
+{::include sourcecode/abnf/jsonpath_collected_abnf}
+~~~
+{: #jsonpath-abnf title="Collected ABNF of JSONPath queries"}
+
+{{normalized-path-abnf}} contains the collected ABNF grammar that
+defines the syntax of a JSONPath normalized path, while also using the rules
```
glyn, line 2281:
```suggestion
defines the syntax of a JSONPath Normalized Path, while also using the rules
```

draft-ietf-jsonpath-base.md
```diff
@@ -2239,6 +2259,34 @@ Other attacks can target the behavior of underlying technologies such as UTF-8 (
...
+`root-identifier`, `ESC`, `DIGIT`, and `DIGIT1` from {{jsonpath-abnf}}.
+
+~~~ abnf
+{::include sourcecode/abnf/normalized_path_collected_abnf}
+~~~
+{: #normalized-path-abnf title="Collected ABNF of JSONPath normalized paths"}
```
glyn, line 2287:
```suggestion
{: #normalized-path-abnf title="Collected ABNF of JSONPath Normalized Paths"}
```

draft-ietf-jsonpath-base.md
```diff
@@ -2239,6 +2259,34 @@ Other attacks can target the behavior of underlying technologies such as UTF-8 (
...
 
+# Collected ABNF grammars
+
+This appendix collects the ABNF grammar from the ABNF passages used
+throughout the document.
+
+<!-- Update the collected grammar files using `make sourcecode`, which -->
+<!-- is currently manual as it creates a little circular dependency. -->
+<!-- The filenames of the ::includes are likely to change when -->
+<!-- kramdown-rfc-extract-sourcecode handles filenames better. -->
```
glyn, line 2270: With this branch, my workflow breaks. I normally issue `scripts/gen.sh`, but this now fails with:
```
 Created file draft-ietf-jsonpath-base.txt
 Created file draft-ietf-jsonpath-base.html
stdin(212:0): error: Rule jsonpath-query was already defined on line 0 of stdin
stdin(213:0): error: Rule segments was already defined on line 1 of stdin
stdin(214:0): error: Rule root-identifier was already defined on line 4 of stdin
...
stdin(389:0): error: Rule normal-index-selector was already defined on line 207 of stdin
parsing failed: 78 errors encountered
```
`make` however succeeds (after I installed npm), so maybe this doesn't matter.

glyn (changes requested):
Suggest capitalising Normalized Path as usual.

Also I wanted to check that `scripts/gen.sh` no longer working is to be expected.

draft-ietf-jsonpath-base.md
cabo: The problem with duplicating the ABNF is that ABNF doesn't accept duplicated rules (we learned from that in CDDL, which happily accepts [and checks!] duplicated rules).
We could change gen.sh to use sourcecode/abnf as well.

cabo:
> So, what should the RFC Editor do?

Well, I hope they don't have to do anything, as the ABNF is already fixed.
But if they do, it would be good if they did the AUTH48 in markdown, as recently has been done a number of times as an experiment.
Of course, we don't have a right to be part of such an experiment, but then RPC knows us well.

draft-ietf-jsonpath-base.md
glyn: > The problem with duplicating the ABNF is that ABNF doesn't accept duplicated rules (we learned from that in CDDL, which happily accepts [and checks!] duplicated rules).

Fair enough.

> We could change gen.sh to use sourcecode/abnf as well.

Yes please.

cabo: > > The problem with duplicating the ABNF is that ABNF doesn't accept duplicated rules (we learned from that in CDDL, which happily accepts [and checks!] duplicated rules).
> 
> Fair enough.
> 
> > We could change gen.sh to use sourcecode/abnf as well.
> 
> Yes please.

Done.

Now we just have to remember to do a `make sourcecode` manually; you could add this to gen.sh but because of the circular dependency I'm refraining from that now.  (We also could put the `bap` checks in make sourcecode...  But then we require people to have bap installed.)

Makefile
```diff
@@ -13,4 +13,5 @@ endif
 
 
 sourcecode: draft-ietf-jsonpath-base.xml
+	rm -rf sourcecode.ba?
```
glyn, line 16: Why does this not remove `sourcecode.bak`?

cabo: > Why does this not remove `sourcecode.bak`?

It does.  kramdown-rfc-extract-sourcecode then creates a new backup of the existing sourcecode directory.
A bit belt and suspenders, but I wanted to make sure the tool doesn't delete user information.
(We could remove the existing sourcecode directory as well, but then recovery is a bit harder if the sourcecode extraction fails for some reason.)

glyn: Sorry, I didn't express myself very well. Why doesn't the code say the following?
```suggestion
	rm -rf sourcecode.bak
```
Are there some other files/directories matching `sourcecode.ba?` that also need removing?

glyn:
> We don't have to merge this, but this PR shows how to do it.

BTW I really like this PR and think we should merge it. It's a convenience to certain users of the spec at very little cost to us or readers who are not interested in the collected ABNF.

glyn (approved):
This is good thanks. The ba?->bak change isn't a big issue.

Makefile
cabo: Because if the backup mechanism can't create a .bak, it will create .bal, .bam etc.
So if you did an extract-sourcecode manually a few times, you might have all these dropping.
.ba? takes care of all ~15 of them.

cabo:
(Based on an ARTART review comment by Darrel Miller.)

PR #499, opened 2023-08-08
Clarify wrapping filter expressions in parentheses

glyn:
Make it clear that this is not required, unlike in the original JSONPath article. Also, avoid using such parentheses in examples for consistency.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -505,8 +505,8 @@ The examples are based on the simple JSON value shown in
 | `$..book[2].publisher`                    | empty result: the third book does not have a "publisher" member |
 | `$..book[-1]`                             | the last book in order                                       |
 | `$..book[0,1]`<br>`$..book[:2]`           | the first two books                                          |
-| `$..book[?(@.isbn)]`                      | all books with an ISBN number                                |
-| `$..book[?(@.price<10)]`                  | all books cheaper than 10                                    |
+| `$..book[?@.isbn]`                        | all books with an ISBN number                                |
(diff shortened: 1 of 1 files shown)


draft-ietf-jsonpath-base.md
```diff
@@ -1173,6 +1173,11 @@ They have the normal semantics of Boolean algebra and obey its laws
 (see, for example, {{BOOLEAN-LAWS}}).
 Parentheses MAY be used within `logical-expr` for grouping.
 
+It is not required that `logical-expr` consist of
+a parenthesized expression (which was required in {{JSONPath-orig}}),
+although it can be and the semantics are the same
```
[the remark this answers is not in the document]
glyn: Yes, I think that works well. Added.

timbray (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1173,6 +1173,11 @@ They have the normal semantics of Boolean algebra and obey its laws
 (see, for example, {{BOOLEAN-LAWS}}).
 Parentheses MAY be used within `logical-expr` for grouping.
 
+It is not required that `logical-expr` consist of
```
cabo, line 1176: We now have removed all examples that actually show what is discussed here?
(We know what it is, but readers may not.)

cabo (approved):

draft-ietf-jsonpath-base.md
glyn: Yes, I was a little concerned about that too, so I've added a specific example near the top of the filter example table.

gregsdennis (approved):

PR #501, opened 2023-08-09
Add acknowledgement

glyn:
@cburgmer do you think the reference should include an organisation?

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -86,8 +86,8 @@ informative:
     target: https://goessner.net/articles/JsonPath/
     title: JSONPath — XPath for JSON
     author:
-      name: Stefan Gössner
-      org: Fachhochschule Dortmund
+    - name: Stefan Gössner
+    - org: Fachhochschule Dortmund
     date: 2007-02-21
   XPath: W3C.REC-xpath20-20101214
   E4X:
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

cabo (approved):
Thank you for catching this!

draft-ietf-jsonpath-base.md
```diff
@@ -77,16 +77,17 @@ informative:
     target: https://cburgmer.github.io/json-path-comparison/
     title: JSONPath Comparison
     author:
-      name: Christoph Burgmer
+    - name: Christoph Burgmer
```
cabo, line 80: That doesn't look right.

cabo: Organizations are not shown in references, unless they *are* the author (you now have two authors, one of which is the person, and one is the organization, which is certainly broken).

draft-ietf-jsonpath-base.md
```diff
@@ -73,15 +73,21 @@ contributor:
   uri: https://github.com/gregsdennis
 
 informative:
+  COMPARISON:
+    target: https://cburgmer.github.io/json-path-comparison/
+    title: JSONPath Comparison
+    author:
+    - name: Christoph Burgmer
+    - org: Thoughtworks
```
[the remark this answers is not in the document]
cabo: https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/pull/501#discussion_r1288406582

draft-ietf-jsonpath-base.md
glyn: Ok, I've switched to the same format as JSONPath-original (which I've also reverted). I think it's more important that the authors get the credit than their employers.

@cburgmer: please note with my apologies for thinking I could include the organisation in the rendered output.

PR #503, opened 2023-08-22
Workaround for vertical bar in table breaks (xml2rfc 3.18.0) and no longer needed

cabo:


gregsdennis:
What's the context for this change?

cabo:
> What's the context for this change?

In kramdown (parent project of kramdown-rfc), it is hard to include vertical bars in a table.
So I used an obscure feature of xml2rfc that converted broken bars into vertical bars.
With xml2rfc 3.18.0, this feature is gone, so we are seeing broken bars in -18.

https://author-tools.ietf.org/iddiff?url2=draft-ietf-jsonpath-base-18

Instead, this now uses a workaround via kramdown-rfc variable set/get.

> There are also two other places where `||` appears. Do we need to change those?

No, these were not using the broken bar hack, which is now entirely gone from the markdown source.

cabo:
I failed to check whether we even need the workaround any more.  We don't.

glyn (approved):
Nice, thanks! I'm going to merge this fast and cut a new base.

PR #507, opened 2023-08-24
Rob Wilton #3: clarify "are not ordered" in filter examples

cabo:
→ do not offer `<` comparison

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -1388,16 +1388,16 @@ Comparisons:
 | `$.arr != $.arr` | false | Array comparison |
 | `$.obj == 17` | false | Type mismatch |
 | `$.obj != 17` | true | Type mismatch |
-| `$.obj <= $.arr` | false | Objects and arrays are not ordered |
-| `$.obj < $.arr` | false | Objects and arrays are not ordered |
+| `$.obj <= $.arr` | false | Objects and arrays do not offer `<` comparison |
+| `$.obj < $.arr`  | false | Objects and arrays do not offer `<` comparison |
 | `$.obj <= $.obj` | true | `==` implies `<=` |
 | `$.arr <= $.arr` | true | `==` implies `<=` |
-| `1 <= $.arr` | false | Arrays are not ordered |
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -1388,16 +1388,16 @@ Comparisons:
 | `$.arr != $.arr` | false | Array comparison |
 | `$.obj == 17` | false | Type mismatch |
 | `$.obj != 17` | true | Type mismatch |
-| `$.obj <= $.arr` | false | Objects and arrays are not ordered |
-| `$.obj < $.arr` | false | Objects and arrays are not ordered |
+| `$.obj <= $.arr` | false | Objects and arrays do not offer `<` comparison |
```
glyn, line 1391: Just wondering about the "offering" language. How about the following (and throughout)?
```suggestion
| `$.obj <= $.arr` | false | Objects and arrays cannot be compared using `<` |
```

glyn (approved):
Suggested wording improvement.

draft-ietf-jsonpath-base.md
cabo: Well, this sounds like a factual statement, but it is not clear if that is a fundamental property (here, it actually pretty much is), or if it is a result of what we decided.

cabo: We could s/offer/provide for/

glyn: "provide for" is about the same as "offer" to my taste. How about "do not support"?

glyn: (On the factual statement point, "not offering" is equally factual - it doesn't say that this spec doesn't offer.)

cabo: The meaning is the same, the connotations aren't.  I'd like stick with "provide for" if you don't like "offer".

glyn: It's not a big deal, just a matter of taste. Thanks for considering the alternatives. Merging...

PR #509, opened 2023-08-24
Rob Wilton comments (editorial)

cabo:


gregsdennis (approved):

draft-ietf-jsonpath-base.md
```diff
@@ -682,6 +682,19 @@ and produces a nodelist consisting of that root node.
...
+<aside markdown="1">
+In this and the following examples in Sections {{<root-identifier}} and
+{{<selector-details}} except for {{tbl-comparison}}, we will present a
+JSON text to show the JSON value used as the query argument to the
+queries in the examples, and then a table with the columns:
+
+* Query: an example query to be applied to the query argument
+* Result: the query result as a list of JSON values that were located in the query argument
+* Result Path: the query result as a list of (normalized) paths into
+  the query argument, giving locations in parallel to the previous column
```
glyn, line 694:
```suggestion
  the query argument, giving locations of the JSON values in the previous column
```

glyn: Do we want a full stop at the end of the last bullet?

draft-ietf-jsonpath-base.md
```diff
@@ -834,11 +847,12 @@ Queries:
...
 The examples in {{tbl-name}} show the name selector in use by child segments:
 
-| Query | Result | Result Paths | Comment |
-| :---: | ------ | :----------: | ------- |
-| `$.o['j j']['k.k']`   | `3` | `$['o']['j j']['k.k']`      | Named value in nested object      |
-| `$.o["j j"]["k.k"]`   | `3` | `$['o']['j j']['k.k']`      | Named value in nested object      |
-| `$["'"]["@"]` | `2` | `$['\'']['@']` | Unusual member names
+| Query                | Result       | Result Paths            | Comment                      |
+| :---:                | ------       | :----------:            | -------                      |
+| `$.o['j j']`        | `{"k.k": 3}` | `$['o']['j j']`        | Named value in nested object |
```
glyn, line 852: It's probably accurate, but having the same comment for this as for the following two rows doesn't read very well.

glyn (approved):
Looks good thanks. I made some editorial suggestions.

draft-ietf-jsonpath-base.md
cabo: > Do we want a full stop at the end of the last bullet?

Not normally.  We either use punctuation throughout the list or leave it off entirely.
There are no sentences here, so the latter is the usual choice.

draft-ietf-jsonpath-base.md
cabo: Yes, I wondered about that.  "nested object" fits both a single and a dual level of nesting...

cabo: See commit below...

PR #510, opened 2023-09-21
RW-C-1: Be more explicit about what these "exact values" are.

cabo:
COMMENT 1 from Rob Wilton.

### DIFF
--- a/draft-ietf-jsonpath-base.md
+++ b/draft-ietf-jsonpath-base.md
@@ -538,8 +538,9 @@ requirements posed by this document, which are:
 
 1. Integer numbers in the JSONPath query that are relevant
 to the JSONPath processing (e.g., index values and steps) MUST be
-within the range of exact values defined in I-JSON {{-i-json}}, namely
-within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
+within the range of exact integer values defined in I-JSON (see {{Section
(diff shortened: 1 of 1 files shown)


glyn (approved):
LGTM thanks.

draft-ietf-jsonpath-base.md
```diff
@@ -538,8 +538,9 @@ requirements posed by this document, which are:
 
 1. Integer numbers in the JSONPath query that are relevant
 to the JSONPath processing (e.g., index values and steps) MUST be
-within the range of exact values defined in I-JSON {{-i-json}}, namely
-within the interval \[-(2<sup>53</sup>)+1, (2<sup>53</sup>)-1].
+within the range of exact integer values defined in I-JSON (see {{Section
+2.2 of -i-json}}), namely within the interval \[−(2<sup>53</sup>)+1,
+(2<sup>53</sup>)−1].
```
gregsdennis, line 543: Is "integer" really necessary here?  "Integer numbers" is already the context from line 539.

cabo: It is not necessary (that's why it wasn't in there), but we have at least one data point that a reader would consider it helpful.  Do you see a downside?

gregsdennis: Not really. Just felt redundant.

gregsdennis (approved):

PR #512, opened 2023-09-21
character-repertoire issues

timbray:


draft-ietf-jsonpath-base.md
```diff
@@ -526,7 +526,7 @@ these nodes as a nodelist.
 
 A query MUST be encoded using UTF-8.
 The grammar for queries given in this document assumes that its UTF-8 form is first decoded into
-Unicode code points as described
+Unicode scalar valuess as described
```
cabo, line 529:
```suggestion
Unicode scalar values as described
```

draft-ietf-jsonpath-base.md
```diff
@@ -831,6 +831,12 @@ sequences of Unicode scalar values. In other words, normalization operations
 MUST NOT be applied to either the member name string `M` from the JSONPath or to
 the member name strings in the JSON prior to comparison.
 
+Note that the "\u" Unicode character escape is dissimilar from the
```
cabo, line 834: No, that is not the way we have defined JSONPath to be.
(Even if we wanted to change this now, \u1F609 is ambiguous (U+1f60 + character 9?), so we'd need to adopt one of the syntaxes \U or \u{}...)

cabo (changes requested):
Looks good, except for the attempt to get rid of the ugly JSON escapes.

cabo (approved):
(Ignoring the change to lib) these changes now look good.  They do show the tedium of skipping the surrogate code points in each ABNF range covering that space, but that may be the best minimum change we can do here.

cabo:
> What's the outcome of this change? 

Over on the jsonpath mailing list, I wrote:

> > The recent extensive discussions on [art@ietf.org](mailto:art@ietf.org) and [i18n@ietf.org](mailto:i18n@ietf.org) have heightened our senses for potential misunderstandings of the way JSONPath handles Unicode.  Pull request 512 [1] is intended to address this.
> >
> > Note that while this PR occurs during IESG processing, this is not based on an IESG comment.
Some of us still would like to make this change before the document is approved.
> >
> > These changes are editorial in nature in that the technical content does not change, it is just expressed in a way that may give less opportunities for confusion.  
> >
> > So I’m requesting that the WG please have a look and speak up if you see a problem with this editorial improvement.

I think that should answer your question.

> Is it that surrogate pairs are no longer to be supported? 

There is no technical change here; JSONPath parsers never see surrogate pairs; these happen only inside UTF-16.
The ABNF changes make it more obvious that lone surrogate code points cannot occur in a JSONPath query; that is already the case now (see Section 1.1 in conjunction with RFC 3269).

> What's the reasoning?
> 
> Providing some context in the description on PRs would be good.

Sorry for putting this in the WG message only; the background is some extensive discussion on art@ and i18n@ that we don't want to repeat here in the JSONPath WG.

gregsdennis:
> I think that should answer your question.

No, that doesn't answer my question.  I saw that email and still asked.

> the background is some extensive discussion on art@ and i18n@ that we don't want to repeat here in the JSONPath WG.

Because I'm not subscribed to those, is there a summary?

timbray:
Basically, the JSONPath draft has always said that the character
repertoire for JSONPath is “Unicode scalar values”, which is all the code
points except surrogates, which only have meaning in UTF-16.  The recent
changes did a few s/code point/scalar value/ and changed one piece of ABNF
to make it clear that surrogates are not supported.

gregsdennis:
Thank you

glyn (approved):
LGTM, once the lib changes are reverted.

sourcecode/abnf/jsonpath_collected_abnf
```diff
@@ -133,7 +134,8 @@ bracketed-selection = "[" S selector *(S "," S selector) S "]"
 member-name-shorthand = name-first *name-char
 name-first          = ALPHA /
                       "_"   /
-                      %x80-10FFFF   ; any non-ASCII Unicode character
+                      %x7E-D7FF /   ; skip surrogate code points
```
cabo, line 137: Oops.  This should be %x80, not %x7E (~ is not a good name character).

draft-ietf-jsonpath-base.md
```diff
@@ -1811,7 +1812,8 @@ bracketed-selection = "[" S selector *(S "," S selector) S "]"
 member-name-shorthand = name-first *name-char
 name-first          = ALPHA /
                       "_"   /
-                      %x80-10FFFF   ; any non-ASCII Unicode character
+                      %x7E-D7FF /   ; skip surrogate code points
```
cabo, line 1815: (See above0

glyn (approved):
@cabo Thanks for fixing the typo! I was just about to dig into it...

PR #513, opened 2023-09-27
Rename collected ABNF files as per kramdown-rfc 1.7.1

cabo:
(Merge this when kramdown-rfc 1.7.1 has been updated in i-d-template.)


cabo:
`make update-deps`
already picks up kramdown-rfc 1.7.1, so this should be merged now.

glyn (approved):
LGTM

